PR #12, opened 2026-05-28
Support raw HSACO custom kernargs

AWoloszyn:
Preserve native kernarg layouts for raw HSACO kernels so prepacked HIP argument buffers from libraries such as hipBLASLt can dispatch correctly through the AMDGPU HAL.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h
@@ -54,8 +54,11 @@ typedef struct iree_hal_amdgpu_device_kernel_args_t {
   uint16_t constant_count;
   // Total number of bindings used by the dispatch (if a HAL dispatch).
   uint16_t binding_count;
+  // Offset of HIP/OpenCL implicit args within custom-direct kernargs, or
+  // UINT16_MAX if the kernel does not use an implicit suffix.
+  uint16_t implicit_args_offset;
   // Reserved for future hot kernel metadata. Must be zero.
-  uint32_t reserved;
+  uint16_t reserved;
 } iree_hal_amdgpu_device_kernel_args_t;
 IREE_AMDGPU_STATIC_ASSERT(
     sizeof(iree_hal_amdgpu_device_kernel_args_t) <= 64,
--- a/runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h
@@ -54,11 +54,8 @@ typedef struct iree_hal_amdgpu_device_kernel_args_t {
   uint16_t constant_count;
   // Total number of bindings used by the dispatch (if a HAL dispatch).
   uint16_t binding_count;
-  // Offset of HIP/OpenCL implicit args within custom-direct kernargs, or
-  // UINT16_MAX if the kernel does not use an implicit suffix.
-  uint16_t implicit_args_offset;
   // Reserved for future hot kernel metadata. Must be zero.
-  uint16_t reserved;
+  uint32_t reserved;
 } iree_hal_amdgpu_device_kernel_args_t;
 IREE_AMDGPU_STATIC_ASSERT(
     sizeof(iree_hal_amdgpu_device_kernel_args_t) <= 64,
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -2035,6 +2035,13 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_prepare_dispatch_plan(
   }
   out_plan->kernarg_strategy =
       IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_HAL;
+  if (IREE_UNLIKELY(out_plan->descriptor->custom_direct_only &&
+                    !iree_hal_amdgpu_aql_dispatch_plan_uses_custom_direct_arguments(
+                        out_plan))) {
+    return iree_make_status(
+        IREE_STATUS_INVALID_ARGUMENT,
+        "dispatch export requires custom direct kernarg arguments");
+  }
   if (IREE_UNLIKELY(inputs->constants.data_length > 0 &&
                     !inputs->constants.data)) {
     return iree_make_status(
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -1348,7 +1348,16 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_validate_dispatch_shape(
     const iree_hal_dispatch_config_t config, iree_hal_dispatch_flags_t flags) {
   const bool uses_indirect_parameters =
       iree_hal_dispatch_uses_indirect_parameters(flags);
-  if (iree_hal_amdgpu_dispatch_config_has_workgroup_size_override(config)) {
+  const bool has_workgroup_size_override =
+      iree_hal_amdgpu_dispatch_config_has_workgroup_size_override(config);
+  if (IREE_UNLIKELY(descriptor->custom_direct_only &&
+                    !has_workgroup_size_override)) {
+    return iree_make_status(
+        IREE_STATUS_INVALID_ARGUMENT,
+        "custom-direct-only ELF exports require a dispatch workgroup size "
+        "override");
+  }
+  if (has_workgroup_size_override) {
     for (iree_host_size_t i = 0; i < 3; ++i) {
       if (IREE_UNLIKELY(!config.workgroup_size[i])) {
         return iree_make_status(
@@ -2035,6 +2044,30 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_prepare_dispatch_plan(
   }
   out_plan->kernarg_strategy =
       IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_HAL;
+  if (out_plan->descriptor->custom_direct_only) {
+    if (IREE_UNLIKELY(
+            !iree_hal_amdgpu_dispatch_config_has_workgroup_size_override(
+                inputs->config))) {
+      return iree_make_status(
+          IREE_STATUS_INVALID_ARGUMENT,
+          "custom-direct-only ELF exports require a dispatch workgroup size "
+          "override");
+    }
+    for (iree_host_size_t i = 0; i < 3; ++i) {
+      if (IREE_UNLIKELY(!inputs->config.workgroup_size[i])) {
+        return iree_make_status(
+            IREE_STATUS_INVALID_ARGUMENT,
+            "dispatch workgroup size override must specify all dimensions");
+      }
+      if (IREE_UNLIKELY(inputs->config.workgroup_size[i] > UINT16_MAX)) {
+        return iree_make_status(
+            IREE_STATUS_OUT_OF_RANGE,
+            "dispatch workgroup size override dimension %" PRIhsz
+            " value %u exceeds %u",
+            i, inputs->config.workgroup_size[i], UINT16_MAX);
+      }
+    }
+  }
   if (IREE_UNLIKELY(out_plan->descriptor->custom_direct_only &&
                     !iree_hal_amdgpu_aql_dispatch_plan_uses_custom_direct_arguments(
                         out_plan))) {
@@ -2081,12 +2114,18 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_prepare_dispatch_plan(
     out_plan->kernarg_block_count =
         iree_max(1u, out_plan->descriptor->custom_kernarg_block_count);
     if (out_plan->layout->total_kernarg_size > 0) {
-      const uint32_t provided_kernarg_block_count =
-          (uint32_t)iree_host_size_ceil_div(
-              out_plan->layout->total_kernarg_size,
-              sizeof(iree_hal_amdgpu_kernarg_block_t));
+      const iree_host_size_t provided_kernarg_block_count =
+          iree_host_size_ceil_div(out_plan->layout->total_kernarg_size,
+                                  sizeof(iree_hal_amdgpu_kernarg_block_t));
+      if (IREE_UNLIKELY(provided_kernarg_block_count > UINT32_MAX)) {
+        return iree_make_status(
+            IREE_STATUS_OUT_OF_RANGE,
+            "dispatch kernargs require too many blocks (%" PRIhsz ", max=%u)",
+            provided_kernarg_block_count, UINT32_MAX);
+      }
       out_plan->kernarg_block_count =
-          iree_max(out_plan->kernarg_block_count, provided_kernarg_block_count);
+          iree_max(out_plan->kernarg_block_count,
+                   (uint32_t)provided_kernarg_block_count);
     }
     out_plan->kernarg_strategy =
         IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT;
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -1690,11 +1690,32 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
       }
       return iree_ok_status();
     }
-    case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT:
-      if (constants.data_length > 0) {
-        memcpy(tail_payload, constants.data, constants.data_length);
+    case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT: {
+      // Custom-direct callers hand us a pre-packed HIP ABI blob. Some raw
+      // HSACO kernels also need an implicit suffix synthesized after that blob,
+      // so zero the full reservation before copying only caller-owned bytes.
+      const iree_host_size_t total_kernarg_size =
+          layout->total_kernarg_size ? layout->total_kernarg_size
+                                     : constants.data_length;
+      memset(tail_payload, 0, total_kernarg_size);
+      const iree_host_size_t explicit_bytes =
+          layout->has_implicit_args ? layout->implicit_args_offset
+                                    : total_kernarg_size;
+      const iree_host_size_t copy_bytes =
+          constants.data_length < explicit_bytes ? constants.data_length
+                                                 : explicit_bytes;
+      if (copy_bytes > 0) {
+        memcpy(tail_payload, constants.data, copy_bytes);
+      }
+      if (layout->has_implicit_args) {
+        iree_amdgpu_kernel_implicit_args_t* implicit_args =
+            (iree_amdgpu_kernel_implicit_args_t*)(tail_payload +
+                                                  layout->implicit_args_offset);
+        iree_hal_amdgpu_aql_command_buffer_write_implicit_args(
+            kernel_args, config, implicit_args);
       }
       return iree_ok_status();
+    }
     case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_INDIRECT:
       return iree_make_status(
           IREE_STATUS_UNIMPLEMENTED,
@@ -2017,18 +2038,40 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_prepare_dispatch_plan(
 
   if (iree_hal_amdgpu_aql_dispatch_plan_uses_custom_direct_arguments(
           out_plan)) {
-    if (IREE_UNLIKELY(inputs->constants.data_length !=
-                      out_plan->descriptor->kernel_args.kernarg_size)) {
+    // Callers (e.g. rocBLAS/Tensile) sometimes omit trailing ABI padding or pad
+    // beyond the declared kernarg_segment_size with extra trailing scalars. The
+    // kernel only reads its declared size, so trailing bytes are ignored and the
+    // memcpy in write_dispatch_tail clamps to the declared size.
+    //
+    // Validate after 8-byte ABI padding so we accept missing tail padding while
+    // still rejecting truly short pre-packed HIP argument buffers.
+    const uint32_t required_explicit_bytes =
+        (uint32_t)out_plan->descriptor->custom_kernarg_layout
+            .explicit_kernarg_size;
+    const iree_host_size_t padded_constant_length =
+        iree_host_align(inputs->constants.data_length, /*alignment=*/8);
+    if (IREE_UNLIKELY(padded_constant_length < required_explicit_bytes)) {
       return iree_make_status(
           IREE_STATUS_INVALID_ARGUMENT,
-          "custom dispatch argument length mismatch; expected %u but got "
-          "%" PRIhsz,
-          out_plan->descriptor->kernel_args.kernarg_size,
-          inputs->constants.data_length);
+          "custom dispatch argument length too short; expected at least %u "
+          "but got %" PRIhsz " (padded to %" PRIhsz ")",
+          required_explicit_bytes, inputs->constants.data_length,
+          padded_constant_length);
     }
     out_plan->layout = &out_plan->descriptor->custom_kernarg_layout;
     out_plan->kernarg_block_count =
         iree_max(1u, out_plan->descriptor->custom_kernarg_block_count);
+    if (out_plan->layout->total_kernarg_size == 0 &&
+        inputs->constants.data_length > 0) {
+      // Some raw kernels have no reflected kernarg size. In that case the
+      // caller-provided blob is the only reservation size we can trust.
+      const uint32_t provided_kernarg_block_count =
+          (uint32_t)iree_host_size_ceil_div(
+              inputs->constants.data_length,
+              sizeof(iree_hal_amdgpu_kernarg_block_t));
+      out_plan->kernarg_block_count =
+          iree_max(out_plan->kernarg_block_count, provided_kernarg_block_count);
+    }
     out_plan->kernarg_strategy =
         IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT;
     return iree_ok_status();
@@ -2102,14 +2145,18 @@ iree_hal_amdgpu_aql_command_buffer_calculate_dispatch_layout(
           ? 0
           : (iree_host_size_t)plan->kernel_args->binding_count *
                 sizeof(uint64_t);
-  const iree_host_size_t tail_byte_length =
-      plan->layout->total_kernarg_size - binding_bytes;
+  const iree_host_size_t total_kernarg_size =
+      iree_hal_amdgpu_aql_dispatch_plan_uses_custom_direct_arguments(plan) &&
+              plan->layout->total_kernarg_size == 0
+          ? inputs->constants.data_length
+          : plan->layout->total_kernarg_size;
+  const iree_host_size_t tail_byte_length = total_kernarg_size - binding_bytes;
   IREE_RETURN_IF_ERROR(iree_hal_amdgpu_aql_command_buffer_qword_length(
       tail_byte_length, "dispatch tail payload",
       &out_layout->kernarg.tail_length_qwords,
       &out_layout->kernarg.tail_padded_length));
   IREE_RETURN_IF_ERROR(iree_hal_amdgpu_aql_command_buffer_qword_length(
-      plan->layout->total_kernarg_size, "dispatch kernarg",
+      total_kernarg_size, "dispatch kernarg",
       &out_layout->kernarg.total_length_qwords,
       &out_layout->kernarg.total_padded_length));
   out_layout->kernarg.implicit_args_offset_qwords =
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -1691,21 +1691,15 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
       return iree_ok_status();
     }
     case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT: {
-      // Custom-direct callers hand us a pre-packed HIP ABI blob. Some raw
-      // HSACO kernels also need an implicit suffix synthesized after that blob,
-      // so zero the full reservation before copying only caller-owned bytes.
-      const iree_host_size_t total_kernarg_size =
-          layout->total_kernarg_size ? layout->total_kernarg_size
-                                     : constants.data_length;
-      memset(tail_payload, 0, total_kernarg_size);
-      const iree_host_size_t explicit_bytes =
-          layout->has_implicit_args ? layout->implicit_args_offset
-                                    : total_kernarg_size;
+      const iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
       const iree_host_size_t copy_bytes =
           constants.data_length < explicit_bytes ? constants.data_length
                                                  : explicit_bytes;
       if (copy_bytes > 0) {
         memcpy(tail_payload, constants.data, copy_bytes);
+      }
+      if (copy_bytes < explicit_bytes) {
+        memset(tail_payload + copy_bytes, 0, explicit_bytes - copy_bytes);
       }
       if (layout->has_implicit_args) {
         iree_amdgpu_kernel_implicit_args_t* implicit_args =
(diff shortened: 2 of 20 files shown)


stellaraccident:
CI caught an error.

runtime/src/iree/hal/executable.h
```diff
@@ -149,14 +149,21 @@ typedef uint16_t iree_hal_executable_function_parameter_flags_t;
...
+  // Widened from uint8_t so we can represent kernarg structs emitted by
+  // user toolchains.
+  uint16_t size;
   // Offset of the parameter in bytes or binding ordinal, depending on type.
   uint16_t offset;
+  // Backend-reported byte offset of this parameter within the raw kernarg
+  // segment. Consumers that synthesize raw kernarg blobs must use this field
+  // for BINDING parameters because |offset| may be a binding ordinal for
+  // backends that follow the generic HAL contract.
+  uint16_t kernarg_offset;
```
benvanik: this feels like leakage: the offset should be the offset required, and there shouldn't be a need for two (and "kernargs" aren't a generic concept, anyway)

runtime/src/iree/hal/drivers/amdgpu/executable.c
```diff
@@ -1923,6 +1930,53 @@ static iree_status_t iree_hal_amdgpu_executable_resolve_raw_hsaco_kernel_args(
             requirements.binding_count, &host_kernel_args[kernel_ordinal]),
         "resolving kernel args for raw kernel `%.*s`", (int)symbol_name.size,
         symbol_name.data);
+    // Raw HSACO metadata is the source of truth for caller-visible kernargs:
```
benvanik:
```suggestion

    // Raw HSACO metadata is the source of truth for caller-visible kernargs:
```

runtime/src/iree/hal/drivers/amdgpu/executable.c
```diff
@@ -1923,6 +1930,53 @@ static iree_status_t iree_hal_amdgpu_executable_resolve_raw_hsaco_kernel_args(
...
+    // some code objects report a smaller HSA symbol size than the metadata
+    // segment needed by pre-packed HIP launch buffers.
+    if (host_kernel_args[kernel_ordinal].kernarg_size <
```
benvanik: now that there are 8 uses of `host_kernel_args[kernel_ordinal]` we should pull that into a local - but better would be to factor out the loop body here so iree_hal_amdgpu_executable_resolve_raw_hsaco_kernel_args is just the body writing to an out_host_kernel_args and put the loop in the caller (since host_kernel_args as an array here indexed by kernel_ordinal is entirely hoistable).

runtime/src/iree/hal/drivers/amdgpu/device/dispatch.h
```diff
@@ -86,10 +86,11 @@ iree_hal_amdgpu_device_dispatch_make_hal_kernarg_layout(
 static inline iree_hal_amdgpu_device_dispatch_kernarg_layout_t
 iree_hal_amdgpu_device_dispatch_make_custom_kernarg_layout(
```
benvanik: this looks shady - it's ignoring the kernarg size entirely and just zero initializing, which changes the meaning of this function (it shouldn't even exist if 0 is always expected) - but I'm not sure why you did this

runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c
```diff
@@ -96,14 +96,34 @@ void iree_hal_amdgpu_device_dispatch_emplace_hal_kernargs(
 }
 
 void iree_hal_amdgpu_device_dispatch_emplace_custom_kernargs(
+    const iree_hal_amdgpu_device_kernel_args_t* IREE_AMDGPU_RESTRICT
```
benvanik: now that custom kernargs is doing the same iree_hal_amdgpu_device_dispatch_emplace_implicit_args call and you have implicit_args_offset, we should hoist that out to the caller of both iree_hal_amdgpu_device_dispatch_emplace_hal_kernargs and iree_hal_amdgpu_device_dispatch_emplace_custom_kernargs instead of plumbing all the args all the way through to land on the same tail call. we'd expose iree_hal_amdgpu_device_dispatch_emplace_implicit_args in the header and update the host_queue_dispatch to do the detection.

runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h
```diff
@@ -54,8 +54,11 @@ typedef struct iree_hal_amdgpu_device_kernel_args_t {
   uint16_t constant_count;
   // Total number of bindings used by the dispatch (if a HAL dispatch).
   uint16_t binding_count;
+  // Offset of HIP/OpenCL implicit args within custom-direct kernargs, or
+  // UINT16_MAX if the kernel does not use an implicit suffix.
+  uint16_t implicit_args_offset;
```
benvanik: this is already present in iree_hal_amdgpu_device_dispatch_kernarg_layout_t - why are you replicating it here?

runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c
```diff
@@ -1143,6 +1150,418 @@ static iree_status_t iree_hal_amdgpu_hsaco_metadata_allocate_storage(
...
+      }
+      ++extra_count;
+      IREE_RETURN_IF_ERROR(iree_host_size_checked_add(
+          extra_symbol_name_storage_size, name.size + 3,
+          &extra_symbol_name_storage_size)
+                            ? iree_ok_status()
+                            : iree_make_status(
+                                  IREE_STATUS_OUT_OF_RANGE,
+                                  "AMDGPU metadata symbol name storage "
+                                  "size overflow"));
```
benvanik: make this an if(!checked) { return iree_make_status }

runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c
```diff
@@ -1143,6 +1150,418 @@ static iree_status_t iree_hal_amdgpu_hsaco_metadata_allocate_storage(
...
+      kernel->private_segment_fixed_size =
+          template_kernel->private_segment_fixed_size;
+      memcpy(kernel->required_workgroup_size,
+             template_kernel->required_workgroup_size,
+             sizeof(kernel->required_workgroup_size));
+      kernel->has_required_workgroup_size =
+          template_kernel->has_required_workgroup_size;
+      kernel->arg_count = template_kernel->arg_count;
+      kernel->args = template_kernel->args;
+      kernel->uses_borrowed_arg_layout = true;
```
benvanik: I don't know what this field indicates, but it doesn't seem used anywhere so you should remove it.

runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
```diff
@@ -1690,11 +1690,32 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
...
       return iree_ok_status();
     }
-    case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT:
-      if (constants.data_length > 0) {
-        memcpy(tail_payload, constants.data, constants.data_length);
+    case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STRATEGY_CUSTOM_DIRECT: {
+      // Custom-direct callers hand us a pre-packed HIP ABI blob. Some raw
+      // HSACO kernels also need an implicit suffix synthesized after that blob,
+      // so zero the full reservation before copying only caller-owned bytes.
+      const iree_host_size_t total_kernarg_size =
```
benvanik: this is our hot path and this is a lot of work, and is memsetting kernargs before writing them - this needs to be reworked. the layout exists so that this kind of calculation can be memoized and this code can be kept cheap. it's very critical to only touch memory once.

runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
```diff
@@ -2102,14 +2145,18 @@ iree_hal_amdgpu_aql_command_buffer_calculate_dispatch_layout(
           ? 0
           : (iree_host_size_t)plan->kernel_args->binding_count *
                 sizeof(uint64_t);
-  const iree_host_size_t tail_byte_length =
-      plan->layout->total_kernarg_size - binding_bytes;
+  const iree_host_size_t total_kernarg_size =
```
benvanik: this also should be memoized on the layout - the plan should never change after load, so there's no code that should ever switch on it - in other words: the "total_kernarg_size" is the total kernarg size regardless of plan.

benvanik (changes requested):

runtime/src/iree/hal/executable.h
AWoloszyn: Removed, I think that was a byproduct of about 12 rebases :D It's not used anywhere anyway.

runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c
AWoloszyn: Rebase chaff, removing.

benvanik (commented):
Second-pass review notes. I agree this is moving in a better direction overall, but I think there are still a few contract issues that need tightening before this is safe to merge. Most of them come from one pressure point: the raw/native kernarg model is starting to share helpers with the regular HAL/flatbuffer model, and those two projections need to stay separate.

## 1. Raw kernarg reflection leaks into flatbuffer executable loading

The new default HSACO reflection projection computes `constant_count` from the full aligned native kernarg segment:

- [`hsaco_metadata.c:L1706-L1725`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1706-L1725)

That helper is still used by the flatbuffer executable path, which validates it against the flatbuffer `ExportDef.constant_count`:

- [`executable.c:L1535-L1544`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/executable.c#L1535-L1544)

For a normal wrapped executable with two 64-bit bindings and two 32-bit constants, the native kernarg segment is 24 bytes, so this projects `constant_count == 6`. The flatbuffer ABI still says there are 2 constants. That means the flatbuffer path can reject valid wrapped executables because the raw-HSACO projection is now pretending padding and binding bytes are HAL constants.

The same leak shows up in parameter offsets. The public HAL contract says binding parameter offsets are binding ordinals:

- [`executable.h:L123-L131`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/executable.h#L123-L131)

But `populate_default_export_parameters` now writes raw kernarg byte offsets for bindings:

- [`hsaco_metadata.c:L1779-L1786`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1779-L1786)

Suggested fix: split this into two explicit projections instead of trying to make one "default" helper mean both things.

- Flatbuffer/HAL projection: compact visible metadata into HAL ABI reflection, with binding offsets as binding ordinals and constants as compact byte offsets/counts. This is what the wrapped executable path validates against `ExportDef`.
- Raw HSACO/native projection: preserve native kernarg offsets and expose enough byte coverage for custom-direct prepacked argument blobs. This should be used only by the raw HSACO executable path.

That split also gives the names room to state the invariant. Something like `calculate_hal_export_parameter_requirements` versus `calculate_native_kernarg_export_parameter_requirements` would make accidental cross-use much harder.

## 2. Custom-direct validation accepts truly short argument buffers for the common no-implicit-suffix case

`iree_hal_amdgpu_executable_initialize_dispatch_descriptor` leaves `custom_kernarg_layout` as all-zero unless an implicit suffix is detected:

- [`executable.c:L1142-L1157`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/executable.c#L1142-L1157)

Both planners then use `custom_kernarg_layout.explicit_kernarg_size` as the required minimum:

- [`aql_command_buffer.c:L2054-L2066`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c#L2054-L2066)
- [`host_queue_dispatch.c:L223-L235`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/host_queue_dispatch.c#L223-L235)

When there is no implicit suffix, that required size is zero. So the relaxed validation accepts any custom direct blob length, including one that is actually shorter than the declared kernel kernarg segment. The writer then uses the caller's length as the dynamic layout size:

- [`dispatch.c:L99-L113`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c#L99-L113)

That is looser than the intended "allow missing trailing ABI padding" rule. It can accept a genuinely short prepacked buffer and dispatch with missing argument bytes.

Suggested fix: even for dynamic custom-direct layouts, preserve a validated minimum explicit byte count from metadata/symbol information. The dispatch-time rule should be roughly:

```c
iree_host_align(constants.data_length, 8) >= layout->explicit_kernarg_size
```

Then the writer may clamp extra trailing bytes and zero missing padding up to the reservation. The dynamic part can still mean "the exact reservation may be caller-sized when no implicit suffix exists", but the minimum legal byte count should not be zero unless the kernel really has a zero-byte kernarg ABI.

## 3. Host-queue custom-direct block count truncates before the safety check

The host queue path computes a runtime block count from `constants.data_length` and immediately casts it to `uint32_t`:

- [`host_queue_dispatch.c:L245-L250`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/host_queue_dispatch.c#L245-L250)

The ring-capacity check later sees only the truncated value:

- [`host_queue_dispatch.c:L304-L310`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/host_queue_dispatch.c#L304-L310)

But the actual emplace path still `memset`/`memcpy`s the full `constants.data_length`:

- [`dispatch.c:L99-L113`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c#L99-L113)

So an oversized `iree_host_size_t` length can wrap to a small block count, pass validation, reserve too few queue kernarg blocks, and then overrun the queue-owned kernarg allocation. This is security/correctness-sensitive because the direct host-queue path is not protected by the command-buffer qword length guards.

Suggested fix: keep the ceil-div result in `iree_host_size_t`, reject values greater than `UINT32_MAX`, then cast. It would be worth sharing a tiny checked helper with executable descriptor initialization so every path that converts kernarg byte length to block count has the same overflow behavior.

## 4. ELF symbol augmentation can read past malformed section bounds

The new synthetic kernel augmentation scans ELF symbol sections, but the appended-symbol pass checks only that `symbol_section_offset <= elf_data.data_length`:

- [`hsaco_metadata.c:L1413-L1421`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1413-L1421)

It does not check `symbol_section_size <= elf_data.data_length - symbol_section_offset`. The first thing `iree_hal_amdgpu_hsaco_metadata_is_elf_kernel_symbol` does is read `symbol[4]`:

- [`hsaco_metadata.c:L1304-L1314`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1304-L1314)

The second pass repeats the same unchecked range pattern:

- [`hsaco_metadata.c:L1503-L1511`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1503-L1511)

So malformed ELF section metadata can produce out-of-bounds symbol reads before the later symbol-name helper gets a chance to reject the symbol.

Suggested fix: use the full range check used elsewhere in this file before deriving `symbol_count`, including multiplication/addition overflow checks for `symbol_count * symbol_entry_size`. If the section is malformed, skip it or fail loudly, but the code should never form a `symbol` pointer outside the ELF byte span.

## 5. `append_elf_kernel_symbols` leaks expansion allocations on failure

After the expansion storage is allocated:

- [`hsaco_metadata.c:L1456-L1464`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1456-L1464)

a failure allocating `extra_symbol_name_storage` leaks `expanded.kernels/args`:

- [`hsaco_metadata.c:L1465-L1470`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1465-L1470)

Later `IREE_RETURN_IF_ERROR` exits during the second scan can leak both allocations, because the expanded buffers have not yet been installed into `metadata` and the caller's cleanup only sees the original metadata object:

- [`hsaco_metadata.c:L1487-L1491`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1487-L1491)
- [`hsaco_metadata.c:L1593-L1595`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1593-L1595)

Suggested fix: make the expansion path a status-gated block with one terminal cleanup section. Until ownership is committed into `metadata`, `expanded.kernels` and `expanded.owned_string_storage` need local cleanup on every error path.

## Larger design concern: synthetic ELF kernels need to be quarantined/opted in

This helper now invents metadata for ELF function symbols that do not have metadata entries by picking one template kernel and copying its ABI/workgroup metadata:

- [`hsaco_metadata.c:L1441-L1454`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1441-L1454)
- [`hsaco_metadata.c:L1519-L1548`](https://github.com/ROCm/hrx-system/blob/cd925b38c50c33a170a1537f1f3d99a713a3426b/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1519-L1548)

That is a very strong producer-specific assumption. It may match a Tensile/hipBLASLt packaging pattern, but it is not generally true for arbitrary raw HSACO: a code object can contain multiple kernels with different argument layouts, required workgroup sizes, group/private segment sizes, or hidden suffix behavior. Publishing synthetic exports with copied metadata makes those kernels look valid until dispatch, where the failure mode is wrong argument layout rather than a clean load-time error.

Suggested fix: quarantine this behind an explicit opt-in. A few possible shapes:

- Put the synthetic-kernel expansion behind a raw-HSACO executable flag/option that names the producer assumption, instead of doing it unconditionally in `iree_hal_amdgpu_hsaco_metadata_initialize_from_elf`.
- Restrict it to a narrow, validated pattern: exactly one metadata template kernel, extra ELF kernel symbols known to share the same ABI by producer contract, and ideally an executable parameter that says HRX is intentionally loading that bundle form.
- Keep the generic parser honest: by default, only kernels with actual AMDGPU metadata should become dispatchable exports. Missing metadata should mean "not reflectable/not dispatchable by this generic path", not "borrow an arbitrary template".

If this expansion stays, tests should include at least one negative case with two real metadata kernels of different ABI plus one extra symbol, so the loader proves it does not silently clone the wrong contract.

benvanik:
(my prompt to gpt5.5-xhigh was: `› github PR review time! https://github.com/ROCm/hrx-system/pull/12/ - I've already done a first pass and think it looks better, but I'd like you to see if there's any other risks in here (I found some performance ones in our critical dispatch path, and safety is always an issue with ensuring proper iree_status_t tracking/etc)` - I recommend doing the same - the concerns identified above are all valid)

AWoloszyn:
Fixed based on that comment

benvanik (commented):
Follow-up after the latest push at `9a0419d249cca8dd7fd41ace3cd7162580bd638e`.

This is materially better than the previous revision. The HAL-vs-native reflection split is real now, the host-queue block-count truncation issue is fixed, the ELF symbol bounds checks look much safer, and the old expansion-allocation leak disappeared with the redesign that stores ELF-only symbols separately instead of synthesizing fake metadata kernels.

The remaining issues I would treat as blocking are all the same contract boundary: generic raw HSACO loading needs to be producer-neutral, while HIP/Tensile/descriptor-symbol conveniences need to be explicit. Given HRX is moving away from flatbuffers, I do not think flatbuffer compatibility is the important pressure here. The important invariant should be:

- Raw HSACO exports are custom-direct by default.
- A raw HSACO export may use normal IREE HAL ABI packing only when the code object carries an explicit HRX/IREE ABI marker.
- HIP/Tensile implicit-suffix and ELF-only descriptor-symbol behavior should be separate opt-in producer modes or very narrowly validated cases.

## 1. Raw metadata-backed HSACO exports still allow non-custom HAL dispatch

Raw metadata-backed kernels are reflected using the native kernarg projection:

- [`hsaco_metadata.c:L1784-L1868`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1784-L1868)

That projection preserves native byte offsets and makes the reflected constant count cover the aligned native kernarg segment. That is the right shape for custom-direct/native-kernarg reflection, but it is not automatically the IREE HAL dispatch ABI.

Only ELF-only symbols are marked `custom_direct_only` today:

- [`executable.c:L1904-L1923`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/executable.c#L1904-L1923)

The normal decoded-metadata kernel loop initializes native reflection but does not mark those exports custom-direct-only:

- [`executable.c:L1857-L1903`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/executable.c#L1857-L1903)

So a raw HSACO export can still be dispatched without `IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS`. In that case the runtime goes through the HAL packing path and synthesizes:

```c
uint64_t bindings[binding_count];
uint32_t constants[constant_count];
```

at the start of kernarg memory:

- [`dispatch.c:L70-L90`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c#L70-L90)

That only works if the raw HSACO kernel happens to use exactly the IREE HAL ABI layout. It is not a safe generic inference from AMDGPU metadata. A generic HSACO from clang, HIP, Tensile, or our future compiler may have native kernarg offsets/padding/interleaving that do not match compact HAL packing.

Suggested fix: raw HSACO decoded-metadata exports should default to `custom_direct_only = true`, just like the ELF-only exports. Clear that bit only when the raw code object carries an explicit HRX/IREE ABI marker saying the kernel follows the IREE HAL ABI. In that marked case, use the HAL projection and normal HAL dispatch packing. Without that marker, dispatch without custom-direct args should fail loudly.

The marker can be whatever is most compatible with our producer path: a custom AMDGPU metadata field, a separate ELF note, or another explicit load option. The important part is that normal HAL packing is opt-in for raw HSACO, not inferred from the presence of metadata.

## 2. Explicit-only raw kernels still accept truly short custom-direct blobs

`iree_hal_amdgpu_executable_raw_hsaco_custom_kernarg_layout` now computes `explicit_args_end`, but it only writes that value to `out_explicit_kernarg_size` when it found a hidden arg:

- [`executable.c:L2046-L2086`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/executable.c#L2046-L2086)

For the ordinary generic raw HSACO case with visible explicit args and no hidden args, the function leaves:

```c
*out_explicit_kernarg_size = 0;
*out_implicit_args_offset = UINT16_MAX;
```

The validation then sees a zero-byte required minimum:

- [`aql_command_buffer.c:L2094-L2105`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c#L2094-L2105)
- [`host_queue_dispatch.c:L233-L243`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/host_queue_dispatch.c#L233-L243)

That means a metadata-backed raw kernel with real explicit arguments can still be dispatched with a genuinely short custom-direct argument blob. This is the case the previous review was trying to protect: accepting missing trailing ABI padding is fine; accepting missing actual argument bytes is not.

Suggested fix: always preserve the minimum explicit byte extent from metadata:

```c
*out_explicit_kernarg_size = explicit_args_end;
```

Then layer any implicit-suffix decision on top. A zero explicit size should mean the metadata proves the kernel has no visible explicit kernarg bytes, not that the loader could not infer the minimum.

## 3. Hidden args should be a strict full-suffix contract, not arbitrary metadata support

I do not think this PR needs to implement arbitrary/interleaved hidden-arg support. The runtime does not support that today.

The existing ABI comments describe the actual contract we rely on: modern LLVM/HIP effectively includes the full implicit args block whenever any implicit arg is used, so the runtime reserves the full block and writes it all:

- [`kernel_args.h:L65-L90`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/abi/kernel_args.h#L65-L90)

The writer matches that model. It has one `has_implicit_args` bit and one `implicit_args_offset`; it does not walk individual hidden metadata records:

- [`dispatch.c:L45-L67`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c#L45-L67)

So the supported execution shape is:

```text
[visible explicit native kernarg bytes][optional full implicit args suffix]
```

The current PR instead treats the minimum hidden argument offset as the start of that full suffix:

- [`executable.c:L2058-L2080`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/executable.c#L2058-L2080)

That is only safe if every visible argument ends before the suffix and the code object has enough kernarg segment space for the whole `iree_amdgpu_kernel_implicit_args_t`. If metadata has a visible arg after the first hidden arg, the runtime will copy the visible native bytes and then overwrite them with the synthesized full implicit struct. The local test metadata even models that kind of interleaving: hidden arg at offset 8, then a visible by-value arg at offset 16:

- [`hsaco_metadata_test.cc:L199-L235`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata_test.cc#L199-L235)

Suggested fix: do not implement sparse hidden args here. Reject them.

The raw-HSACO custom layout logic should validate the one shape we actually execute:

```c
explicit_args_end = max(end of all non-hidden args);
first_hidden_offset = min(offset of all hidden args);

if (has_hidden_args) {
  if (explicit_args_end > first_hidden_offset) {
    return invalid_argument("unsupported interleaved hidden kernarg metadata");
  }
  if (kernel_args->kernarg_size <
      first_hidden_offset + IREE_AMDGPU_KERNEL_IMPLICIT_ARGS_SIZE) {
    return invalid_argument("truncated implicit kernarg suffix");
  }
  *out_explicit_kernarg_size = explicit_args_end;
  *out_implicit_args_offset = first_hidden_offset;
} else {
  *out_explicit_kernarg_size = explicit_args_end;
  *out_implicit_args_offset = UINT16_MAX;
}
```

That keeps the implementation honest: full LLVM/HIP suffix is supported; arbitrary hidden-arg layouts are a separate feature enablement, not silently half-supported today.

## 4. ELF-only symbol exports are safer now, but should still be opt-in producer behavior

The previous revision synthesized fake metadata kernels by copying a template. That is gone, which is a big improvement. The new shape stores ELF-only symbols separately and marks them custom-direct-only.

However, discovery still runs unconditionally inside the generic metadata initializer:

- [`hsaco_metadata.c:L1403-L1543`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1403-L1543)
- [`hsaco_metadata.c:L1556-L1575`](https://github.com/ROCm/hrx-system/blob/9a0419d249cca8dd7fd41ace3cd7162580bd638e/runtime/src/iree/hal/drivers/amdgpu/util/hsaco_metadata.c#L1556-L1575)

For a generic HSACO loader, I would expect dispatchable exports to come from real AMDGPU kernel metadata by default. Exporting extra descriptor-symbol pairs with no decoded metadata is a producer/package convention, not a generic AMDGPU metadata rule.

Suggested fix: keep the separate `elf_kernel_symbols` representation, but populate/use it only for an explicit raw-HSACO producer mode or loader option that says this bundle shape is expected. The default raw HSACO path should expose metadata-backed kernels, require custom-direct dispatch unless marked IREE ABI, and not invent extra exports from symbol-table shape alone.

## Suggested test coverage

The tests should pin these contracts so future changes do not drift back toward producer inference:

- Raw HSACO metadata kernel with visible args and no hidden args records a nonzero custom-direct minimum and rejects a short blob.
- Raw HSACO metadata kernel without an explicit IREE ABI marker rejects non-custom HAL dispatch.
- Raw HSACO metadata kernel with visible args after a hidden arg fails load or descriptor initialization as unsupported, rather than overwriting the visible bytes.
- A canonical full implicit suffix case is accepted.
- ELF-only descriptor-symbol exports are absent by default and present only when the producer mode/option is explicitly enabled.

AWoloszyn:
Re: 
4. ELF-only symbol exports are safer now, but should still be opt-in producer behavior

I am pretty sure we try to avoid mode-switches inside iree in general. Let me know if you want me to plumb a switch through, but it seems counter to our previous design.

benvanik (approved):
yeah, the concern is that doing the automatic scanning always will break any executable that is relying on ordinals as it'll pick up anything in an arbitrary order. but, with loom, I'm not going to be relying on ordinals anyway so I think it's ok to make the call that executable function ordinals are now unspecified, and then this is ok (sloppy, because HIP is sloppy, but the world we have to play in)

RE the failures, here's some free tokens :P 
```
Linux / CMake / GPU gfx942 fails in the new HRX executable CTS:

     libhrx/cts/hrx_cts_executable
     OUT_OF_RANGE; workgroup count buffer does not have the capacity to store the required 3 uint32_t values (length=0, min_length=12)

     The mechanism is clear: HRX defines its own dispatch flags as:

     HRX_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS = 1 << 0,
     HRX_DISPATCH_FLAG_ALLOW_INLINE_EXECUTION = 1 << 1,

     but hrx_queue_dispatch and hrx_stream_dispatch just cast those bits directly to iree_hal_dispatch_flags_t.

     In IREE, bit 0 is not custom-direct; bit 0 is IREE_HAL_DISPATCH_FLAG_DYNAMIC_INDIRECT_PARAMETERS. Custom-direct is bit 2:

     IREE_HAL_DISPATCH_FLAG_DYNAMIC_INDIRECT_PARAMETERS = 1ull << 0
     IREE_HAL_DISPATCH_FLAG_STATIC_INDIRECT_PARAMETERS = 1ull << 1
     IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS = 1ull << 2

     So the test passes HRX_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS, libhrx casts it to IREE bit 0, HAL thinks this is an indirect-parameters dispatch, and validation quite correctly demands a workgroup-count buffer. That’s why it
     reports length=0.

  The fix is not ceremony; it’s a real API boundary bug. Add a tiny translator and use it in both queue and stream dispatch:

  static hrx_status_t hrx_to_iree_dispatch_flags(
      uint32_t flags, iree_hal_dispatch_flags_t* out_flags) {
    iree_hal_dispatch_flags_t result = IREE_HAL_DISPATCH_FLAG_NONE;

    if (flags & HRX_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS) {
      result |= IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS;
    }
    if (flags & HRX_DISPATCH_FLAG_ALLOW_INLINE_EXECUTION) {
      result |= IREE_HAL_DISPATCH_FLAG_ALLOW_INLINE_EXECUTION;
    }

    const uint32_t known_flags =
        HRX_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS |
        HRX_DISPATCH_FLAG_ALLOW_INLINE_EXECUTION;
    if (flags & ~known_flags) {
      return hrx_make_status(HRX_STATUS_INVALID_ARGUMENT,
                             "unknown dispatch flags");
    }

    *out_flags = result;
    return hrx_ok_status();
  }

  Then replace:

  (iree_hal_dispatch_flags_t)flags

  with the translated hal_flags.
```

PR #13, opened 2026-05-28
Export AMDGPU host allocations and device properties

AWoloszyn:
Expose the device properties and host allocation exports needed by HRX so PyTorch can see accurate AMDGPU memory and architecture details.


stellaraccident (approved):
Since you are just flushing your perma-branches, I'm ok landing this as-is. But can you file an issue to add tests for these cases as a followup?

PR #14, opened 2026-05-28
A merge of a bunch of different changes for pytorch.

AWoloszyn:
This is a roll-up of a significant number of changes that are required for getting pytorch running on top of hrx.


libhrx/include/hrx_compiler.h
```diff
@@ -0,0 +1,69 @@
+// Copyright 2026 The HRX Authors
```
stellaraccident: I think we dropped this entirely (and the cc file).

stellaraccident (approved):
I don't think we need to carry the compiler support forward (likely you just got this from the merge but it was deleted in this branch).

As discussed, since you are flushing your perma-branches, I am ok with this going in as-is. But this needs unit and/or CTS tests, especially for things like the ELF/CCOB manipulation. Please file an issue so we don't lose track and can get this test coverage in. Not blocking this one so we can get out of branch debt.

libhrx/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ function(hrx_cc_test)
 endfunction()
 
 add_subdirectory(src/libhrx)
+add_subdirectory(src/binding/common)
```
zjgarvey, line 45: I might be misunderstanding something, but is there a reason we are still adding `src/streaming`?

libhrx/src/binding/common/stream.c
```diff
@@ -0,0 +1,1250 @@
...
+    }
+
+    // Track the submitted value for wait_submitted.
+    if (iree_status_is_ok(status)) {
+      stream->submitted_value = stream->pending_value;
+    }
+
+    // Release command buffer (we're done with it).
+    timing_step_ns = timing_enabled ? hrx_launch_timing_now_ns() : 0;
+    iree_hal_command_buffer_release(stream->command_buffer);
```
zjgarvey, line 408: Making note that the `stream->command_buffer` gets released here in `stream_flush`

libhrx/src/libhrx/transfer.c
```diff
@@ -51,3 +51,82 @@ hrx_status_t hrx_synchronous_d2h(hrx_device_t device, hrx_buffer_t src,
...
+  // Chunked synchronous H2D (matches streaming layer: 63KB chunks for
+  // inline BUFFER_UPDATE path compatibility with remote HAL).
+  const size_t chunk_size = 63 * 1024;
+  const uint8_t *src_ptr = (const uint8_t *)host_src;
+  size_t remaining = size;
+  size_t chunk_offset = 0;
+
+  while (remaining > 0) {
+    size_t this_chunk = remaining < chunk_size ? remaining : chunk_size;
+    iree_status_t iree_status = iree_hal_device_transfer_h2d(
```
zjgarvey, line 83: Isn't this API fully synchronous? Why is this section labelled `Async Stream Transfers`? I guess the same below for d2h. I wanted to understand how async transfers worked between host and device and I'm kind of confused here. Are the async transfers just not implemented yet?

zjgarvey: Maybe this is a terminology question more than anything: does "host" literally mean "host thread" so it must inherently block on the host thread? Or can some separate thread on the CPU do the transfer and signal to the host thread when it's done? Not sure how important this is unless we are paging data for a training job?

libhrx/src/binding/common/stream.c
```diff
@@ -0,0 +1,1250 @@
...
+    if (iree_status_is_ok(status)) {
+      status = iree_hal_device_queue_flush(stream->context->device,
+                                           stream->queue_affinity);
+    }
+    if (iree_status_is_ok(status)) {
+      stream->pending_value = signal_value;
+      stream->submitted_value = signal_value;
+    }
+  } else {
+    status = iree_hal_command_buffer_dispatch(
```
zjgarvey, line 1070: This writes the kernel launch into the stream->command_buffer, correct?

I'm trying to understand what the lock stream->mutex is being used for (It's only used in stream_begin and stream_flush), and since this isn't under lock, I'm wondering if it possible that this could be interleaved with a stream_flush from a different thread to result in a UAF/nullptr deref or something?

zjgarvey (commented):
I tried to read the bits which seemed important for me to understand better: e.g., kernel launching, and command buffer/stream interplay handling.

I have a few other questions, but could probably wait.

libhrx/src/libhrx/transfer.c
AWoloszyn: So I will answer both of the questions in ordedr

1)  Are the async transfers just not implemented yet? << Yes this is the case. This was originally used by the streaming layer, but then we ended up writing it ourselves in the streaming layer. (There is more work to be done to move more of binding/common to use the libhrx functions instead of iree functions directly.
2) "host" here means "memory residing in system RAM not GPU ram"

libhrx/CMakeLists.txt
AWoloszyn: Good catch, no its no longer needed.

libhrx/src/binding/common/stream.c
AWoloszyn: So it is legal to send commands to a hip stream from multiple threads at once. So, this is SOME work in the API to handle this, but it is incomplete.
I have filed #26 to add tests and fix the implementation here.

libhrx/include/hrx_compiler.h
AWoloszyn: Removed

PR #24, opened 2026-06-01
Fail iree_cc_test on TSAN reports

jtuyls:
Define __tsan_default_options() in gtest_main to set halt_on_error=1:abort_on_error=1, so a TSAN warning aborts the test binary instead of exiting 0 and being marked PASSed by ctest. TSAN_OPTIONS in the environment still wins for local debugging.

### DIFF
--- a/runtime/src/iree/testing/gtest_main.cc
+++ b/runtime/src/iree/testing/gtest_main.cc
@@ -5,8 +5,21 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree/base/api.h"
+#include "iree/base/target_platform.h"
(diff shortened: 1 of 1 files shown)


benvanik:
this was never needed in IREE - did something change external to this? TSAN definitely triggered failures there - is this to ensure that leak attribution is per test instead of the entire test run? (maybe you have an example before/after?)

jtuyls:
> this was never needed in IREE - did something change external to this? TSAN definitely triggered failures there - is this to ensure that leak attribution is per test instead of the entire test run? (maybe you have an example before/after?)

This was for the other PR  (https://github.com/ROCm/hrx-system/pull/25) in which I added a tsan test and I missed that this is not actually needed so closing.

PR #34, opened 2026-06-04
[Infra] Add shared AMDGPU target selection tooling

benvanik:
Move the AMDGPU target map, selector expansion, and code-object binary helpers under build_tools/amdgpu so runtime, HAL CTS, utility binaries, and libhrx CTS use the same target vocabulary.

Runtime AMDGPU support now uses IREE_HAL_AMDGPU_TARGETS in CMake and //runtime/src/iree/hal/drivers/amdgpu:targets in Bazel. Embedded device libraries, AMDGPU HAL CTS, source-built utility assets, and libhrx CTS native executable assets consume that same selector. HIP HAL CTS can use the selector when HIP testing is enabled.

HIP driver enablement is supported for Bazel testing and development without making RCCL mandatory. RCCL collectives are behind //runtime/src/iree/hal/drivers/hip:enable_rccl and IREE_HAL_DRIVER_HIP_RCCL.

Make the root presubmit dispatcher linked-worktree-safe and cap large non-verbose failure payloads. --verbose still streams full command output.

Staged precommit hooks using test-bearing profiles now run mechanical fixups before validation, and failing presubmit/precommit runs end with explicit suggested commands for repair and verbose reruns. The Python dev.py/Lefthook dispatcher is the only hook policy path.

Runtime Bazel presubmit excludes runtime resource tests by default, matching the CMake default-resource shape: local configured builds can cross-build AMDGPU/HIP-enabled targets without trying to run GPU resource tests outside dedicated resource lanes.

Migration examples:

CMake runtime, HAL CTS, and libhrx CTS target selection:
```
  old: -DIREE_ROCM_TEST_TARGET_CHIP=gfx942
  new: -DIREE_HAL_AMDGPU_TARGETS=gfx942
```

Bazel runtime, HAL CTS, and libhrx CTS target selection:
```
  old: --//build_tools/bazel:rocm_test_target=gfx942
  new: --//runtime/src/iree/hal/drivers/amdgpu:targets=gfx942
```

Multiple targets use commas in Bazel, such as gfx942,gfx950, and semicolons in CMake, such as gfx942;gfx950.


stellaraccident (approved):
Good improvement. Thanks

PR #36, opened 2026-06-05
[Infra] Add pinned ROCm header modes for AMDGPU runtime builds

benvanik:
This PR splits ROCm header availability from ROCm runtime/toolchain availability and builds on the already-landed shared AMDGPU target selection surface.

The short version:

- ROCm header facades now live under `build_tools/third_party/rocm/`.
- HSA runtime/AQL profile SDK and HIP API headers can come from pinned iree-org header archives.
- The same facades can still bind to a configured ROCm/TheRock root in package mode.
- RCCL remains optional and package-only.
- IREE AMDGPU CI now exercises pinned ROCm headers instead of accidentally selecting TheRock/package headers through `IREE_ROCM_PATH`.
- Local CMake precommit now uses the selected `dev.py` CMake build directory, builds in parallel, and fails early on stale Makefiles or mixed-generator CMake trees.

The intent is to let normal runtime/Loom development build AMDGPU host-side code without a multi-GB TheRock/ROCm install, while preserving libhrx CI as a canary for a full TheRock package distribution.

## Motivation

The old world assumed ROCm was either "there" or "not there", and a lot of build logic treated `IREE_ROCM_PATH` as the one switch for everything: headers, runtime libraries, device tooling, package discovery, and CI setup. That made AMDGPU support hard to reason about after moving out of the monorepo because host-side runtime code only needs a small header subset, while live execution, device compilation, and libhrx distribution testing need much more.

This PR establishes sharper dependency boundaries:

- Having HSA headers does not imply a ROCm runtime library exists.
- Having HIP API headers does not imply HIP is enabled as a production path.
- Having a ROCm/TheRock root does not require all source dependencies to come from packages.
- Enabling AMDGPU host-side code should not require Linux, AMD hardware, or a full TheRock install.
- libhrx can still validate against latest TheRock without making IREE runtime CI depend on it for headers.

This is also meant to be the model for future dependency work: pinned mode is deterministic source-tree development and IREE CI; package mode is explicit embedding/distribution validation; auto mode is local integration convenience.

### Dependency Modes

The existing global dependency mode remains:

```text
IREE_DEPENDENCY_MODE=pinned|package|auto
```

ROCm header facades add a narrower override:

```text
IREE_ROCM_DEPENDENCY_MODE=pinned|package|auto
```

Resolution is:

- If `IREE_ROCM_DEPENDENCY_MODE` is set, use it.
- Else if `IREE_ROCM_PATH` is set, use ROCm package mode.
- Else inherit `IREE_DEPENDENCY_MODE`, which defaults to `pinned`.

That means normal AMDGPU host builds can be:

```bash
python dev.py bazel configure -DIREE_HAL_DRIVER_AMDGPU=ON
python dev.py cmake configure -DIREE_HAL_DRIVER_AMDGPU=ON
```

TheRock/package validation is explicit:

```bash
python dev.py cmake configure \
  -DIREE_HAL_DRIVER_AMDGPU=ON \
  -DIREE_ROCM_PATH=/opt/rocm \
  -DIREE_ROCM_DEPENDENCY_MODE=package
```

And a build that needs a ROCm path for device tooling but still wants pinned headers can say that directly:

```bash
python dev.py bazel configure \
  -DIREE_HAL_DRIVER_AMDGPU=ON \
  -DIREE_ROCM_PATH=/opt/rocm \
  -DIREE_ROCM_DEPENDENCY_MODE=pinned
```

For Bazel portable configuration, `-DIREE_ROCM_PATH=/path` still writes the native repo env, but it also writes `IREE_ROCM_DEPENDENCY_MODE=package` unless the user explicitly provided a ROCm dependency mode. That keeps the meaning of "use ROCm at this path" concrete while still allowing pinned-header override.

### ROCm Header Facades

The ROCm-specific dependency code is now grouped in:

```text
build_tools/third_party/rocm/
```

It owns:

- Bazel `rocm_repository` mode selection and package/pinned repository setup.
- CMake ROCm header configuration logic.
- Narrow build files for ROCm header facades.

The exposed targets are:

```text
//third_party:hsa_runtime_headers
//third_party:aqlprofile_sdk_headers
//third_party:hip_api_headers
//third_party:rccl_headers

iree::third_party::hsa_runtime_headers
iree::third_party::aqlprofile_sdk_headers
iree::third_party::hip_api_headers
iree::third_party::rccl_headers
```

Pinned header sources are:

- `hsa_runtime_headers`: `iree-org/hsa-runtime-headers`, currently including the HSA runtime and AQL profile SDK header surface.
- `hip_api_headers`: `iree-org/hip-build-deps`, providing the HIP API header surface.
- `rccl_headers`: package-only for now; no pinned RCCL source is declared.

The facades are deliberately separate even when a package root or pinned archive can provide more than one header family. Code should depend on the capability it actually needs.

### Optionality

AMDGPU and HIP header consumers can compile in pinned mode without `IREE_ROCM_PATH`.

Package mode remains strict. If code explicitly asks for ROCm package headers and no configured ROCm/TheRock root provides them, configure fails loudly.

RCCL remains optional. Nothing in AMDGPU HAL host support or normal runtime development should imply RCCL. Code that needs RCCL must depend on the RCCL facade explicitly and use package mode or a package root that provides `rccl/rccl.h`.

### CI Split

IREE AMDGPU CI jobs still fetch TheRock where they need clang/runtime libraries, but they no longer export `IREE_ROCM_PATH` into the IREE configure environment. The source-tree IREE AMDGPU CI commands now configure:

```text
-DIREE_ROCM_DEPENDENCY_MODE=pinned
```

This makes CI prove that AMDGPU host-side runtime code builds against the pinned header facades.

libhrx workflows remain the TheRock/package canary. They intentionally configure with a ROCm/TheRock root and package mode for the distribution-style build path.

### CMake Precommit Tooling

While validating this, the local CMake precommit path exposed a real tool bug: project presubmit scripts hardcoded `../builds/<checkout>` and ran `cmake --build` without `--parallel`. If that build tree had stale `Unix Makefiles` or mixed-generator state, validation was painfully slow or misleading.

This PR adds a shared project-presubmit helper that:

- accepts the selected CMake build dir through `IREE_CMAKE_BUILD_DIR`,
- makes runtime/libhrx project presubmit use the build dir selected by `dev.py --cmake-build-dir`,
- builds with `cmake --build ... --parallel`,
- fails early if the tree is unconfigured,
- fails early if the tree uses `Unix Makefiles`,
- fails early if the cache says Ninja but `build.ninja` is missing.

### Normal AMDGPU host build

Before, users often set a ROCm root even for host-side compile coverage:

```bash
python dev.py bazel configure \
  -DIREE_HAL_DRIVER_AMDGPU=ON \
  -DIREE_ROCM_PATH=/opt/rocm
```

For pinned headers, the preferred path is now:

```bash
python dev.py bazel configure -DIREE_HAL_DRIVER_AMDGPU=ON
```

If a ROCm path is needed for other reasons but headers should stay pinned:

```bash
python dev.py bazel configure \
  -DIREE_HAL_DRIVER_AMDGPU=ON \
  -DIREE_ROCM_PATH=/opt/rocm \
  -DIREE_ROCM_DEPENDENCY_MODE=pinned
```

### TheRock/package validation

For libhrx distribution-style validation:

```bash
python dev.py cmake configure \
  -DIREE_ROCM_PATH=/opt/rocm \
  -DIREE_ROCM_DEPENDENCY_MODE=package \
  -DIREE_HAL_DRIVER_AMDGPU=ON
```


stellaraccident (approved):
I'm fine with this for the moment, but I would like to get things fixed upstream so that this isn't so all or nothing.

benvanik:
the example artifacts in https://therock-ci-artifacts.s3.amazonaws.com/26913628217-linux/index.html are interesting - if we can make it all (full TheRock, as libhrx is building with) or header-only packages with stable URLs we can pin to then I'd definitely want to do that. the iree-org stuff is here as a placeholder because I know I can get the runtime/loom working with those, but the second we have something published from the TheRock for the particular slices we need I'll switch to that - I don't want to maintain those iree-org repos :)

PR #37, opened 2026-06-05
Add amdxdna HAL driver for AMD XDNA NPUs

jtuyls:
Introduce an IREE HAL runtime driver targeting AMD XDNA NPUs directly through the in-kernel amdxdna driver's DRM ioctl ABI (kernel-managed queue / KMQ). It is self-contained: a small static user-space shim links into the runtime and talks to the device directly, keeping the dependency surface minimal. The driver is gated by the IREE_HAL_DRIVER_AMDXDNA build option and wired into the HAL driver registry, init.c, and libhrx's GPU-driver selection.

Executable format: a new pdi_executable_def.fbs ("PDIX") schema models an executable as a shared PDI pool plus per-entry-point runs. Each run is an XAie transaction ("TXN") control-code stream with an optional control-packet data_payload (array reconfiguration) and an optional host patch_table. Entry points reference PDIs by index, so several can share a single loaded PDI (e.g. manually merged kernels).

Submission paths: by default each command is submitted as ERT_START_CU and the firmware patches shim-DMA addresses. An opt-in path (--amdxdna_cmd_chain) batches a command buffer's dispatches into one ERT_CMD_CHAIN, host-patching the buffer-descriptor addresses from the compiler-emitted patch_table (validated for npu4).

Device model: host-side timeline semaphores and a single-worker async queue that defers HAL queue ops until their waits are satisfied and serializes all NPU access. Adds the allocator, buffers, no-op executable cache, events, and the Linux/KMQ native binding.

Shim: the user-space shim under shim/linux/kmq is self-contained, exception-free code rewritten from amd/xdna-driver's shim, plus the verbatim kernel UAPI (amdxdna_accel.h, GPL-2.0-WITH-syscall-note) and ERT ABI (ert.h, dual-licensed) headers. Provenance and per-file licenses are documented in shim/linux/kmq/README.md; per-file SPDX headers are authoritative.

Tests cover the allocator, buffers, async queue, driver, executable parsing/verification, semaphores/events, and the host patch-table logic (TXN op sizing, sentinel constant patching, and address patching).


jtuyls:
@benvanik Could you help review this PR?

benvanik (changes requested):
Requesting changes. This branch establishes a substantial amount of useful
XDNA scaffolding—schemas, Linux and Windows native adapters, host-side tests,
and an initial hardware lane—but it is not safe to land as a broad provisional
HAL driver in its current form. The review found independent allocator-contract,
memory-safety, synchronization, dispatch-semantics, executable-identity,
failure-path, ABI-provenance, and Windows lifetime defects. Several are
deterministically reproducible without XDNA hardware.

The bar here is necessarily high. A HAL device driver is core infrastructure:
incorrect metadata changes what generic runtime code is legally allowed to do;
an ordering shortcut can deadlock work that was correctly synchronized by the
application; an unchecked address patch becomes device memory corruption; and
an allocation or teardown gap can leave the process or device in an
unrecoverable state. “Provisional” may describe the supported feature set, but
it cannot weaken those contracts.

The strongest next revision is therefore a narrower, honest driver rather than
the current surface plus local patches. Unsupported dispatch, pool, or ABI
shapes should fail before allocation or submission. Required device state should
be initialized before publication. Context and cache identity should be explicit.
The imported native ABI should have a reproducible third-party boundary. Once
those invariants are small and named, the optimizations and broader CTS surface
can be added one at a time with production-shaped tests.

The companion comment immediately following this review contains complete
standalone sources and commands for the allocator metadata, semaphore-producer
hijack, already-satisfied-wait UAF, descriptor patching, and cache-publication
race probes summarized below.

## Blocking findings at a glance

| Area | Observed failure | Required direction |
| --- | --- | --- |
| Build/presubmit | Supported configure entry is missing; strict deps, generated CMake, formatting, clang-tidy, LSAN, and real-MSVC checks fail. | Make the feature buildable through normal entry points and run the paranoid profile with the driver actually enabled. |
| Allocator | A `HOST_ONLY` allocation is advertised as `HOST_COHERENT`, `HOST_LOCAL`, and sometimes `DEVICE_LOCAL`. | Report only properties implemented by the selected native heap and reject unsatisfied requirements. |
| Queue/semaphore | A later timeline producer can hijack an earlier wait; an already-satisfied wait can free its op while enqueue still reads it. | Preserve exact timeline identity and give callback registration an explicit lifetime sentinel. |
| Dispatch | Workgroup counts, indirect configuration, local memory, and semantic flags are discarded. | Implement the semantics or reject them before queue-op allocation; zero dimensions remain ordered no-ops. |
| Executable/context | PDI-less reuse means “last host recording won”; Windows cache identity omits a PDI that changes native creation. | Give every entry point a stable context identity and key caches by every native-creation input. |
| Address patching | Misaligned addresses are silently rounded; values above 48 bits corrupt descriptor metadata; binding extents are unavailable. | Use checked arithmetic, alignment/range validation, and preserve non-address bits exactly. |
| Initialization | Required context-cache OOM escapes device creation; two lazy command caches have C data races and discard creation errors. | Eagerly initialize required caches with status propagation before publishing the device. |
| Windows native path | Path-B activation is not atomic with aperture use; 69 DLL references are leaked per device through unsafe default lookup. | Define the hardware ownership scope and use one safely loaded, explicitly owned KMT dispatch table. |
| HAL pool/coherency | Test-oriented pool machinery is advertised but bypassed; public “no copy” flush/invalidate claims are false on Windows. | Return `UNIMPLEMENTED` for unsupported pool semantics and document/test the actual coherency contract. |
| Native ABI/failure model | Modified third-party ABI code is placed in the first-party tree without reproducible provenance; `-fno-exceptions` code uses throwing allocation. | Move imported sources and patches under `third_party/`; use checked IREE allocation or fixed-capacity state. |
| Status/error ownership | `DEFERRED` encodes signal ownership, `UNIMPLEMENTED` is a hot-path capability query, and 24 production statuses are ignored. | Keep status for terminal errors; represent scheduling/capabilities with ordinary typed state and join/propagate failures. |

The sections below give the mechanism, consequence, and expected production
shape for each item.

## A. Make the branch mechanically reviewable and buildable

These gates should be clean before the next deep review. They are not cosmetic:
the missing configuration caused the normal clang-tidy command to check zero
translation units, and clang-cl acceptance hid a real MSVC failure.

### A1. Add the supported Bazel configuration entry

The branch adds the `amdxdna` config setting, registry, and targets but not the
`IREE_HAL_DRIVER_AMDXDNA` option/mapping in
`build_tools/bazel/configure.py`. The normal repository entry point currently
fails with:

```text
iree-bazel-configure -DIREE_HAL_DRIVER_AMDXDNA=ON
Unsupported Bazel configure option: IREE_HAL_DRIVER_AMDXDNA
```

Raw `--//runtime/config/hal:drivers=amdxdna` selection reaches the targets, so
the missing piece is supported configuration integration. Add the option and a
driver/configure test; developers and CI should not need a private raw setting
to exercise a shipping driver.

### A2. Declare the direct Bazel dependencies

Strict include checking reports four test-target violations:

```text
driver_test.cc: missing provider for iree/base/threading/numa.h
semaphore_test.cc: missing provider for iree/async/proactor_platform.h
async_queue_test.cc: missing provider for iree/async/proactor_platform.h
completion_queue_test.cc: missing providers for iree/async/proactor.h and
                           iree/async/proactor_platform.h
```

Declare those direct dependencies. Please do not weaken strict checking or add
transitive include workarounds; the BUILD graph is the ownership map for these
tests.

### A3. Regenerate CMake and run the paranoid hook/profile

The exact current-base merge reports:

```text
python build_tools/bazel_to_cmake/bazel_to_cmake.py --check -v 1
File needs regeneration: runtime/src/iree/hal/drivers/amdxdna/CMakeLists.txt
```

Install the project hook once and run the full profile before every push:

```bash
```

The reviewed branch is not clean under that profile; the formatting step rewrites
PR files, and the generated CMake check fails. Please treat those commands as the
minimum submission gate rather than relying on a subset of live checks.

### A4. Run clang-tidy with the driver enabled

This apparently appropriate command exits successfully:

```bash
python dev.py bazel clang-tidy --profile ci -v \
  //runtime/src/iree/hal/drivers/amdxdna:all
```

but Bazel performs only one internal action because the driver is incompatible
under the default HAL selection. No source is checked. Until the configure entry
above supplies the feature selection, the exact working invocation is:

```bash
build_tools/bin/iree-bazel-build --keep_going \
  --repo_env=IREE_CLANG_TIDY_LLVM=auto \
  --aspects=//build_tools/clang_tidy:clang_tidy.bzl%collect_clang_tidy_aspect \
  --output_groups=iree_clang_tidy_reports \
  --//runtime/config/hal:drivers=amdxdna \
  //runtime/src/iree/hal/drivers/amdxdna:all
```

That run produces 114 diagnostics:

- 99 `iree-cpp-designated-initializer`;
- 5 `iree-refcount-lifecycle`;
- 4 `iree-guarded-release`;
- 3 `iree-status-discarded`; and
- 3 `iree-status-lifetime`.

The imported Linux shim accounts for most designated initializers. Moving
genuinely imported code under the third-party boundary requested below may
change how it is linted; code remaining in first-party runtime targets must
follow the repository rule. The status diagnostics include concrete ownership
bugs discussed in section G.

We will separately harden the clang-tidy wrapper so an explicit nonempty scope
that yields zero translation-unit reports cannot be green. That tooling gap does
not waive the branch diagnostics.

### A5. Make the new tests sanitizer-clean

`direct_command_buffer_chain_cache_test` completes all 19 assertions and then
LeakSanitizer reports 20,817 leaked bytes in 241 allocations from the test's
heap-owning move helpers. The test moves groups into stack cache entries and
does not deinitialize the moved ownership.

This does not by itself prove a production cache leak, but it means the test is
not validating the ownership contract it models. Deinitialize every moved group
and cache entry, retain the LSAN run, and add failure-injection coverage for
partial construction/move paths.

### A6. Preserve real MSVC support

`command_buffer_test.cc` uses C++20 designated initializers at three sites while
the target is configured as C++17. MSVC 19.44 reports C7555. Replacing them with
ordinary aggregate initialization is sufficient: after that diagnostic repair,
all 19 built AMDXDNA host tests pass under real MSVC, and all five hardware CTS
binaries compile.

Reproduction after the separate mainline MSVC-enablement repairs land, from an
x64 Visual Studio developer PowerShell:

```powershell
cmake -S . -B build-msvc-amdxdna -G Ninja `
  -DCMAKE_BUILD_TYPE=Release `
  -DCMAKE_C_COMPILER=cl.exe `
  -DCMAKE_CXX_COMPILER=cl.exe `
  -DCMAKE_ASM_COMPILER=cl.exe `
  -DIREE_BUILD_TESTS=ON `
  -DIREE_HAL_DRIVER_AMDXDNA=ON `
  -DIREE_HAL_DRIVER_AMDGPU=OFF `
  -DLIBHRX_BUILD_PASSTHROUGH=OFF

cmake --build build-msvc-amdxdna `
  --target iree_hal_drivers_amdxdna_command_buffer_test `
  --parallel 12

ctest --test-dir build-msvc-amdxdna `
  --output-on-failure `
  -R amdxdna `
  -j 12
```

The Windows CI matrix should eventually compile this driver in the mandatory
MSVC base lane and use clang-cl for the broader diagnostic/sanitizer fanout.
Clang-cl accepting a C++20 extension is not evidence for the MSVC product
contract.

## B. Restore the allocator and memory-visibility contract

### B1. Do not advertise memory properties the native heap does not implement

`iree_hal_amdxdna_allocator_query_buffer_compatibility` starts from “allocatable,”
preserves requested `DEVICE_LOCAL`, and unconditionally adds
`HOST_LOCAL | DEVICE_VISIBLE`. `HOST_LOCAL` includes `HOST_VISIBLE` and
`HOST_COHERENT`; coherent memory promises that callers do not need explicit
flush/invalidate operations.

The implementation does not provide that behavior. Every generic allocation is
a native `HOST_ONLY` buffer. Linux invokes native BO synchronization, while
Windows may allocate a separate `host_mirror` and copy ranges during
flush/invalidate. Preserving `DEVICE_LOCAL` on that allocation is also false.

The no-hardware probe reports:

```text
compatibility=0x00000801 type=0x00000076 host_visible=1 host_coherent=1 host_cached=0 size=4096
```

`0x76` also contains `DEVICE_LOCAL`. Generic HAL code may legally skip required
visibility operations after seeing those bits, so this can expose stale data to
either processor.

Report the resolved properties of the selected native heap, not the requested
properties:

- reject a required bit that cannot be satisfied;
- for the current `HOST_ONLY` path, report at least
  `HOST_VISIBLE | DEVICE_VISIBLE` without `HOST_COHERENT`, `HOST_LOCAL`, or
  `DEVICE_LOCAL`;
- establish `HOST_CACHED` separately for each backend rather than inferring it;
  and
- keep heap query, compatibility query, allocated-buffer metadata, and tests in
  agreement.

Regression coverage must prove that persistent non-coherent mappings require
flush/invalidate on Linux and Windows, including the Windows host-mirror path.

### B2. Document and test the real libhrx flush/invalidate behavior

The new public comments call flush and invalidate “Cheap (no copy).” Windows
host-mirror buffers perform `memcpy` for both operations, so that cross-platform
cost claim is false. The comments also do not say that the offset is relative to
the active mapped range, as required by the underlying HAL mapping contract.

Document ownership, coordinate system, and visibility semantics rather than an
implementation cost. Add a nonzero mapping-offset test and real bidirectional
visibility tests on both native backends. The existing mapping CTS uses offset
zero and does not call these APIs, so it does not cover this contract.

## C. Fix queue, semaphore, and callback lifetime semantics

### C1. A later timeline producer must not hijack an earlier wait

Each amdxdna timeline semaphore stores one `native_signal_batch` and one
`native_signal_value`. Publishing a later submission replaces the earlier
producer. `try_retain_native_signal` then redirects an infinite wait for value
`V` to the stored batch whenever the stored value is greater than or equal to
`V`.

For batch A signaling 1 and later batch B signaling 2, a waiter for 1 can
therefore wait on B. If B depends on host work that follows completion of 1, the
driver deadlocks even after A advances the semaphore to 1. HAL queues are not
implicitly FIFO; the application expressed the correct timeline edge, and the
driver introduced a dependency on later work.

The no-hardware reproducer completes A and observes:

```text
after value 1 completed: semaphore=1 waiter_returned=0
waiter status=0 hijacked=1
```

At minimum, the shortcut may retain a producer only for an exact value match.
An older-value wait can fall back to the semaphore's ordinary async wait. If a
single native batch is meant to cover a value range, represent the ordered
producer range explicitly and prove that selecting it cannot introduce a
dependency after the requested timeline point.

### C2. Registration must survive synchronous callback delivery

`iree_async_semaphore_acquire_timepoint` may invoke its callback synchronously
when the requested value is already satisfied. The amdxdna enqueue loop keeps
reading `op->wait_list.count` while arming timepoints. The last synchronous
callback can publish the op; the worker can execute it and free the op arena
before `acquire_timepoint` returns; then enqueue reads the freed op at
`async_queue.c:614`.

A 128-byte arena block forces the larger op through the immediately poisoned
oversized-allocation path and makes the legal interleaving deterministic under
ASAN:

```text
AddressSanitizer: heap-use-after-free
READ of size 8 in
  iree_hal_amdxdna_async_queue_enqueue_with_failure_handler
  async_queue.c:614
freed by the async queue worker through iree_arena_deinitialize
```

Caching the loop bound is not sufficient: an earlier async callback can still
consume the final count while later entries are being armed. Registration must
own a sentinel slot. Initialize the count to `wait_count + 1`; let callbacks and
failed registrations consume their entry slots; consume the registration slot
only after the loop finishes. Whichever decrement reaches zero publishes the
op. Validate the host-sized count before narrowing it into the signed atomic.

Keep the already-satisfied, oversized-arena reproducer as a checked-in ASAN
test. A pooled allocation that is intentionally unpoisoned can otherwise hide
the same use-after-free.

## D. Make dispatch semantics explicit and complete

### D1. Do not silently discard dispatch configuration or semantic flags

The recorded path resolves an `iree_hal_dispatch_config_t` and then drops it.
The direct vtable also drops its `config`, and `queue_dispatch` does not retain
one. The planner discards `flags`. Consequently all of these requests become
one native control-packet submission:

- static counts `1x1x1`, `2x1x1`, and `0x1x1`;
- indirect workgroup counts;
- default and explicit workgroup sizes;
- zero and nonzero dynamic local memory; and
- custom direct/indirect or indirect-parameter semantic flags.

`0x1x1` must be an ordered no-op but currently executes once; larger counts
silently execute once instead of failing or producing the requested work.

If the native artifact represents one whole NPU invocation, the correct first
contract is narrow:

- any zero static dimension is a semaphore-ordered no-op;
- only static `1x1x1` is executable;
- workgroup size is the all-zero default or an explicitly supported `1x1x1`;
- dynamic local memory is zero; and
- indirect/custom semantic modes return `UNIMPLEMENTED` until implemented.

Validate this before queue-op allocation/copying. If broader shapes are added,
carry the full config through recorded replay and prove their native semantics.

The current gfx1151 job registers allocator, mapping, and driver CTS tags with
no executable. It does not instantiate queue, semaphore, command-buffer, or
dispatch CTS, so its green result does not cover this surface.

### D2. Give PDI-less entry points a stable context identity

The schema defines `pdi_index = -1` as reusing a sibling-loaded context. The
branch's own `reuse_context` fixture has no PDI/xclbin and no reconfiguration
payload. Dispatch classifies `data_payload_count == 0` as self-contained and
calls context creation with both image spans empty, which deterministically
returns `INVALID_ARGUMENT`. The existing test parses the fixture but never
dispatches it.

Changing that condition alone is insufficient. The fallback context is mutable
executable-global state meaning “the most recent PDI-carrying dispatch recorded
on the host.” Independent command buffers can be recorded concurrently, and
host recording order is not HAL execution order. A-load/B-load/A-reuse/B-reuse
can therefore bind a reuse operation to the other context without a raw data
race.

Every entry point needs a stable context identity: the selected PDI/xclbin or an
explicit context-group ordinal. Resolve and retain by that identity while
recording, and express reconfiguration execution ordering through commands and
semaphores. `-1` must not mean “last host caller won.”

Add a real dispatch test for the existing fixture plus a forced concurrent
A/B recording interleaving and semaphore-ordered execution proving each reuse
operation reaches its declared context.

### D3. Include every native-creation input in the Windows cache key

For xclbin-capable Windows devices, the context-cache key is raw xclbin only.
Windows creation also executes the selected PDI through Path B. The schema and
the existing xclbin fixture permit one xclbin with multiple distinct PDI
indices, so the second entry point can reuse the context initialized with the
first PDI.

Key by every input that changes native creation: at least xclbin plus selected
PDI, and keep the CU/export name if it remains part of the public cache identity.
Test one xclbin with two PDI/name selections and observe two native creations;
test an identical tuple and observe one reuse.

### D4. Reject malformed patched addresses rather than changing them

The patch-table helper builds a 48-bit address, then clears the low two bits
instead of requiring alignment. A requested `0x80001003` therefore becomes
`0x80001000`. Addition is unchecked; a value above 48 bits spills into the
upper descriptor metadata. The helper also receives addresses but not binding
lengths, so it cannot prove `arg_plus` lies inside the submitted binding.

The no-hardware probe reports:

```text
misaligned accepted=1 bd2:bd1=00000000:80001000
overflow accepted=1 bd2:bd1=00010000:7ffffff0
```

Use checked additions, reject misaligned binding offsets/addends, enforce the
48-bit representable range, pass binding extents, and preserve template metadata
bits exactly. Delete the test that currently expects low-bit rounding. Add
rejections for an unaligned base, unaligned `arg_plus`, `arg_plus >= length`,
addition overflow, and a result above the address field.

## E. Publish only completely initialized device state

### E1. Context-cache allocation failure must fail device creation

`iree_hal_amdxdna_device_context_cache_create` converts allocator failure to a
null pointer. A `void` device initializer stores it and device creation succeeds;
the first lookup dereferences `device->context_cache->mutex`.

Make initialization status-returning or embed the cache. A device without
required dispatch state must never escape creation. Add a fault-injecting host
allocator test that receives `RESOURCE_EXHAUSTED` and proves all already-created
native resources are released.

### E2. Remove the lazy command-cache publication races

The single- and chain-command cache getters perform ordinary unsynchronized
pointer reads before creating under `command_cache_mutex`. The writer holds a
mutex that the reader does not, so the C load/store pairs are data races. Both
creation statuses are also cast to void.

TSAN reports both:

```text
Write in iree_hal_amdxdna_chain_command_cache_create
Previous read in iree_hal_amdxdna_get_chain_command_cache

Write in iree_hal_amdxdna_single_command_cache_create
Previous read in iree_hal_amdxdna_get_single_command_cache
```

Eagerly initialize both caches with status propagation before publishing the
device. That removes the hot-path branch/lock, eliminates the absent-cache state,
and makes allocation failure part of device creation. If lazy state is retained,
every access needs a real atomic publication protocol and a durable failure
channel; merely locking the writer is not sufficient.

## F. Fix Windows native ownership and loader lifetime

### F1. Path-B activation must remain owned through aperture use

Windows has one process-global `g_pathb_active_context`, but
`activate_pathb_context_for_submit` releases the global mutex before register
finalization, residency waits, aperture synchronization, staging, and KMT
submission. A second device worker can activate B, retire A, and release A's
aperture mapping while A is still using it. Per-device serialization does not
protect independent devices.

The relevant interleaving is:

```text
device A worker                 device B worker
-----------------------------   ---------------------------------
activate(A); unlock
finalize/stage A ...            activate(B); lock
                                retire A; release A mapping; unlock
submit through A aperture
```

Name the hardware scope explicitly. If Path B is process-global, one ownership
transaction must cover every operation that relies on the active mapping,
including destruction. If it is adapter-scoped, key and lock state at that
scope instead of globally serializing unrelated adapters. Remove the unused
per-device active-context members once the real owner is clear.

The Windows hardware regression needs two independently constructed device
objects, distinct recognizable command streams, forced concurrent activation,
and context creation/destruction while the peer submits. A single-device
sequential test cannot exercise this race.

### F2. Load each KMT system module once with an explicit search policy

`ResolveKmtProc` calls `LoadLibraryW` for `win32u.dll`, `gdi32.dll`, and
`dxcore.dll` on every symbol lookup. `KmtApi::Load` resolves 23 symbols, causing
69 reference-count increments per native-device creation. Only function
pointers are retained; no handle is released. Bare names also use the default
DLL search policy for privileged system graphics libraries.

Use one process-global immutable dispatch table, or a `KmtApi` owner that loads
each module once and balances it after all function pointers are dead. Use
`LoadLibraryExW(..., LOAD_LIBRARY_SEARCH_SYSTEM32)`, retain the documented
three-module fallback where Windows-version compatibility requires it, and
report the specific missing module/symbol. A repeated device-create/destroy test
should observe one balanced load per module or one deliberate process-lifetime
reference—not 69 leaked references per device.

## G. Remove test-shaped HAL semantics and status-shaped state

### G1. Do not advertise a queue pool that queue allocation bypasses

The driver constructs a slab provider, notification, passthrough pool, frontier
axis, and epoch paths specifically for explicit-pool CTS. It then advertises
that backend, while `queue_alloca` ignores `pool` and `flags`, performs an
ordinary synchronous allocation, and defers only the signal. `queue_dealloca`
sets a backend bit but neither decommits nor returns a pool reservation.

As a result, explicit capacity, strategy, reuse, notification, death frontier,
pool exhaustion, `ALLOW_POOL_WAIT_FRONTIER`, and
`INDETERMINATE_LIFETIME` do not affect production behavior. Queue-deallocating
an ordinary synchronous buffer can also poison later use even though HAL says
that case degrades to an ordering barrier.

The correct narrow first version is to remove the fake default-pool/frontier
machinery, return `UNIMPLEMENTED` from `query_queue_pool_backend`, reject a
non-null explicit pool, and validate every alloca/dealloca flag. Direct queue
allocation can remain separately supported. If pool support is required now,
reserve from the actual supplied pool and represent/return the transient backing
according to its death frontier. CTS convenience is not a production contract.

### G2. Keep `iree_status_t` for terminal errors

The private completion and async APIs use `IREE_STATUS_DEFERRED` to mean “another
object owns semaphore completion”:

- normal completion-batch submission returns `DEFERRED`;
- the worker and two direct-ready paths inspect it, discard it, and convert it
  to OK; and
- the nonblocking host-call wrapper manufactures it solely to prevent a second
  signal.

These are not suspended operations that the caller will resume. Completion
ownership is already known when the queue op is created. Store a typed
completion-policy enum on the op or select distinct callback paths; let callback
status report only actual failure. The existing public host-call callback's
asynchronous `DEFERRED` result is a narrow resumable API boundary and should be
translated there rather than generalized into the private queue protocol.

The cached dispatch path has a second misuse. Windows native argument update
always constructs `UNIMPLEMENTED`; two cache-rewrite paths inspect and discard
that status to decide to rebuild. That places status/diagnostic creation on an
affected cached-dispatch hot path to rediscover a static backend capability.
Publish the capability or select an update/rebuild strategy at device
initialization. A failure from an advertised operation must propagate.

Two additional sites collapse errors into ordinary state:

- semaphore readiness consumes a query failure and returns false; return status
  plus readiness so a failed semaphore is not merely “pending”;
- the Windows context-blob `FailStatus` helper discards detailed status into a
  generic boolean error; use a no-status primitive or preserve the diagnostic.

Clang-tidy also exposes a concrete failure-path bug at
`direct_command_buffer.c:2108`. Command creation/setup is status-gated, but the
dispatch-model `if / else if / else` includes the status predicate in its first
two branches. On failure it reaches the unconditional `else`, overwrites the
original status, and calls `add_arg_64` through a command that may be null. Nest
the semantic classifier inside one success gate.

The driver contains 24 production `iree_status_ignore` calls. Later native
completion errors should be joined into the terminal batch status; teardown
paths need no-status/raw cleanup primitives with asserted invariants rather than
manufacturing errors that cannot escape. The three clang-tidy lifetime findings
and three discarded-status findings must all be resolved, not suppressed.

### G3. Make exception-disabled allocation genuinely fallible

The shim is built with `-fno-exceptions` and described as exception-free, but
status/errno-returning construction paths use raw `new`, `std::make_unique`,
allocating vectors/strings, `push_back`, and `emplace_back`. Allocation failure
there terminates the process instead of returning `RESOURCE_EXHAUSTED`, and the
allocations bypass the supplied `iree_allocator_t`.

`return_start_npu_command_to_pool` can even grow a vector while holding the
command-pool mutex on dispatch completion. The pool is logically bounded at 64
entries, so this should be fixed-capacity and allocation-free after device
initialization.

Use the device host allocator with checked placement construction for C-visible
wrappers. Use allocator-aware status-returning storage or bounded fixed arrays
for fallible runtime state. Add fault-injection for device, buffer, context,
context-ref, and command creation, including cleanup after the native resource
already exists. The claimed exception-free boundary is established only when no
potentially throwing operation remains reachable.

## H. Establish a reproducible third-party ABI boundary

The Linux shim README says `amdxdna_accel.h` and `ert.h` are verbatim copies of
kernel/XRT sources that must byte-match those roots, but it does not identify a
commit, tag, release, or blob. The files are not verbatim copies of the named
upstreams.

`amdxdna_accel.h` matches blob
`67a0485d477fccb2832aa74e8be20a0a1afeaa91` from
`nod-ai/iree-amd-aie`. Relative to the referenced XDNA driver snapshot
`929e8ab459cab5915631849b9f1ef9a4982d1c11`, it includes representation/ABI
changes including a trailing `pad` that changes structure size/ioctl encoding
and a flexible-array-to-one-element-array conversion. `ert.h` is also a
downstream/reformatted copy and includes a semantic preemption fix relative to
XRT `4bc6a3f1673d3b9e2a856705d0258f0dec88047b`.

Those changes may be correct for a particular supported kernel/firmware package,
but the repository does not currently say which one. A hardware pass against
one installed image does not establish compatibility with the upstream source
named by the README.

Please establish this import shape:

1. Pin repository, commit, path, and source blob for every imported kernel/XRT
   file and name the supported driver/firmware package.
2. Put verbatim imported source, licenses, a source manifest, and separately
   reviewable local patches under `third_party/`.
3. Keep only the small first-party native adapter under the HAL driver.
4. Add compile-time sizes, offsets, and ioctl-number assertions for every ABI
   structure/request sent by the shim.
5. Record the supported driver package in the hardware lane so CI proves the
   intended matrix.

This is not a licensing conclusion. It is an auditability and maintenance
requirement: locally modified derivative code cannot be safely resynchronized
when its source and deviations are not reproducible, and putting it in the
first-party runtime tree obscures that ownership.

## I. Split files by invariant ownership, not mechanically by size

`direct_command_buffer.c` is 2,728 lines, `native_windows_mcdm.cc` is 4,045,
`kmt_api.cc` is 1,902, and `device.c` is 1,863. The problem is not a line-count
limit by itself; these files interleave queue issue, completion, cache mutation,
fallback planning, executable-context identity, Path-B ownership, native ABI,
and teardown. That makes it difficult to prove any one lifecycle without
holding several unrelated state machines in mind.

Split along invariant clusters such as:

- queue wait registration/issue;
- completion ownership and frontier advancement;
- dispatch validation/planning/patching;
- executable context identity and cache keys;
- Windows Path-B session/aperture ownership; and
- KMT module/symbol ownership.

`device_internal.h` is only an umbrella and `executable_internal.h` owns a core
representation despite its generic name. Replace those with component-scoped
declaration/implementation/test groupings. Do not move the same tangled private
state into a generic `internal.h`; the goal is one obvious owner for each
representation and failure path.

Both native submission waits also accept a timeout they ignore and then block.
Every current production caller passes `UINT64_MAX`, so no finite-wait regression
has been demonstrated, but the native API advertises a contract it cannot honor.
Remove the parameter until a backend implements it, or implement and test the
deadline semantics.

## Required regression and hardware coverage

The revised driver should have no-hardware unit/sanitizer tests for:

- truthful allocator compatibility and resolved memory types;
- exact timeline producer selection;
- already-satisfied and asynchronously satisfied multi-wait registration;
- all supported/unsupported dispatch config and flag combinations;
- PDI/context identity and concurrent recording;
- Windows cache-key completeness through a fake native creator;
- checked address patching at every boundary;
- context/cache allocation fault injection;
- cache initialization under TSAN;
- KMT module load ownership; and
- status/failure joins, including a later native completion failure.

The hardware lane must instantiate more than allocator/mapping/driver CTS. It
needs production executables and the queue, semaphore, command-buffer, dispatch,
and—once genuinely implemented—pool suites. Tests must express ordering with
semaphores rather than submission order, and valid async work must use explicit
completion rather than arbitrary short wall-clock deadlines.

Linux KMQ and Windows MCDM both need real visibility tests. Windows additionally
needs the two-device/context Path-B stress test. The mandatory Windows MSVC lane
should compile and run host tests; the broader clang-cl lane can carry sanitizer
fanout.

After the build metadata and dependencies are fixed, the focused local commands
are:

```bash
  --//runtime/config/hal:drivers=amdxdna \
  //runtime/src/iree/hal/drivers/amdxdna/...

build_tools/bin/iree-bazel-test --config=tsan \
  --//runtime/config/hal:drivers=amdxdna \
  //runtime/src/iree/hal/drivers/amdxdna/...
```

Run the five standalone probes from the companion appendix before and after the
fixes; each should become a checked-in regression at the owning component.

## Evidence and expected next shape

This review was performed against live head
`835d2ab4778665091468fa62786c2eeb30292bb2` and the exact current-base merge
`ead27d348368284175de8c8688c30d9cbc974332`.

Observed locally:

- the normal configure, strict deps, generated CMake, formatting, LSAN, and
  real-MSVC failures above;
- 13 remaining host tests passed under ASAN after excluding the four strict-dep
  build failures; hardware was unavailable on the review host;
- all 19 built host tests passed under MSVC after the C++17-only test repair,
  and five hardware CTS executables compiled;
- allocator metadata and malformed address patching reproduced without
  hardware;
- semaphore-producer hijacking and the queue-op UAF reproduced under ASAN;
- both cache-publication races reproduced under TSAN; and
- correctly configured clang-tidy produced the 114 diagnostics above.

The branch's gfx1151 result is useful evidence for the limited allocator,
mapping, and driver tests it runs. It is not evidence for queue/dispatch paths
that the backend does not instantiate.

For the next revision, please start by making the supported feature set narrow
and explicit, eliminating partial initialization and status-shaped state, and
moving the native ABI import under `third_party/`. Then add the production-shaped
tests for each advertised operation. That gives us a reviewable driver core on
which broader dispatch, pooling, and caching can be built without preserving
unsafe intermediate contracts.

benvanik:
PR37 reproducer appendix for the change request above. These five probes were
run against head `835d2ab4778665091468fa62786c2eeb30292bb2` merged onto the
reviewed current base as `ead27d348368284175de8c8688c30d9cbc974332`. None
requires XDNA hardware.

Run each command from the repository root. The sources deliberately use public
or driver-leaf interfaces only; they do not modify the branch. Each reproducer
should become a checked-in sanitizer/unit regression at the component that owns
the repaired invariant.

## 1. Allocator metadata advertises false coherency and locality

Save as `/tmp/pr37_memory_type_contract_probe.c`:

<details>
<summary>Allocator contract probe source</summary>

```c
#include <stdio.h>

#include "iree/base/api.h"
#include "iree/hal/api.h"
#include "iree/hal/drivers/amdxdna/allocator.h"

int main(void) {
  iree_hal_allocator_t* allocator = NULL;
  iree_status_t status = iree_hal_amdxdna_allocator_create(
      iree_allocator_system(), /*native_device=*/NULL, &allocator);
  if (!iree_status_is_ok(status)) {
    iree_status_fprint(stderr, status);
    iree_status_free(status);
    return 1;
  }

  iree_hal_buffer_params_t params = {
      .type = IREE_HAL_MEMORY_TYPE_DEVICE_LOCAL,
      .access = IREE_HAL_MEMORY_ACCESS_ALL,
      .usage = IREE_HAL_BUFFER_USAGE_DISPATCH_STORAGE |
               IREE_HAL_BUFFER_USAGE_MAPPING_PERSISTENT,
  };
  iree_hal_buffer_params_t resolved_params = {0};
  iree_device_size_t resolved_size = 0;
  iree_hal_buffer_compatibility_t compatibility =
      iree_hal_allocator_query_buffer_compatibility(
          allocator, params, 4096, &resolved_params, &resolved_size);
  printf("compatibility=0x%08x type=0x%08x host_visible=%d "
         "host_coherent=%d host_cached=%d size=%llu\n",
         compatibility, resolved_params.type,
         iree_all_bits_set(resolved_params.type,
                           IREE_HAL_MEMORY_TYPE_HOST_VISIBLE),
         iree_all_bits_set(resolved_params.type,
                           IREE_HAL_MEMORY_TYPE_HOST_COHERENT),
         iree_all_bits_set(resolved_params.type,
                           IREE_HAL_MEMORY_TYPE_HOST_CACHED),
         (unsigned long long)resolved_size);

  const bool falsely_coherent = iree_all_bits_set(
      resolved_params.type, IREE_HAL_MEMORY_TYPE_HOST_COHERENT);
  iree_hal_allocator_release(allocator);
  return falsely_coherent ? 2 : 0;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr37_memory_type_contract_probe.c \
  --config=asan \
  --//runtime/config/hal:drivers=amdxdna \
  --dep //runtime/src/iree/hal/drivers/amdxdna:amdxdna --
```

The reviewed branch prints:

```text
compatibility=0x00000801 type=0x00000076 host_visible=1 host_coherent=1 host_cached=0 size=4096
```

and exits 2. After repair, the selected `HOST_ONLY` heap must not report
`HOST_COHERENT`, `HOST_LOCAL`, or `DEVICE_LOCAL`; the exact compatible bits
must match the backend's implemented visibility contract.

## 2. A later timeline producer hijacks an earlier wait

Save as `/tmp/pr37_native_signal_producer_probe.cc`:

<details>
<summary>Timeline producer probe source</summary>

```cpp
#include <atomic>
#include <chrono>
#include <cstdio>
#include <thread>

#include "iree/async/proactor_platform.h"
#include "iree/base/api.h"
#include "iree/hal/api.h"
#include "iree/hal/drivers/amdxdna/completion_queue.h"
#include "iree/hal/drivers/amdxdna/semaphore.h"

static bool CheckStatus(const char* operation, iree_status_t status) {
  if (iree_status_is_ok(status)) return true;
  std::fprintf(stderr, "%s failed: ", operation);
  iree_status_fprint(stderr, status);
  std::fprintf(stderr, "\n");
  iree_status_free(status);
  return false;
}

int main() {
  iree_async_proactor_t* proactor = nullptr;
  if (!CheckStatus("create proactor", iree_async_proactor_create_platform(
                                          iree_async_proactor_options_default(),
                                          iree_allocator_system(), &proactor))) {
    return 1;
  }

  iree_hal_semaphore_t* semaphore = nullptr;
  if (!CheckStatus(
          "create semaphore",
          iree_hal_amdxdna_semaphore_create(
              proactor, IREE_HAL_QUEUE_AFFINITY_ANY, 0,
              IREE_HAL_SEMAPHORE_FLAG_DEFAULT, iree_allocator_system(),
              &semaphore))) {
    iree_async_proactor_release(proactor);
    return 1;
  }

  iree_hal_amdxdna_completion_queue_t* queue = nullptr;
  if (!CheckStatus("create completion queue",
                   iree_hal_amdxdna_completion_queue_create(
                       iree_allocator_system(), &queue))) {
    iree_hal_semaphore_release(semaphore);
    iree_async_proactor_release(proactor);
    return 1;
  }

  iree_hal_semaphore_t* semaphores[] = {semaphore};
  uint64_t first_value[] = {1};
  uint64_t later_value[] = {2};
  iree_hal_semaphore_list_t first_signal_list = {
      1, semaphores, first_value};
  iree_hal_semaphore_list_t later_signal_list = {
      1, semaphores, later_value};

  iree_hal_amdxdna_completion_batch_t* first_batch = nullptr;
  iree_hal_amdxdna_completion_batch_t* later_batch = nullptr;
  if (!CheckStatus("create first batch",
                   iree_hal_amdxdna_completion_batch_create(
                       queue, first_signal_list, &first_batch)) ||
      !CheckStatus("create later batch",
                   iree_hal_amdxdna_completion_batch_create(
                       queue, later_signal_list, &later_batch))) {
    iree_hal_amdxdna_completion_batch_destroy(first_batch);
    iree_hal_amdxdna_completion_batch_destroy(later_batch);
    iree_hal_amdxdna_completion_queue_destroy(queue);
    iree_hal_semaphore_release(semaphore);
    iree_async_proactor_release(proactor);
    return 1;
  }

  // Publish in queue order. The later producer replaces the semaphore's one
  // producer slot even though the first producer still owns value 1.
  iree_hal_amdxdna_completion_batch_publish_signals(first_batch);
  iree_hal_amdxdna_completion_batch_publish_signals(later_batch);

  std::atomic<bool> waiter_entered{false};
  std::atomic<bool> waiter_returned{false};
  std::atomic<iree_status_code_t> waiter_code{IREE_STATUS_UNKNOWN};
  std::thread waiter([&] {
    waiter_entered.store(true, std::memory_order_release);
    iree_status_t status = iree_hal_semaphore_wait(
        semaphore, 1, iree_infinite_timeout(), IREE_ASYNC_WAIT_FLAG_NONE);
    waiter_code.store(iree_status_code(status), std::memory_order_release);
    iree_status_free(status);
    waiter_returned.store(true, std::memory_order_release);
  });

  while (!waiter_entered.load(std::memory_order_acquire)) {
    std::this_thread::yield();
  }
  // Diagnostic probe only: give the waiter time to enter the batch shortcut.
  std::this_thread::sleep_for(std::chrono::milliseconds(100));

  // Retain a diagnostic handle because submit transfers the original owner to
  // the completion queue.
  iree_hal_amdxdna_completion_batch_retain(first_batch);
  iree_status_t submit_status =
      iree_hal_amdxdna_completion_batch_submit(first_batch);
  if (!iree_status_is_deferred(submit_status)) {
    CheckStatus("submit first batch", submit_status);
    return 1;
  }
  iree_status_free(submit_status);
  if (!CheckStatus("wait first batch",
                   iree_hal_amdxdna_completion_batch_wait(
                       first_batch, iree_infinite_timeout(),
                       IREE_ASYNC_WAIT_FLAG_NONE))) {
    return 1;
  }
  iree_hal_amdxdna_completion_batch_destroy(first_batch);

  uint64_t current_value = 0;
  if (!CheckStatus("query value 1",
                   iree_hal_semaphore_query(semaphore, &current_value))) {
    return 1;
  }
  const bool incorrectly_waiting_for_later_batch =
      current_value == 1 &&
      !waiter_returned.load(std::memory_order_acquire);
  std::printf("after value 1 completed: semaphore=%llu waiter_returned=%d\n",
              static_cast<unsigned long long>(current_value),
              waiter_returned.load(std::memory_order_acquire));

  submit_status = iree_hal_amdxdna_completion_batch_submit(later_batch);
  if (!iree_status_is_deferred(submit_status)) {
    CheckStatus("submit later batch", submit_status);
    return 1;
  }
  iree_status_free(submit_status);
  waiter.join();

  iree_hal_amdxdna_completion_queue_destroy(queue);
  iree_hal_semaphore_release(semaphore);
  iree_async_proactor_release(proactor);

  std::printf("waiter status=%d hijacked=%d\n",
              static_cast<int>(waiter_code.load(std::memory_order_acquire)),
              incorrectly_waiting_for_later_batch);
  return incorrectly_waiting_for_later_batch ? 2 : 0;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr37_native_signal_producer_probe.cc \
  --config=asan \
  --//runtime/config/hal:drivers=amdxdna \
  --dep //runtime/src/iree/hal/drivers/amdxdna:amdxdna \
  --dep //runtime/src/iree/async:platform --
```

The reviewed branch exits 2 after reporting:

```text
after value 1 completed: semaphore=1 waiter_returned=0
waiter status=0 hijacked=1
```

After repair, the wait for value 1 must return when the value-1 producer
completes; publishing value 2 must not redirect it to the later batch. This
probe necessarily drives the branch's current `DEFERRED` submission protocol;
the checked-in post-fix regression should use the typed completion policy
requested in the main review.

## 3. An already-satisfied wait frees its queue op during registration

Save as `/tmp/pr37_async_queue_satisfied_wait_probe.cc`:

<details>
<summary>Synchronous callback lifetime probe source</summary>

```cpp
#include <cstdio>

#include "iree/async/proactor_platform.h"
#include "iree/async/semaphore.h"
#include "iree/base/api.h"
#include "iree/base/internal/arena.h"
#include "iree/hal/drivers/amdxdna/async_queue.h"

int main() {
  iree_async_proactor_t* proactor = nullptr;
  IREE_CHECK_OK(iree_async_proactor_create_platform(
      iree_async_proactor_options_default(), iree_allocator_system(),
      &proactor));

  iree_async_semaphore_t* wait_semaphore = nullptr;
  IREE_CHECK_OK(iree_async_semaphore_create(
      proactor, /*initial_value=*/1,
      IREE_ASYNC_SEMAPHORE_DEFAULT_FRONTIER_CAPACITY,
      iree_allocator_system(), &wait_semaphore));

  iree_arena_block_pool_t block_pool;
  // Keep the reusable block smaller than the async op so the op is an
  // oversized arena allocation. ASAN then poisons it immediately when the
  // worker releases the op instead of pooled storage hiding the bug.
  iree_arena_block_pool_initialize(/*total_block_size=*/128,
                                   iree_allocator_system(), &block_pool);
  iree_hal_amdxdna_async_queue_t* queue = nullptr;
  IREE_CHECK_OK(iree_hal_amdxdna_async_queue_create(
      &block_pool, iree_allocator_system(), &queue));

  iree_hal_semaphore_t* wait_semaphores[] = {
      reinterpret_cast<iree_hal_semaphore_t*>(wait_semaphore)};
  uint64_t wait_values[] = {1};
  const iree_hal_semaphore_list_t wait_list = {
      /*count=*/1, wait_semaphores, wait_values};
  for (int i = 0; i < 1000000; ++i) {
    IREE_CHECK_OK(iree_hal_amdxdna_async_queue_enqueue(
        queue, wait_list, iree_hal_semaphore_list_empty(),
        /*op_fn=*/nullptr, /*cleanup_fn=*/nullptr, /*user_data=*/nullptr,
        /*retained_resources=*/nullptr, /*retained_resource_count=*/0));
  }

  iree_hal_amdxdna_async_queue_destroy(queue);
  iree_arena_block_pool_deinitialize(&block_pool);
  iree_async_semaphore_release(wait_semaphore);
  iree_async_proactor_release(proactor);
  std::puts("completed");
  return 0;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr37_async_queue_satisfied_wait_probe.cc \
  --config=asan \
  --//runtime/config/hal:drivers=amdxdna \
  --dep //runtime/src/iree/hal/drivers/amdxdna:amdxdna \
  --dep //runtime/src/iree/async:platform --
```

The reviewed branch terminates with an ASAN heap-use-after-free reading
`op->wait_list.count` at `async_queue.c:614` after the async worker releases
the op arena. After repair, this must complete cleanly under ASAN. The checked-in
test should preserve the oversized-arena storage shape so arena pooling cannot
hide the poison.

## 4. Descriptor patching accepts and mutates malformed addresses

Save as `/tmp/pr37_patch_table_contract_probe.cc`:

<details>
<summary>Patch-table contract probe source</summary>

```cpp
#include <cinttypes>
#include <cstdio>

#include "iree/hal/drivers/amdxdna/direct_command_buffer_planning.h"

int main() {
  const uint32_t patches[] = {/*offset=*/0, /*arg_index=*/0,
                              /*arg_plus=*/3};

  uint32_t misaligned_control[4] = {};
  const uint64_t ordinary_argument[] = {0x1000};
  const bool misaligned_ok = iree_hal_amdxdna_apply_patch_table(
      misaligned_control, 4, patches, 3, ordinary_argument, 1);
  std::printf("misaligned accepted=%d bd2:bd1=%08" PRIx32 ":%08" PRIx32
              "\n",
              misaligned_ok, misaligned_control[2], misaligned_control[1]);

  uint32_t overflow_control[4] = {};
  const uint64_t overflowing_argument[] = {0x0000FFFFFFFFFFF0ull};
  const bool overflow_ok = iree_hal_amdxdna_apply_patch_table(
      overflow_control, 4, patches, 3, overflowing_argument, 1);
  std::printf("overflow accepted=%d bd2:bd1=%08" PRIx32 ":%08" PRIx32
              "\n",
              overflow_ok, overflow_control[2], overflow_control[1]);
  return 0;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr37_patch_table_contract_probe.cc \
  --config=asan \
  --//runtime/config/hal:drivers=amdxdna \
  --dep //runtime/src/iree/hal/drivers/amdxdna:amdxdna --
```

The reviewed branch reports both malformed patches as accepted:

```text
misaligned accepted=1 bd2:bd1=00000000:80001000
overflow accepted=1 bd2:bd1=00010000:7ffffff0
```

After repair, both calls must reject their input without modifying the control
packet. The owning test also needs binding extents so it can reject an addend
outside the submitted binding; this standalone helper signature cannot test
that missing information until the production API is corrected.

## 5. Lazy command-cache publication has two C data races

Save as `/tmp/pr37_command_cache_init_race_probe.cc`:

<details>
<summary>Command-cache TSAN probe source</summary>

```cpp
#include <atomic>
#include <thread>
#include <vector>

#include "iree/base/api.h"
#include "iree/base/threading/mutex.h"
#include "iree/hal/drivers/amdxdna/device.h"
#include "iree/hal/drivers/amdxdna/direct_command_buffer.h"
#include "iree/hal/drivers/amdxdna/direct_command_buffer_chain_cache.h"
#include "iree/hal/drivers/amdxdna/direct_command_buffer_single_cache.h"

int main() {
  iree_hal_amdxdna_device device = {};
  device.host_allocator = iree_allocator_system();
  iree_slim_mutex_initialize(&device.command_cache_mutex);

  std::atomic<bool> start = false;
  std::vector<std::thread> threads;
  for (int i = 0; i < 16; ++i) {
    threads.emplace_back([&, i]() {
      while (!start.load(std::memory_order_acquire)) {
      }
      for (int j = 0; j < 1000; ++j) {
        if (i & 1) {
          (void)iree_hal_amdxdna_get_single_command_cache(&device);
        } else {
          (void)iree_hal_amdxdna_get_chain_command_cache(&device);
        }
      }
    });
  }
  start.store(true, std::memory_order_release);
  for (auto& thread : threads) thread.join();

  iree_hal_amdxdna_device_destroy_single_command_cache(&device);
  iree_hal_amdxdna_device_destroy_chain_command_cache(&device);
  iree_slim_mutex_deinitialize(&device.command_cache_mutex);
  return 0;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr37_command_cache_init_race_probe.cc \
  --config=tsan \
  --//runtime/config/hal:drivers=amdxdna \
  --dep //runtime/src/iree/hal/drivers/amdxdna:amdxdna --
```

The reviewed branch produces one TSAN report for each ordinary pointer
publication:

```text
Write in iree_hal_amdxdna_chain_command_cache_create
Previous read in iree_hal_amdxdna_get_chain_command_cache

Write in iree_hal_amdxdna_single_command_cache_create
Previous read in iree_hal_amdxdna_get_single_command_cache
```

The preferred repair is eager, status-returning initialization before device
publication, which removes both races and the lazy dispatch-path branches. The
checked-in test should exercise concurrent public dispatch after device
creation rather than preserve these private getters if they disappear in that
design.

## Completion criteria for these probes

A local patch is not complete merely because it changes the output above. Each
fix needs a production-shaped regression at the owning abstraction, sanitizer
coverage where lifetime/concurrency is involved, and an error path that leaves
the object in one documented state. In particular:

- allocator tests must cover Linux KMQ and Windows MCDM visibility, not only
  metadata bits;
- the semaphore test must add a real dependency after value 1 so the hijack
  cannot be hidden by promptly completing value 2;
- the queue registration test must cover several waits with both synchronous
  and asynchronous callback delivery;
- patch tests must prove rejection is non-mutating and include binding bounds;
  and
- cache tests must prove allocation failure is returned during device creation
  as well as running cleanly under TSAN.

After integrating the fixes, run the repository gate from the main review in
addition to these focused tests:

```bash
```

benvanik:
The three MCDM commits pushed after the reviewed head appear to have crossed with the detailed change request, so I am treating them as work that was already in flight rather than as an attempted response to that review.

Please pause further feature, protocol-version, and performance work on this branch until the existing change request has been absorbed and the correctness foundation has been repaired. The new negotiation work has some useful direction — in particular, querying an explicit identity, failing closed on unknown identities, centralizing packet-layout data, and adding byte-level tests — but continuing to add protocol variants on top of the current ownership and failure model is compounding the review surface faster than the underlying invariants are becoming sound.

The five deterministic reproducers from the review remain current because their owning allocator, semaphore, async-queue, descriptor-patching, and device-cache source files are unchanged. The false memory metadata, later-producer timeline hijack, already-satisfied-wait use-after-free, malformed-address mutation, and lazy cache-publication races therefore still need the production-shaped repairs and checked-in regressions described there. The status-as-control-flow, ignored-error, dispatch-semantics, context-identity, failure-path, and ABI-provenance findings also remain open.

The new MCDM code adds three concerns that need to be resolved before more variants are stacked onto it.

## Preserve the negotiated protocol revision

`QueryMcdmAbi` now recognizes legacy identities `{0, 2}` and `{0, 3}` and compact identities `{0, 2, 0}` and `{0, 3, 0}`, but immediately reduces each pair to only `McdmAbi::legacy` or `McdmAbi::compact`. `GetMcdmAbiInfo` consequently cannot distinguish revision 2 from revision 3. Accepting revision 3 therefore asserts that every private-packet size, field offset, flag, chain-handle representation, status-object policy, aperture ownership transition, cache-publication rule, and remapping behavior is identical to revision 2.

The new fake-query test proves that `{0, 3, 0}` is accepted; it does not prove that the revision-3 ABI is compatible with the revision-2 layout table. Please retain the negotiated `{family, revision}` in the representation and select an explicit reviewed contract for each supported pair. If the two revisions are intentionally identical, preserve the source/capture evidence for that conclusion and add real revision-3 Windows hardware coverage for setup, single and chain submission, code-range ownership, cache publication, context switching, completion, and teardown.

## Make aperture activation an owned transaction

The new `AcquirePathBCodeRange` and `ReleasePathBCodeRange` helpers do not repair the process-global aperture race from the review. They submit miniport opcode-9 ownership/cache markers; they do not provide host-side exclusion or retain the active context.

`activate_pathb_context_for_submit` still releases `g_pathb_aperture_mutex` before the caller inspects or writes the aperture, stages code and descriptors, and submits the KMT command. Another context can enter activation, retire the first context, and call `ReleaseCommandApertureGpuMapping` while the first thread is still using that mapping. That release also writes `gpu_va` and `code_gpu_va` while the first path reads them without synchronization. Adding more range and cache states around this does not make the ownership valid; it creates more state whose correctness depends on the missing lease.

Please represent activation as an explicit ownership transaction covering every aperture-dependent host operation through KMT submission. Whether that ownership must extend through device completion must be derived from and documented against the actual miniport mapping-lifetime contract. A process-global raw context pointer plus short critical sections cannot encode that lifetime.

The KMT module lifetime issue is unchanged and should be fixed as part of the same Windows ownership cleanup: 23 `ResolveKmtProc` calls each execute three `LoadLibraryW` calls, retaining 69 unowned module references per `KmtApi::Load` with default DLL search behavior. One explicitly owned dispatch object should load each system module once with a system-directory search policy and hold or release those modules through a deliberate lifetime.

## Establish an independent ABI oracle

The compact path now constructs a 0x120-byte `CompactPathBChainHandleV1` containing allocation handles, page counts, GPU and CPU addresses, and fixed observed state values, alongside additional magic private types, policies, sizes, and offsets. The comments describe this as an observed stable prefix derived from XRT behavior. Tests that construct expected packets from the same handwritten assumptions are useful regression tests, but they are not independent evidence that the private DDI contract is correct.

This makes the existing third-party/native-ABI provenance requirement more important. Preserve the exact source or capture provenance, supported driver identities, independent packet fixtures or captures, and real Windows hardware results at a reproducible ABI boundary. Otherwise each new driver revision multiplies a collection of unversioned assumptions that only agree with their own tests.

## Restore the mechanical review gate before the next revision

The current presubmit reports 79 clang-format diagnostics in the changed Windows files and generated CMake drift. The branch also conflicts with current `main` and therefore has not exercised the newly mandatory native-MSVC baseline. Before requesting another deep review, please:

1. Rebase onto current `main`, preserving its MSVC and no-HIP core contracts.
2. Install the paranoid hook and run the complete profile:

   ```bash
   ```

3. Address the existing correctness findings and turn the supplied reproducers into owning regression tests.
4. Split the Windows implementation by invariant ownership before adding more protocol machinery. `native_windows_mcdm.cc` is still approximately 4,000 lines, while `kmt_api.cc` has grown past 2,400 lines and currently owns module loading, discovery, negotiation, packet encoding, allocation, residency, context lifetime, aperture mapping, cache operations, synchronization, submission, completion, and teardown.

Pausing here is intended to save work, not discard it. Once the memory, synchronization, failure, ownership, and ABI boundaries are explicit, the negotiated MCDM work can be reviewed and extended without every new driver detail depending on unresolved process-global state and reverse-engineered assumptions.

PR #44, opened 2026-06-08
Fix HIP graph launch sequencing

AWoloszyn:
Graph launches could hang or mis-order repeated launches because stream sequencing was based on completed values instead of the last submitted stream work, and graph memset nodes were represented as host callbacks.

This changes graph launch submission to chain from the stream pending value and keep graph semaphore bases monotonic across repeated launches. It also records HIP graph memset nodes as native graph memset nodes, avoiding a host callback in the execution path.

### DIFF
--- a/libhrx/src/binding/common/graph_exec.c
+++ b/libhrx/src/binding/common/graph_exec.c
@@ -1009,21 +1009,39 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(z0,
                                     iree_hal_streaming_stream_flush(stream));
 
-  // Query current values of internal semaphores for delta encoding.
-  // Strategy A: Reuse semaphores - query current values.
-  for (uint32_t i = 0; i < exec->semaphore_count; i++) {
-    IREE_RETURN_AND_END_ZONE_IF_ERROR(
-        z0, iree_hal_semaphore_query(exec->semaphores[i],
-                                     &exec->semaphore_base_values[i]));
-  }
-
-  // Mutex needed for launch per CUDA docs.
+  // Mutex needed for launch per CUDA docs. It also protects the executable
+  // semaphore base values below, which are reused and advanced across launches
+  // of the same executable graph.
   iree_slim_mutex_lock(&exec->mutex);
 
-  // Save the stream's initial state.
-  uint64_t stream_wait_value = stream->completed_value;
-  ++stream->pending_value;
-  uint64_t stream_signal_value = stream->pending_value;
+  // Each graph block stores semaphore payload offsets relative to these base
+  // values. Refresh the bases from the device, but never move them backward.
+  // Back-to-back graph launches can submit a later launch before an earlier one
+  // has completed; in that case the queried device value may still be behind
+  // the last payload value this exec submitted.
+  for (uint32_t i = 0; i < exec->semaphore_count; i++) {
+    uint64_t current_value = 0;
+    iree_status_t status =
+        iree_hal_semaphore_query(exec->semaphores[i], &current_value);
+    if (!iree_status_is_ok(status)) {
+      iree_slim_mutex_unlock(&exec->mutex);
+      IREE_TRACE_ZONE_END(z0);
+      return status;
+    }
+    exec->semaphore_base_values[i] =
+        iree_max(exec->semaphore_base_values[i], current_value);
+  }
+
+  iree_slim_mutex_lock(&stream->mutex);
+
+  // Reserve the next stream timeline value while holding the stream lock so
+  // concurrent host threads cannot submit same-stream work with the same wait
+  // or signal value. The graph waits on the current stream tail, not the last
+  // completion observed by the host; HIP stream ordering requires repeated
+  // graph launches on the same stream to execute FIFO even when the caller does
+  // not synchronize between launches.
+  uint64_t stream_wait_value = stream->pending_value;
+  uint64_t stream_signal_value = stream_wait_value + 1;
 
   // Submit all blocks. They will synchronize with each other via internal
   // semaphores. This allows concurrent execution of independent blocks.
@@ -1194,7 +1212,12 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
     memcpy(exec->semaphore_base_values, new_base_values,
            exec->semaphore_count * sizeof(uint64_t));
   }
+  if (iree_status_is_ok(status)) {
+    stream->pending_value = stream_signal_value;
+    stream->submitted_value = stream_signal_value;
+  }
 
+  iree_slim_mutex_unlock(&stream->mutex);
   iree_slim_mutex_unlock(&exec->mutex);
   IREE_TRACE_ZONE_END(z0);
   return status;
--- a/libhrx/src/binding/common/graph_exec.c
+++ b/libhrx/src/binding/common/graph_exec.c
@@ -1009,9 +1009,9 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(z0,
                                     iree_hal_streaming_stream_flush(stream));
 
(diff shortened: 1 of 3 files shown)


libhrx/src/binding/hip/api.c
```diff
@@ -10611,31 +10572,12 @@ HIPAPI hipError_t hipGraphAddMemsetNode(hipGraphNode_t* pGraphNode,
...
-  callback_data->count = params->width * params->height;
-  callback_data->context = stream_graph->context;
-
-  // Create a host callback node that performs the memset.
   HIP_RETURN_STATUS_AND_END_ZONE_IF_ERROR(
       z0,
-      iree_hal_streaming_graph_add_host_call_node(
-          stream_graph, deps, numDependencies, iree_hip_graph_memset_callback,
-          callback_data, &node),
+      iree_hal_streaming_graph_add_memset_node(
```
zjgarvey, line 10577: Is this new or something? I definitely love the +4 / -62

libhrx/src/binding/common/graph_exec.c
```diff
@@ -1009,21 +1009,28 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
...
+    uint64_t current_value = 0;
     IREE_RETURN_AND_END_ZONE_IF_ERROR(
-        z0, iree_hal_semaphore_query(exec->semaphores[i],
-                                     &exec->semaphore_base_values[i]));
+        z0, iree_hal_semaphore_query(exec->semaphores[i], &current_value));
+    exec->semaphore_base_values[i] =
+        iree_max(exec->semaphore_base_values[i], current_value);
   }
 
   // Mutex needed for launch per CUDA docs.
```
zjgarvey: This is a weird comment, but unrelated to the changes here, so feel free to ignore. 

Just something I noticed when reading the surrounding context.

libhrx/src/binding/common/graph_exec.c
```diff
@@ -1009,21 +1009,28 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(z0,
                                     iree_hal_streaming_stream_flush(stream));
 
-  // Query current values of internal semaphores for delta encoding.
-  // Strategy A: Reuse semaphores - query current values.
+  // Query current values of internal semaphores for delta encoding, but never
```
zjgarvey: What exactly does "delta encoding" mean here?

I think I understand the change/bugfix here (IIUC, we should never decrease semaphore values in general).

libhrx/src/binding/common/graph_exec.c
zjgarvey: And now that I think about it, would it make more sense to put the mutex lock before writing to the exec semaphore base values? I need to look into the code below a bit more to get an idea what these semaphores are doing there.

libhrx/src/binding/common/graph_exec.c
```diff
@@ -1009,21 +1009,28 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
...
 
-  // Save the stream's initial state.
-  uint64_t stream_wait_value = stream->completed_value;
-  ++stream->pending_value;
-  uint64_t stream_signal_value = stream->pending_value;
+  // Chain this graph launch after the last submitted stream operation, not the
+  // last operation whose completion the host has observed. HIP stream ordering
+  // requires repeated graph launches on the same stream to execute FIFO even
+  // when the caller does not synchronize between launches.
+  uint64_t stream_wait_value = stream->pending_value;
```
zjgarvey, line 1043: I'm trying to sus out the different stream timepoints.

As-per the typedef for `iree_hal_streaming_stream_t` in `libhrx/src/binding/common/internal.h`:

```cpp
  // Semaphore chain for synchronization.
  iree_hal_semaphore_t* timeline_semaphore;
  uint64_t pending_value;    // Next value to be used for signaling
  uint64_t submitted_value;  // Last value that was actually submitted (for
                             // wait_submitted)
  uint64_t completed_value;  // Last value we've verified as completed
```

I'm not sure I follow this. Is the submitted value the one we want to wait on here? Even the comment here suggests we wait on the last submitted stream op:

```
  // Chain this graph launch after the last submitted stream operation, not the
  // last operation whose completion the host has observed. HIP stream ordering
```

Separately, how important is it that we wrap any of this stream access in a mutex? Could multiple threads be submitting work to the same stream, or are streams supposed to be host-thread-local?

zjgarvey (commented):
I still don't fully understand the internal semaphore structure for the graph blocks, but I'll put up a few questions for now.

libhrx/src/binding/hip/api.c
AWoloszyn: Yes, this is intentional. The old graph memset implementation encoded memset as a host callback that allocated a host pattern buffer and performed a synchronous host-to-device transfer. That is the wrong shape for graph execution because it puts blocking host work into the graph timeline. This change routes hipGraphAddMemsetNode through the native graph memset node path instead, so memset participates in the graph block scheduler like other device work.

libhrx/src/binding/common/graph_exec.c
AWoloszyn: Agreed. I moved exec->mutex before the base-value refresh. The base array is mutable per-executable launch state, so querying and updating it should be serialized with the rest of graph launch. The updated path unlocks and returns cleanly if a semaphore query fails before any stream timeline value is reserved.

libhrx/src/binding/common/graph_exec.c
AWoloszyn: Good call. “Delta encoding” was too vague. What it means here is that each graph block stores semaphore payload offsets relative to exec->semaphore_base_values. Repeated launches reuse the same internal semaphores, so each launch has to choose a base payload and then add the block-local offsets. I updated the comment to say that directly and to explain why the base is monotonic: if launch N+1 is submitted before launch N has completed, querying the device semaphore can return a value behind the last payload already submitted by this exec.

libhrx/src/binding/common/graph_exec.c
AWoloszyn: The old field comment made this more confusing than it needed to be. pending_value is the stream tail value already reserved for stream work, not the next value to signal. Ordinary stream flush waits on the current pending_value, then increments/signals the next value. Graph launch needs the same tail-reservation model: wait on the current stream tail, reserve tail + 1, and publish that as both pending/submitted after successful submission.

completed_value is definitely wrong for this path because it is only the last completion observed by the host. If two graph launches are submitted back-to-back without synchronization, waiting on completed_value can let the second launch bypass the first. submitted_value is useful for APIs that wait on already submitted work without flushing; graph launch is submitting new work, so it reserves from the stream tail while holding stream->mutex.



Yes, same-stream submission can come from multiple host threads, and the stream timeline fields are protected by stream->mutex in the normal flush path. I updated graph launch to reserve and publish its stream wait/signal values while holding stream->mutex, and to keep that lock through queue submission, matching the stream flush pattern. That prevents two host threads from observing the same stream tail and submitting graph launches with duplicate wait/signal values.

libhrx/src/binding/common/graph_exec.c
```diff
@@ -1009,21 +1009,39 @@ iree_status_t iree_hal_streaming_graph_exec_launch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(z0,
                                     iree_hal_streaming_stream_flush(stream));
 
-  // Query current values of internal semaphores for delta encoding.
-  // Strategy A: Reuse semaphores - query current values.
+  // Mutex needed for launch per CUDA docs. It also protects the executable
```
zjgarvey: I still don't get the "CUDA docs" reference. Are we relying on some CUDA software?

zjgarvey (approved):
I'm still confused about how the internal semaphore structure works overall (I.e., how the base values and semaphore values evolve during multiple launches being enqueued), but I don't want to block this PR. At least on the stream and async memset side of things, this makes sense to me.

PR #45, opened 2026-06-08
Fix HIP module load data bundle parsing

AWoloszyn:
hipModuleLoadData passes embedded images without a known byte length. The fat-binary parser accepted unknown-length inputs for wrapper detection, but the Clang offload bundle extractor still validated entries against data_length == 0 and rejected valid embedded gfx1100 kernels as out of range.

This treats zero-length bundle spans as unbounded for entry-table traversal and skips bounded offset validation only in that unknown-length case. Known-length bundles keep the existing range checks.


zjgarvey (approved):
LGTM

PR #46, opened 2026-06-08
Fix HIP primary context pool lifetime

AWoloszyn:
The full hip-cts device_test exposed an ordering-dependent failure where hipMemset and hipMemsetAsync returned 500 after primary-context retain/reset tests. The runtime could leave TLS pointing at a primary context while the device primary-context pointer was being cleared, and the global contiguous allocation pool was reused without checking which context owned it.

This clears TLS against the actual primary context before releasing it, tracks the owning context for the contiguous pool, resets stale pools when the owning context is released or replaced, and releases the pool wrapper's context retain during pool teardown.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/libhrx/src/binding/common/device.c
+++ b/libhrx/src/binding/common/device.c
@@ -410,9 +410,8 @@ iree_status_t iree_hal_streaming_device_release_primary_context(
     iree_hal_streaming_context_t* released_context = device->primary_context;
 
     // Wait for all operations to complete.
-    iree_status_t status =
-        iree_hal_streaming_context_wait_idle(released_context,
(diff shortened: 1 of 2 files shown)


zjgarvey (approved):
This looks good to me. I'll rebase and fix any conflicts with the per-device ordinal work in #47 after we land this.

zjgarvey:
Note: the 1201 segfaults on lint seem unrelated, so landed anyway.

PR #47, opened 2026-06-08
Fix streaming device-ordinal aliasing; per-device hipMalloc pool

zjgarvey:
## Summary

Fixes device-ordinal aliasing in the HRX streaming device init and reworks the HIP `hipMalloc` pool to be per-device. Two related fixes:

1. **Streaming device-ordinal aliasing** (`libhrx/src/binding/common/init.c`). `iree_hal_streaming_initialize_device` assigned the device ordinal *before* a `memset(out_device, 0, …)` that immediately cleared it, so every device aliased ordinal 0. Context-ordinal lookups on a non-default device returned 0 and per-context buffer resolution collapsed onto device 0. The ordinal is now threaded in and assigned *after* the memset.

2. **Per-device `hipMalloc` pool** (`libhrx/src/binding/hip/api.c`). Replaces the process-global contiguous bump pool with one pool per device ordinal, keyed by `context->device_ordinal`. A single global pool — created on whichever device was current first (device 0) and registered only in that context — made allocations on non-default devices unresolvable (`hipErrorNotFound`, seen via hipBLASLt on gfx942) and physically placed "device N" memory on device 0. Required for concurrent multi-GPU; depends on fix (1) for correct ordinals.

## Relationship to #46

Rebased on top of #46 ("Fix HIP primary context pool lifetime"). This PR **supersedes #46's process-global pool** with per-device pools, while:

- **carrying over #46's context-retain release on pool teardown** — the pool buffer wrapper retains its creating context, so the per-device `free_and_reset` now releases it to avoid leaking that context for the pool's lifetime; and
- **inheriting #46's `device.c` TLS-clear fix** unchanged (this PR does not touch `device.c`).

#46's global-pool owner tracking (`g_contiguous_pool_context`, `reset_contiguous_pool_if_owned_by`) is intentionally dropped — per-device pools don't need it.

## Deferred / follow-up

- **Pool sizing policy (75%-of-VRAM default):** kept as-is for now. The bump pool never reclaims, so it must be sized for the cumulative working set; a small default would OOM real models. The proper fix (lazy physical commit over a reserved contiguous VA range / HIP VMM) is a follow-up and mostly a design question for @AWoloszyn.

## Validation

- Native-HIP vs HRX A/B parity on the one-op + ResNet smoke suite (async, multi-ordinal 0/1/2/7).
- hip-cts non-default-ordinal repro (`hipGetDevice` on a non-default ordinal) to lock the ordinal fix against regression.
- Rebuild + revalidation and a self-review pass after today's rebase onto #46 are in progress.

### DIFF
--- a/libhrx/src/binding/hip/api.c
+++ b/libhrx/src/binding/hip/api.c
@@ -436,15 +436,15 @@ static hipError_t iree_hip_ensure_initialized(void) {
   return hipSuccess;
 }
 
+// Frees and resets the per-device hipMalloc pool (defined with the pool).
+static void iree_hip_free_and_reset_contiguous_pool(
+    iree_hal_streaming_context_t* context);
+
 // Ensures context exists for current thread and returns it.
 // This implements HIP's implicit initialization behavior:
 // - Automatically calls hipInit() if needed
 // - Creates primary context for device 0 if no context exists
 // - Sets the context as current for the thread
-// Frees and resets the per-device hipMalloc pool (defined with the pool).
-static void iree_hip_free_and_reset_contiguous_pool(
-    iree_hal_streaming_context_t* context);
-
 static hipError_t iree_hip_ensure_context(
     iree_hal_streaming_context_t** out_context) {
   // First ensure initialized.
@@ -3165,13 +3165,14 @@ static hrx_device_pool_t* iree_hip_pool_containing_locked(uintptr_t addr) {
 // buffer_tables are per-context but the pool is shared per-device; a context
 // that cannot resolve pool pointers fails hipMemset/hipMemcpy with
 // hipErrorNotFound. Called from hipMalloc only — never on hot paths.
-static void iree_hip_pool_register_in_context_locked(
+// Returns OK when the buffer is already registered; propagates real failures
+// (e.g. out-of-memory growing the table) so the caller can surface them.
+static hrx_status_t iree_hip_pool_register_in_context_locked(
     hrx_device_pool_t* p, iree_hal_streaming_context_t* context) {
-  if (!p->buffer) return;
-  hrx_status_t s = hrx_buffer_table_insert(
+  if (!p->buffer) return hrx_ok_status();
+  return hrx_buffer_table_insert_if_new(
       &context->buffer_table, p->buffer->device_ptr, p->buffer->host_ptr,
       p->buffer->size, p->buffer->hrx_buf, p->buffer);
-  hrx_status_ignore(s);  // ALREADY_EXISTS (already registered) is fine.
 }
 
 // Frees and resets the pool for |context|'s device (e.g. on hipDeviceReset).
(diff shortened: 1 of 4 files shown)


libhrx/src/binding/hip/api.c
```diff
@@ -3061,119 +3081,230 @@ HIPAPI hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+    if (addr >= start && addr < end) return p;
+  }
+  return NULL;
+}
+
+// Registers the pool buffer into |context|'s buffer_table (idempotent).
+// buffer_tables are per-context but the pool is shared per-device; a context
+// that cannot resolve pool pointers fails hipMemset/hipMemcpy with
+// hipErrorNotFound. Called from hipMalloc only — never on hot paths.
+static void iree_hip_pool_register_in_context_locked(
```
AWoloszyn: There are other possible returns from hrx_buffer_table_insert (memory allocation failures/etc). We should create a hrx_buffer_table_insert_if_new, and then we can propagate the error instead of ignoring it.

libhrx/src/binding/hip/api.c
```diff
@@ -3240,48 +3371,45 @@ HIPAPI hipError_t hipMalloc(void** ptr, size_t size) {
...
-        "top=%zu/%d\n",
-        g_pool_offset, aligned_size, g_pool_size,
-        100.0 * g_pool_offset / g_pool_size, g_pool_allocs_top,
-        POOL_MAX_ALLOCS);
+  // Pool created by another context on this device? Make the sub-allocation
+  // resolvable in THIS context's buffer table before handing it out.
+  iree_hip_pool_register_in_context_locked(p, context);
+
+  // Bump-allocator capacity + tracking-table capacity checks.
+  if (p->offset + aligned_size > p->size) {
```
AWoloszyn: `if (aligned_size > p->size || p->offset > p->size - aligned_size)`
Should prevent issues if we are close to SIZE_MAX (really only an issue if we end up on 32-bit archs but still)

AWoloszyn:
A couple minor comments, also I would update the commit message, this really has nothing specifically to do with amdgpu, this is at the hrx streaming layer.

libhrx/src/binding/hip/api.c
```diff
@@ -441,6 +441,10 @@ static hipError_t iree_hip_ensure_initialized(void) {
 // - Automatically calls hipInit() if needed
 // - Creates primary context for device 0 if no context exists
 // - Sets the context as current for the thread
+// Frees and resets the per-device hipMalloc pool (defined with the pool).
+static void iree_hip_free_and_reset_contiguous_pool(
+    iree_hal_streaming_context_t* context);
+
```
zjgarvey: This code block interrupted a previous comment block.

libhrx/src/binding/hip/api.c
```diff
@@ -3061,119 +3081,230 @@ HIPAPI hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+// CRITICAL (multi-GPU correctness): the pool is PER DEVICE, keyed by
+// context->device_ordinal -- NOT process-global. Buffer resolution
+// (iree_hal_streaming_memory_lookup) is per-context: it searches
+// context->buffer_table. A single global pool, created on whatever device was
+// current first (device 0) and registered only in that context, made every
+// allocation on a non-default device unresolvable -- hipMemset/hipMemcpy etc.
+// returned IREE NOT_FOUND -> hipErrorNotFound (seen via hipBLASLt on gfx942) --
+// and physically placed "device N" memory on device 0. Each device now owns a
+// pool created in, and registered in the buffer_table of, that device's
+// context.
```
zjgarvey, line 3100: Reduce dead-context comments.

zjgarvey: Better yet, file an issue, tag the issue in the PR, and put this explanation in the issue or PR description rather than a code comment. We can always look at the git history if we need to know more.

libhrx/src/binding/hip/api.c
```diff
@@ -3061,119 +3081,230 @@ HIPAPI hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
   }
 
-  // Determine pool size adaptively: 75% of total device memory, with a
-  // minimum of 256MB.  Falls back to POOL_INITIAL_SIZE if the query fails.
-  // Using 75% instead of 90% to leave headroom for the HAL driver, server
-  // overhead, and system allocations.
+  // Pool size: env override HRX_HIP_POOL_BYTES (recommended on shared
+  // machines), else 75% of total device memory (min 256MB). NOTE: 75%-of-total
+  // is very aggressive on a multi-tenant box -- one pool per device touched --
+  // so HRX_HIP_POOL_BYTES exists to cap it.
```
zjgarvey, line 3235: @AWoloszyn I know this is an intentional design choice, but it seems aggressive, and I really don't like needing to micromanage allocation sizes by an environment variable. 

Can we allocate a conservative chunk and re-allocate to double or max if we have to re-allocate an entirely new contiguous chunk (like std::vector)? Also, are there archs which have significantly less VRAM where we should change the pool minimum size?

zjgarvey (commented):
A few lingering self-review comments.

libhrx/src/binding/hip/api.c
zjgarvey: Maybe we can have the environment variable for the aggressive option rather than the other way around? We can't expect people using this software to know all the env vars when using a shared machine.

AWoloszyn: I agree, this was a matter of getting everything up and running, we should absolutely be more general. I think we can file a bug and address it as its own complete solution

AWoloszyn (approved):

zjgarvey:
I'm going to land this in-spite of the test failures on 1201 since they are orthogonal to the code changes here.

PR #54, opened 2026-06-09
hip: export 8 versioned symbols the nightly ROCm libraries import

zjgarvey:
## Summary
Nightly ROCm math/comm libraries (rccl, rocdecode/rocjpeg, hipBLASLt) import HIP symbols the HRX `libamdhip64` binding did not export, so `rocm_sdk.preload_libraries` fails and `import torch` aborts before any HRX code runs. Exports them as honest single-backend stubs returning `hipErrorNotSupported`; an unversioned export satisfies the versioned references (`foo@hip_6.5`) via glibc's unversioned-default rule.

Symbols: `hipDrvLaunchKernelEx` (@hip_6.5), `hipGetFuncBySymbol`, `hipStreamBatchMemOp`, `hipMemGetHandleForAddressRange`, `hipMemcpyBatchAsync` (rccl); `hipImportExternalMemory`, `hipExternalMemoryGetMappedBuffer`, `hipDestroyExternalMemory` (rocdecode/rocjpeg).

## Review
Agent review: clean. Completeness check confirmed these 8 close the entire runtime-HIP load-time gap (nothing imported-but-unmet remains) and the unversioned-default mechanism resolves librccl/libhipblaslt/librocdecode under RTLD_NOW. Nit fixed: the `hipDrvLaunchKernelEx` warning is now one-shot (was per-call).

## Test
Paired hip-cts: `iree-org/hip-cts` `users/zjgarvey/nightly-versioned-symbols-test` (GPU-free; asserts all 8 exported + return NotSupported; discriminating). Verified on MI300X.

GEMM-via-hipBLASLt is not yet exercised (stubs return NotSupported); implementing `hipDrvLaunchKernelEx` is follow-up. Peeled from the nightly bring-up branch.

### DIFF
--- a/libhrx/src/binding/hip/api.c
+++ b/libhrx/src/binding/hip/api.c
@@ -8285,6 +8285,116 @@ HIPAPI hipError_t hipModuleLaunchKernel(
   return result;
 }
 
+// hipDrvLaunchKernelEx — driver-style extended kernel launch. The config struct
+// carries grid/block/shared/stream plus launch attributes. The ROCm GPU math
+// libraries (hipBLASLt, hipSPARSELt, MIOpen, rocSOLVER, rocSPARSE) import this
+// as a *versioned* symbol (hipDrvLaunchKernelEx@hip_6.5). When the HRX binding
+// is the sole libamdhip64, those libraries cannot resolve their relocation and
+// `import torch` aborts during rocm_sdk's preload step. Exporting an unversioned
+// hipDrvLaunchKernelEx satisfies the versioned reference via the glibc
+// unversioned-default compatibility rule (the same way the binding's unversioned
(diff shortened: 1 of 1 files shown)


zjgarvey:
@AWoloszyn should I just actually implement these? I'm fine doing that now, but figured we could get them in as runtime failures so the nightly torch can at least not fail on import.

AWoloszyn:
> @AWoloszyn should I just actually implement these? I'm fine doing that now, but figured we could get them in as runtime failures so the nightly torch can at least not fail on import.

Let's get them in, and we can implement them as a follow-up. But please create an issue for the implementations so we don't forget.

AWoloszyn (approved):

PR #55, opened 2026-06-09
hip: resolve kpack (HIPK) out-of-band code objects

zjgarvey:
## Summary
Nightly ROCm strips code objects into sibling `.kpack` archives; `__hipRegisterFatBinary`/`hipModuleLoadData` then receive a "HIPK" wrapper whose payload is msgpack metadata, not an inline code object. Recognize the HIPK magic and resolve the real HSACO via the vendor loader `librocm_kpack`, then reuse the normal ELF/bundle path. Compiled only when rocm-kpack is available (`HRX_ENABLE_KPACK`); otherwise HIPK resolves to a propagating error. Also logs the real status on module-load failure (was an opaque `hipErrorInvalidDeviceFunction` at launch).

## Review
Agent review found two issues, both fixed: (1) [blocker] the `kpack.h` include was unconditional — now guarded by `HRX_ENABLE_KPACK` (CMake defines it when `find_package(rocm-kpack)` succeeds; `#else` returns UNAVAILABLE) so it builds without the package; (2) [major] the HIPK->bundle subpath passed an unnormalized arch that could never match — now passes the bare arch. Memory ownership reviewed leak-free.

## Test
Paired hip-cts: `iree-org/hip-cts` `users/zjgarvey/kpack-hipk-test` (a HIPK wrapper routes to the kpack resolver and fails gracefully, no crash). Verified on MI300X. Full e2e resolution needs a real `.kpack` fixture (documented).

Peeled from the nightly bring-up branch.

### DIFF
--- a/libhrx/src/binding/common/CMakeLists.txt
+++ b/libhrx/src/binding/common/CMakeLists.txt
@@ -119,11 +119,11 @@ if(HRX_ENABLE_ZSTD)
   endif()
 endif()
 
-# Out-of-band code objects (nightly ROCm "kpack"): __hipRegisterFatBinary hands
-# us a "HIPK" wrapper whose payload is msgpack metadata pointing at a sibling
-# .kpack archive rather than an inline code object. fat_binary.c resolves it with
-# the vendor loader librocm_kpack. find_package resolves via the rocm-sdk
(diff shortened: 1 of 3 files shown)


zjgarvey:
Closing so we can in-house what we need from the runtime side (we don't need the packaging side) instead of adding the dependency.

PR #56, opened 2026-06-09
amdgpu: make the AQL queue doorbell kind-aware

zjgarvey:
## Summary
Nightly ROCm on MI300X returns a USER-kind queue doorbell signal (not DOORBELL-kind), so the cached `hardware_doorbell_ptr` union slot is a signal value, not a writable MMIO address; the raw atomic store to it faulted — every H2D copy / kernel dispatch SEGV'd. Keep the raw-MMIO fast path for DOORBELL-kind; fall back to `hsa_signal_store_screlease` (system-scope release) for USER/interrupt-kind.

## Review
Agent review confirmed the HSA release-ordering correctness and that the fast path is byte-for-byte preserved, and caught a [blocker]: the `aql_ring_initialize` signature change left a second caller (`pm4_dispatch_live_test.cc`) on the old form — fixed (so the tests-enabled build compiles). Doc comment on the now-conditionally-NULL doorbell field corrected.

## Test
Paired hip-cts: `iree-org/hip-cts` `users/zjgarvey/aql-doorbell-kind-test` (32 H2D/D2H roundtrips + data integrity ring the doorbell repeatedly). Verified on MI300X (98 assertions).


runtime/src/iree/hal/drivers/amdgpu/util/aql_ring.h
```diff
@@ -70,12 +71,19 @@ typedef struct iree_hal_amdgpu_aql_ring_t {
...
+  // When non-NULL, writing a packet ID here wakes the CP to process new packets
+  // via a direct atomic store to MMIO — no libhsa indirection. Resolved at init
+  // from the doorbell signal's iree_amd_signal_t.hardware_doorbell_ptr.
   volatile int64_t* doorbell;
 
+  // Fallback doorbell signaling for non-DOORBELL-kind queue doorbell signals
+  // (nightly ROCm/MI300X returns a USER-kind doorbell). When |doorbell| above is
+  // NULL, ring through the HSA signal-store API using this handle + libhsa.
+  iree_hsa_signal_t doorbell_signal;
+  const iree_hal_amdgpu_libhsa_t* libhsa;
```
benvanik: move this up to the top of the struct on its own and comment about its lifetime (unretained, must outlive the ring) - we don't want to hide an injected dependency deep in the struct like this

runtime/src/iree/hal/drivers/amdgpu/util/aql_ring.h
```diff
@@ -70,12 +71,19 @@ typedef struct iree_hal_amdgpu_aql_ring_t {
...
+  // signal is not DOORBELL-kind (then ring via |doorbell_signal| below).
+  // When non-NULL, writing a packet ID here wakes the CP to process new packets
+  // via a direct atomic store to MMIO — no libhsa indirection. Resolved at init
+  // from the doorbell signal's iree_amd_signal_t.hardware_doorbell_ptr.
   volatile int64_t* doorbell;
 
+  // Fallback doorbell signaling for non-DOORBELL-kind queue doorbell signals
+  // (nightly ROCm/MI300X returns a USER-kind doorbell). When |doorbell| above is
+  // NULL, ring through the HSA signal-store API using this handle + libhsa.
+  iree_hsa_signal_t doorbell_signal;
```
benvanik: make a `struct { volatile int64_t* ptr; iree_hsa_signal_t signal; } doorbell;` to keep the related fields grouped together and make it clear they are representing the same thing

benvanik (changes requested):

benvanik (approved):

zjgarvey:
Something seems to be wrong with the tests I added. I'm taking a look.

zjgarvey:
Ah, stale branch and there was a new caller added for the old ring initialize func. Pushing a merge + fix soon.

PR #57, opened 2026-06-09
hip: capture GPU-reach timestamps in hipEventElapsedTime

zjgarvey:
## Summary
`hipEventElapsedTime` measured host enqueue spacing, so back-to-back records around a kernel reported ~0 ms; MIOpen's auto-tuner rejects non-positive elapsed, breaking conv tuning. Sample a monotonic timestamp inside a stream-ordered host callback (when the GPU reaches the event). Also reject a negative delta (events on independent streams) rather than returning a bogus negative value.

## Review
Agent review: the host-callback approach is coarse host-observed GPU-reach timing (documented in-code), sufficient for the auto-tuner unblock. Bounded fix applied (the negative-delta guard); device-side timestamps and the re-record error-code refinement are documented follow-ups.

## Test
Paired hip-cts: `iree-org/hip-cts` `users/zjgarvey/hip-event-elapsed-time-test` (elapsed > 0 around GPU work + disable-timing rejected). Verified on MI300X (30 assertions, 2 cases).

Peeled from the nightly bring-up branch.

### DIFF
--- a/libhrx/src/binding/common/event.c
+++ b/libhrx/src/binding/common/event.c
@@ -106,6 +106,17 @@ iree_status_t iree_hal_streaming_event_query(iree_hal_streaming_event_t* event,
   return iree_ok_status();
 }
 
+// Captures the host wall-clock time when the GPU *reaches* this point in the
+// stream (invoked as a stream-ordered host callback). Recording the time here —
+// rather than when hipEventRecord is called on the host — is what makes
+// hipEventElapsedTime reflect real GPU progress instead of host enqueue latency
+// (which is ~0 and breaks consumers like MIOpen's kernel auto-tuner, which
+// rejects non-positive elapsed times).
+static void iree_hal_streaming_event_capture_timestamp(void *user_data) {
+  iree_hal_streaming_event_t *event = (iree_hal_streaming_event_t *)user_data;
+  event->record_time_ns = iree_time_now();
+}
+
 iree_status_t iree_hal_streaming_event_record(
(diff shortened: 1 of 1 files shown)


libhrx/src/binding/common/event.c
```diff
@@ -218,8 +240,23 @@ iree_status_t iree_hal_streaming_event_elapsed_time(
...
+  // Elapsed time in ms. Both timestamps come from the same monotonic clock
+  // when the GPU reaches each event, so on a single recording stream
+  // stop >= start. A negative delta means the events were recorded on
+  // independent streams (whose host-callback timelines have no mutual order);
+  // returning a bogus negative value would reintroduce the non-positive-elapsed
+  // failure this fixes, so reject it. NOTE: this is coarse host-observed
+  // GPU-reach timing (a stream-ordered callback samples iree_time_now), not a
+  // device timestamp; cross-stream measurement and concurrent re-record are
+  // unsupported, but this is sufficient for per-launch kernel timing such as
+  // library auto-tuners.
```
zjgarvey, line 252: This is bad. Is there anyway we can get a device timestamp like native hip instead of doing a host callback?

zjgarvey (commented):
I'm going to look into the device timestamps a bit closer. The current approach is not tenable.

zjgarvey:
Closing. I have some branches up to try out wiring this up in the runtime:

https://github.com/ROCm/hrx-system/tree/users/zjgarvey/feat/hal-device-timestamp

hip-side:

https://github.com/ROCm/hrx-system/tree/users/zjgarvey/feat/hip-event-device-timestamp

PR #59, opened 2026-06-09
amdgpu: accept raw-HSACO kernels with a partial implicit kernarg suffix

zjgarvey:
## Summary
A PR#12 over-strict check rejected raw-HSACO kernels (notably MIOpen hand-written asm conv kernels) whose declared kernarg segment is smaller than the full implicit-args block the loader expects. MIOpen ignores the load failure and unloads the uninitialized handle -> SIGSEGV (conv2d). Relax the validation: the reservation is already `max(kernarg_size, explicit, implicit_offset + IMPLICIT_ARGS_SIZE)` and the dispatch path over-reserves + zero-fills, so a partial suffix is safe.

## Review
Agent review: clean-to-merge. The OOB hypothesis was adversarially disproven — the kernarg reservation is the correct MAX, computed independently of the removed check, so the full implicit-args write is always in-bounds. Rebased cleanly onto current main.

## Test
Paired hip-cts: `iree-org/hip-cts` `users/zjgarvey/miopen-asmconv-kernarg-note` — a hand-assembled gfx942 partial-kernarg kernel + load/dispatch harness (runs on MI300X), plus a note. The hand-written kernel does not yet reproduce the exact pre-fix MIOpen-unload SIGSEGV (brittle metadata); practical repro stays `conv2d_min.py`.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/executable.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/executable.c
@@ -2138,13 +2138,19 @@ iree_status_t iree_hal_amdgpu_executable_raw_hsaco_custom_kernarg_layout(
     if (!iree_host_size_checked_add(
             (iree_host_size_t)implicit_args_offset,
             (iree_host_size_t)IREE_AMDGPU_KERNEL_IMPLICIT_ARGS_SIZE,
-            &implicit_args_end) ||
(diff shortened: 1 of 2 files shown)


zjgarvey:
I don't think this is relevant anymore, so I'm closing. Will rebase and reopen if needed.

PR #63, opened 2026-06-10
[Infra] Adding clang-tidy and custom plugin

benvanik:
This adds a native clang-tidy static-analysis lane for C/C++ code and lands the first repo-owned contract check: discarded `iree_status_t` results.

The goal is to make IREE's existing C contracts visible to tools without turning static analysis into a bolt-on script that fights the build. Bazel is the source of truth for CI and normal local/agent workflows, so clang-tidy now participates in the Bazel graph as cacheable per-translation-unit actions. CMake gets a matching developer entry point for people using that lane locally, but the presubmit path does not require configuring or building the main project with CMake.

The branch is organized around four pieces of infrastructure.

First, `build_tools/bazel/compile_commands.bzl` adds repo-owned C/C++ compile command extraction. The output is useful for clang-tidy, clangd, and later static-analysis tools, so it lives under the generic Bazel tooling package rather than under clang-tidy. The implementation uses Bazel's configured C/C++ compilation contexts instead of an external helper project, which keeps it aligned with the repository's Bazel version, configuration flags, generated inputs, and future remote-execution shape.

Second, `build_tools/clang_tidy/` introduces the out-of-tree IREE clang-tidy plugin project. Bazel discovers the local LLVM/clang-tidy installation, exposes the relevant host tools and headers, builds the plugin as a host shared library, and runs plugin tests through small fixture-driven Python harnesses. The same plugin can also be built directly with CMake against the clang-tidy/LLVM installation that will load it, which keeps plugin development straightforward without putting CMake on the Bazel presubmit path.

Third, the Bazel clang-tidy layer adds one clang-tidy action per C/C++ translation unit. Each action declares the source, generated inputs, compilation flags, clang-tidy binary identity, checked-in configuration, and plugin artifact, so normal Bazel invalidation and caching apply. This is the important scaling property: when a local precommit or CI run only changes one runtime file, clang-tidy does not need to rediscover and reprocess the whole repository serially.

Fourth, `dev.py` and the presubmit provider know how to run the clang-tidy lane. Developers can run explicit Bazel or CMake clang-tidy commands, presubmit can run changed-file or full-tree scopes through the same provider registry as Semgrep, and CI has a dedicated `Presubmit / clang-tidy` workflow. Local runs skip cleanly when the optional LLVM tooling is missing, while CI can require the toolchain and fail loudly.

The first real check is `iree-status-discarded`. It diagnoses direct discards of calls returning `iree_status_t`, including expression statements and explicit `(void)` casts. A produced status must be returned, stored for later propagation, joined into an owning terminal status, or passed to an explicit consumer such as `iree_status_ignore` or `iree_status_free`. The check is deliberately narrow: it does not try to infer that an API is infallible, rewrite API contracts, or perform whole-function status lifetime analysis. Deeper status ownership checks are now a follow-up on top of this enabled rule.

The runtime, Loom, and HRX changes are the cleanup needed to make that first rule green. Most sites now either propagate the returned status or make the discard explicit at a real ownership boundary. A few APIs that were shaped like status-returning functions despite having no failure channel are tightened so callers no longer manufacture statuses just to throw them away.

The branch also includes one bit of dogfood hardening found while rebasing: Bazel's legacy Python runfiles behavior can synthesize empty `__init__.py` files. In a symlinked runfiles tree that can collide with checked-in Python package initializers and even race tests into seeing an emptied package file. The repository Bazel configuration now uses explicit Python package initializers, so tests consume the files tracked in the checkout instead of generated empty sentinels. I've "fixed" this bug 3 times today. Hopefully this one works.

Reviewer Notes

- The highest-value review surface is the Bazel action shape in `build_tools/clang_tidy/clang_tidy.bzl`, the LLVM discovery model in `build_tools/clang_tidy/llvm_repo.bzl`, and the provider dispatch in `build_tools/lefthook/presubmit.py`.
- The status check implementation intentionally lives in a status-check category rather than one file per tiny rule. The next status checks should extend that category and its fixture tests instead of creating a new plugin/test target per diagnostic.
- The status cleanup commits are meant to be behavioral no-ops except where a status-returning API was tightened because it did not actually have a failure path. Those places are worth scanning as API-shape changes rather than pure mechanical fixups.
- The dedicated clang-tidy CI lane is intentionally separate from the main presubmit workflow. That keeps LLVM toolchain provisioning and clang-tidy diagnostics isolated while we grow the check set.

Future work that'll follow this PR: local status lifetime analysis, trace-zone balance checking, retain/release/refcount ownership checks, sparse contract annotations, and the CodeQL/Infer experiments should land as separate rule-focused changes using this infrastructure.


benvanik:
this is currently conservative and running all of clang-tidy (so we ensure that we never get into a case where --all is red but all PRs breaking it are green). future improvements will drop the time and make this more targeted. it's still ~30-35s, catches real bugs, and is in the shadow of the sanitizers job so it's not the long pole. I'll watch it.

PR #68, opened 2026-06-10
Document installed test component name

AWoloszyn:


benvanik (approved):
The documented component now matches HRX_INSTALL_TESTS_COMPONENT and its HrxTestsDist default, and the explicit install prefix makes the subsequent CTest path directly runnable.

PR #69, opened 2026-06-10
Fix HRX HIP graph lifecycle query semantics

AWoloszyn:
HRX treated several HIP graph lifecycle APIs as shallow stubs: hipGraphNodeGetType always reported an empty node, hipGraphDestroyNode was a no-op, dependency insertion trusted caller-owned node handles, and hipGraphGetNodes always returned the total graph size even when the caller supplied a smaller output buffer. Those APIs expose graph-template structure directly, so the runtime needs to preserve node membership, liveness, dependency identity, and query counts consistently while the graph is being edited.

Track the owning graph on each template node while it is active, validate added dependency edges before mutating the graph, reject null/self/duplicate/cross-graph dependencies, and keep separately-added edges visible to duplicate detection. Implement node destruction by removing the node from graph/root node lists, dropping dependency references and extra edges that mention it, clearing its owning graph, and renumbering the remaining active nodes so query order stays dense.

Also map HRX internal node kinds back to HIP graph node types and make hipGraphGetNodes follow native small-buffer behavior: copy only the requested number of nodes, report the copied count, and clear the unused tail when the caller asks for more entries than exist. This makes graph lifecycle/query behavior deterministic and prevents destroyed or foreign nodes from remaining observable through later graph queries.

### DIFF
--- a/libhrx/src/binding/common/graph.c
+++ b/libhrx/src/binding/common/graph.c
@@ -46,7 +46,6 @@ iree_status_t iree_hal_streaming_graph_create(
   graph->flags = flags;
   graph->context = context;
   iree_hal_streaming_context_retain(context);
-  iree_slim_mutex_initialize(&graph->mutex);
   graph->host_allocator = host_allocator;
 
   *out_graph = graph;
@@ -64,9 +63,6 @@ static void iree_hal_streaming_graph_destroy(
 
   // Release context.
   iree_hal_streaming_context_release(graph->context);
-
-  // Deinitialize synchronization.
-  iree_slim_mutex_deinitialize(&graph->mutex);
 
   // Free graph memory itself (not allocated from arena).
   const iree_allocator_t host_allocator = graph->host_allocator;
@@ -166,8 +162,7 @@ static bool iree_hal_streaming_graph_remove_from_blocks(
 }
 
 static void iree_hal_streaming_graph_remove_dependency_refs(
-    iree_hal_streaming_graph_t* graph,
-    iree_hal_streaming_graph_node_t* node) {
+    iree_hal_streaming_graph_t* graph, iree_hal_streaming_graph_node_t* node) {
   for (iree_hal_streaming_node_block_t* block = graph->node_blocks; block;
        block = block->next) {
     for (iree_host_size_t i = 0; i < block->count; ++i) {
@@ -189,8 +184,7 @@ static void iree_hal_streaming_graph_remove_dependency_refs(
 }
 
 static void iree_hal_streaming_graph_remove_additional_edges(
-    iree_hal_streaming_graph_t* graph,
-    iree_hal_streaming_graph_node_t* node) {
+    iree_hal_streaming_graph_t* graph, iree_hal_streaming_graph_node_t* node) {
   iree_hal_streaming_graph_edge_t** next_edge = &graph->additional_edges;
   while (*next_edge) {
     iree_hal_streaming_graph_edge_t* edge = *next_edge;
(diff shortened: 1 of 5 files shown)


libhrx/src/binding/common/graph.c
```diff
@@ -115,6 +117,92 @@ void iree_hal_streaming_graph_get_nodes(
...
+    const iree_hal_streaming_graph_node_t* node) {
+  return node && node->graph == graph;
+}
+
+static bool iree_hal_streaming_graph_dependency_exists(
+    const iree_hal_streaming_graph_t* graph,
+    const iree_hal_streaming_graph_node_t* from_node,
+    const iree_hal_streaming_graph_node_t* to_node) {
+  for (uint32_t i = 0; i < to_node->dependency_count; ++i) {
+    if (to_node->dependencies[i] == from_node) return true;
```
zjgarvey, line 138: Are node deps transitive?

libhrx/src/binding/common/graph.c
```diff
@@ -115,6 +117,92 @@ void iree_hal_streaming_graph_get_nodes(
...
+        if (existing_node->dependencies[dependency_index] != node) {
+          ++dependency_index;
+          continue;
+        }
+        for (uint32_t j = dependency_index + 1;
+             j < existing_node->dependency_count; ++j) {
+          existing_node->dependencies[j - 1] = existing_node->dependencies[j];
+        }
+        --existing_node->dependency_count;
+      }
```
zjgarvey, line 181: I do wonder if having to define a "remove_item" like this just warrants using a better data structure for the dependencies in the first place?

libhrx/src/binding/common/graph.c
```diff
@@ -506,6 +629,44 @@ iree_status_t iree_hal_streaming_graph_add_dependencies(
...
+iree_status_t iree_hal_streaming_graph_destroy_node(
+    iree_hal_streaming_graph_node_t* node) {
+  if (!node || !node->graph) {
+    return iree_make_status(IREE_STATUS_INVALID_ARGUMENT,
+                            "node must belong to an active graph");
+  }
+  iree_hal_streaming_graph_t* graph = node->graph;
+  IREE_TRACE_ZONE_BEGIN(z0);
+
+  iree_slim_mutex_lock(&graph->mutex);
```
zjgarvey: Wait, why do the graphs have a lock? Is there a legitimate use-case for modifying graphs on multiple threads?

libhrx/src/binding/common/graph.c
zjgarvey, line 261: Do we not need to renumber nodes in this function because this is really appending a node to the end of the graph (rather than inserting it somewhere in the middle?

zjgarvey (commented):
Besides the usual mutex related question, I'm not sure I understand the node deps. Should we be able to remove a node from a graph if it has existing consumers?

Also, very brief doc strings would be helpful.

libhrx/src/binding/common/graph.c
AWoloszyn: No, this duplicate check is for direct edges only. Transitive ordering is handled by graph analysis, but hipGraphAddDependencies adds explicit dependency edges. Native HIP allows adding A -> C when A -> B -> C already exists, and rejects only a second direct A -> C.

libhrx/src/binding/common/graph.c
AWoloszyn: Remove should be a rare operation; we want instantiation and query/iteration to be dense. I think if we start seeing any performance questions around removal we could look deeper.

libhrx/src/binding/common/graph.c
AWoloszyn: Correct, the nodes are stored dense, and this just assigns the next dense index. We only have to renumber after deletion.

AWoloszyn:
> Besides the usual mutex related question, I'm not sure I understand the node deps. Should we be able to remove a node from a graph if it has existing consumers?
> 
> Also, very brief doc strings would be helpful.

Yes, we can remove a node that has consumers. it removes the incident edges and may leave the nodes as new roots. This is totally valid.

libhrx/src/binding/common/graph.c
AWoloszyn: I removed the mutex all-together for graph modifications. The docs state explicitly that graphs must be externally synchronized.

zjgarvey (approved):

PR #86, opened 2026-06-12
Add missing hip symbol stubs, and add version script for amdhip64.so.

AWoloszyn:
This adds stubs for all missing symbols (as expected from rocm hip tests). It also adds a python script to fetch the rocm hip-tests and check them against hrx.


libhrx/src/binding/hip/amdhip64.map
```diff
@@ -0,0 +1,632 @@
+# HIP libamdhip64 symbol version script for HRX.
+# Version nodes are generated from TheRock nightly libamdhip64.so exports.
```
zjgarvey, line 2: How is it generated? 

How is this map different than:

https://github.com/ROCm/hrx-system/blob/f2b54725aec198b88962ecd0d2877f939262556b/libhrx/src/passthrough/passthrough_full.map

zjgarvey (commented):
It would help to pull out the included "TheRock" testing script into a different PR, just because that might be a bit more contentious. E.g., does the script go in `libhrx/build_tools`? There are lot's of options and things to look at there, and it would all block the stubs getting in.

libhrx/src/binding/hip/amdhip64.map
AWoloszyn: This was generated by running the hip tests against HRX and seeing what symbols they require. 
It differs from passthrough in that it does not export the passthrough-specific symbols. 
    hip_passthrough_get_real_table;
    hip_passthrough_get_active_table;

But DOES export a few more tables (as grepped from the nightly)

AWoloszyn:
> It would help to pull out the included "TheRock" testing script into a different PR, just because that might be a bit more contentious. E.g., does the script go in `libhrx/build_tools`? There are lot's of options and things to look at there, and it would all block the stubs getting in.

Swapped this out to JUST be the symbols.

zjgarvey:
Do you know why the build is failing? Also, it's failing lint.

zjgarvey (approved):
Do you want @benvanik to check out the Bazel to cmake changes, too? I don't think he approved that part of my PR against your branch.

PR #96, opened 2026-06-15
Fix libamdhip64 link: stage version script via additional_linker_inputs

zjgarvey:
## Problem

The `Linux / AMDGPU Build + Host Tests` job on #86 fails at **Build libhrx** when linking `libamdhip64.so`:

```
ld.lld: error: cannot find version script libhrx/src/binding/hip/amdhip64.map
clang-23: error: linker command failed with exit code 1
```

(Failing run: https://github.com/ROCm/hrx-system/actions/runs/27438641025/job/81106895407)

## Fix 1 — Bazel link (`additional_linker_inputs`)

The version script was wired through `data`, which only populates **runfiles** and is not an input to the link action. Under the sandbox the `CppLink` action never sees `amdhip64.map`, so `ld.lld` can't open it (`$(location ...)` still resolves the path, which is why the flag is correct but the file is absent).

Stage it through `additional_linker_inputs` instead. The `iree_cc_binary` symbolic macro didn't expose that attribute (its link-attr group is only `linkopts` + `linkstatic`), so this adds it to the macro schema; it forwards through `**kwargs` → rules_cc `cc_binary`, which stages it into the link.

## Fix 2 — `bazel_to_cmake` so the generated CMake matches (addresses review)

`bazel_to_cmake` mirrored `linkopts` verbatim, so the generated CMake `LINKOPTS` kept the Bazel `$(location :amdhip64.map)` make-variable, which CMake can't expand. The hand-written `LINK_FLAGS` worked around it, leaving the autogen block out of sync (the `[fail] Bazel-to-CMake` presubmit).

The converter now expands `$(location ...)` in `linkopts`: a same-package source file resolves to `${CMAKE_CURRENT_SOURCE_DIR}/<name>`, which is correct regardless of the enclosing CMake project root. (The general location machinery uses `${PROJECT_SOURCE_DIR}/<repo-package>`, which is the repo root only for the top-level project — for the `libhrx` **sub-project** `PROJECT_SOURCE_DIR` is `<repo>/libhrx`, so that form would double the path.) `libhrx/src/binding/hip/CMakeLists.txt` is regenerated so the autogen `LINKOPTS` carries `${CMAKE_CURRENT_SOURCE_DIR}/amdhip64.map`, and the now-redundant hand-written `LINK_FLAGS` is dropped so the version script isn't applied twice. Added a converter unit test.

## Verification

- `bazel query` loads the target and all `//libhrx/...` cc targets (macro accepts the new attribute; no sibling breakage).
- Converter unit tests pass (incl. the new one); `ruff` clean; repo-wide `bazel_to_cmake.py --check` reports 0 files to update (blast radius limited to the hip target).
- The generated `LINKOPTS` path is byte-identical to the version-script flag that already builds today — applied via `target_link_options` instead of `LINK_FLAGS`. CI's CMake jobs validate the regenerated link end-to-end.

### DIFF
--- a/build_tools/bazel/cc.bzl
+++ b/build_tools/bazel/cc.bzl
@@ -141,6 +141,10 @@ iree_cc_binary = macro(
         cc_attrs.binary_source,
         cc_attrs.link,
         {
+            "additional_linker_inputs": attr.label_list(
+                allow_files = True,
+                doc = "Extra files made available to the link action, such as linker version scripts referenced from linkopts.",
+            ),
             "args": attr.string_list(
                 doc = "Command-line arguments used when this binary is run by Bazel.",
             ),
--- a/build_tools/bazel_to_cmake/bazel_to_cmake_converter.py
+++ b/build_tools/bazel_to_cmake/bazel_to_cmake_converter.py
@@ -382,7 +382,7 @@ class BuildFileFunctions(object):
         return deps_block, var_block
 
     def _convert_platform_select_strings(
-        self, name, block_name, values, quote=True, sort=False
+        self, name, block_name, values, quote=True, sort=False, expand_locations=False
     ):
         """Handles string lists that may contain ConditionSelect entries.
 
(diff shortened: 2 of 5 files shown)


AWoloszyn:
Looks like you will have to update the bazel to cmake python script as well. But other than that lgtm. (You will also have to rebase)

benvanik (approved):

PR #100, opened 2026-06-16
[HAL] Replace scalar device queries with cached specs

benvanik:
This branch replaces the HAL's scalar device query surface with a cached, structured description of the realized device and a separate observation path for live state. The old API was built around category/key integer lookups and ad hoc capability snapshots. That made immutable hardware and driver facts hard to cache, hard to serialize, easy to misinterpret, and awkward for consumers like HRX, remoting, and Loom JIT that need a coherent device contract instead of a loop of stringly typed probes.

The new center of gravity is `iree_hal_device_spec_t`. Every HAL device owns one immutable spec created at device startup or remote bootstrap and returns a borrowed pointer from `iree_hal_device_spec()`. The spec describes the logical device identity, participating physical devices, memory heaps and memory types, virtual-memory classes, structured external buffer and timepoint handles, queue families, dispatch limits, timing/profiling facts, executable formats, executable targets, and typed backend facets. It is retained/released like other HAL objects, can be serialized and parsed as one self-contained byte stream, and has a digest so specs can be cached, compared, replayed, and eventually embedded in reproducers.

Dynamic state moved out of the immutable device description. `iree_hal_device_observation_t` is the explicit snapshot path for facts such as currently available memory. Callers request observation flags, drivers populate the state they can sample, and helpers can fill stable totals from the cached spec. This separation is the important behavioral contract: a spec is safe to cache, send over remoting, store in a corpus, or use to impersonate a device in query-only tests; an observation is a sampled moment in time.

Drivers populate specs through `iree_hal_device_spec_builder_t` instead of hand-constructing partially-owned aggregate state. The builder owns the copying, validation, string storage, array storage, and final object creation. Driver implementation code can depend on live backend state while building the spec, but consumers see a compact borrowed object with stable pointers and no query-path allocation. That keeps fast paths such as HIP/CUDA-style `getDeviceProperties` or `memGetInfo` wrappers cheap enough for programs that ask silly questions in inner loops.

The common spec is deliberately HAL-shaped rather than backend-shaped. Dispatch dimensions, workgroup limits, subgroup width, queue counts, memory capacities, virtual-memory support, external-handle support, UUID, PCI address, NUMA attachment, timing support, executable catalogs, and topology source facts live in common structures. Driver-local facets are extension records for native facts that remain native after normalization, such as Vulkan feature/property snapshots useful to SPIR-V target construction or AMDGPU target-id details that are not already represented by executable target rows. Core HAL does not expose HIP-shaped or CUDA-shaped helpers; compatibility projection belongs to HRX and other compatibility layers.

Executable support is now represented by an explicit target catalog instead of loose architecture strings. `iree_hal_executable_target_t` records the target family, architecture, processor, feature string, artifact format, runtime ABI, loader namespace, loader target, metadata schema, kind, priority, flags, and physical-device affinity. `iree_hal_executable_target_selection_t` lets callers filter that catalog, and selection returns ordinary value-shaped results for selected, no-match, and ambiguous cases instead of forcing status-shaped probing loops. This is the bridge Loom needs for JIT target selection and the bridge HRX needs for choosing native executable compatibility without reverse-engineering driver strings.

The branch also reworks topology construction around specs. Public scalar capability queries and the old `query_capabilities` topology hook are gone. Device groups now derive a conservative row-major topology matrix from immutable specs and then allow same-backend driver refinement when the live driver can prove stronger facts. Unknown peers stay host-staged; matching names or driver families are not treated as proof of native synchronization, peer DMA, coherency, atomics, or import/export support. The topology representation keeps hot edge lookup direct while still preserving logical devices, physical placements, host NUMA nodes, external handle intersections, timepoint wait/signal capability, and backend-refined costs where available.

P0 drivers now build cached specs during device creation. Local CPU devices expose CPU identity, memory, queue, dispatch, timing, and loader-provided executable target rows. AMDGPU exposes physical identity, memory, queues, dispatch, timing, HSACO executable targets, metadata schema, external handle facts, and backend refinement for topology. Vulkan exposes its physical-device snapshot, enabled feature/extension-derived facts, memory and virtual-memory support, queue and dispatch limits, timing facts, and SPIR-V/BDA executable targets when the device was created with buffer device address support. HIP exposes native runtime identity and the memory, queue, dispatch, timing, observation, and executable facts needed by HRX as a debug oracle. Null/mock/minimal and best-effort drivers create honest minimal specs instead of leaving the vtable hole empty.

The old scalar HAL device query API is deleted rather than wrapped. Callers no longer ask for `hal.device` integer keys or fetch topology through a capability callback. They ask the device for its spec, use typed accessors and selectors, and sample observations when they need live state. This is more churn than a compatibility layer, but it removes the ambiguity that made the old surface hard to evolve: immutable device facts, dynamic state, executable compatibility, topology, and backend-native extension data now have separate contracts.

HRX was updated to consume the new HAL model directly. Immutable HIP/CUDA-like property projection now reads the common device spec instead of calling scalar HAL queries or deriving architecture from `hal.device/gfxip` strings. Generic HRX memory-info APIs sample `iree_hal_device_observation_t`, while the HIP/CUDA streaming layer preserves its local allocation ledger for `memGetInfo` compatibility and takes total memory from the stable spec. Native executable selection uses the HAL executable target catalog, including loader target and metadata schema, which keeps HRX policy in HRX without forcing HIP or CUDA compatibility structs into the HAL. HIP fat-binary loading now builds a ranked exact-then-generic target list from that catalog: specialized HSACO entries such as `gfx1100` win when present, while compatible generic code-object entries such as `gfx11-generic` are used as a fallback when an exact image is absent.

For HRX users this should show up as richer and more consistent device reporting. HIP and CUDA compatibility queries now draw from one startup-time device snapshot instead of several independent fallbacks, so properties such as architecture, total memory, launch limits, warp/subgroup size, multiprocessor count, shared-memory limits, register limits, and native executable target identity agree with the HAL device HRX is actually running on. `hipGetDeviceProperties`, `hipDeviceGetAttribute`, CUDA driver attributes, `hipDeviceTotalMem`, `cuDeviceTotalMem`, and `memGetInfo`-style APIs all move closer to native-runtime behavior without making the HAL expose HIP- or CUDA-shaped structs. Libraries that use these properties for kernel selection, occupancy estimates, fat-binary dispatch, or memory-pool sizing should see fewer conservative guesses and fewer contradictory answers between APIs. Libraries that ship fat HIP modules with both exact and generic AMDGPU images should also get native-style target selection: exact device images are preferred, generic family images remain loadable, and canonical AMDGPU target IDs with feature suffixes are accepted instead of being forced into identifier-shaped strings.

This design also gives remoting and Loom the shape they asked for, even though their deeper migrations are separate branches. A remote HAL server can capture and serialize the immutable spec once, ship it to clients, and reserve live RPCs for observations or actual work. Loom can build target profiles from executable target rows and backend facets without walking scalar query loops or guessing which metadata schema a loader expects. Tooling can later dump specs as structured human- and agent-readable data, round-trip JSON into query-only mock devices, and maintain a corpus of real machine specs for diffing driver and hardware behavior.

The remaining follow-on boundary is executable packaging. This branch adds structured executable target rows and selection semantics, but the existing executable cache/load APIs still carry some historical naming and packaging pressure around format, loader target, metadata, and artifact compatibility. That cleanup is intentionally kept as a follow-up so this PR can land the device-information model first and let Loom, remoting, HRX, and executable loading converge on the same spec-backed vocabulary.


benvanik:
For HRX folks:

> HRX was updated to consume the new HAL model directly. Immutable HIP/CUDA-like property projection now reads the common device spec instead of calling scalar HAL queries or deriving architecture from hal.device/gfxip strings. Generic HRX memory-info APIs sample iree_hal_device_observation_t, while the HIP/CUDA streaming layer preserves its local allocation ledger for memGetInfo compatibility and takes total memory from the stable spec. Native executable selection uses the HAL executable target catalog, including loader target and metadata schema, which keeps HRX policy in HRX without forcing HIP or CUDA compatibility structs into the HAL. HIP fat-binary loading now builds a ranked exact-then-generic target list from that catalog: specialized HSACO entries such as gfx1100 win when present, while compatible generic code-object entries such as gfx11-generic are used as a fallback when an exact image is absent.

> For HRX users this should show up as richer and more consistent device reporting. HIP and CUDA compatibility queries now draw from one startup-time device snapshot instead of several independent fallbacks, so properties such as architecture, total memory, launch limits, warp/subgroup size, multiprocessor count, shared-memory limits, register limits, and native executable target identity agree with the HAL device HRX is actually running on. hipGetDeviceProperties, hipDeviceGetAttribute, CUDA driver attributes, hipDeviceTotalMem, cuDeviceTotalMem, and memGetInfo-style APIs all move closer to native-runtime behavior without making the HAL expose HIP- or CUDA-shaped structs. Libraries that use these properties for kernel selection, occupancy estimates, fat-binary dispatch, or memory-pool sizing should see fewer conservative guesses and fewer contradictory answers between APIs. Libraries that ship fat HIP modules with both exact and generic AMDGPU images should also get native-style target selection: exact device images are preferred, generic family images remain loadable, and canonical AMDGPU target IDs with feature suffixes are accepted instead of being forced into identifier-shaped strings.

---

The values weren't quite right before (some hardcoding, partial info) and they should be more right now, but it does perturb results that may have been skating by before by chance. My hope performance-wise is that the new executable target selection will help as it used to sometimes pick generic over specific in fatbins - now it always prefers more specific. The rules may still need adjusting, but now there's a bunch of info to base the heuristics on vs. having to hope the string slicing works.

AWoloszyn (approved):

AWoloszyn (approved):

PR #120, opened 2026-06-18
Fix HIP memory, graph, and stream handling.

AWoloszyn:
- Adds HRX-backed HIP pooled allocation support.
  - Fixes graph capture/launch semantics and graph dependency validation.
  - Implements stable HIP stream IDs and serialized stream recording.
  - Tightens stream/context teardown so pending work is synchronized before handles are detached.
  - Removes stale HIP-local memory-pool helper code after rebasing onto the HRX pool implementation.
  - Preserves HRX pool buffer ownership for pooled allocations.
  - Adds overflow/range checks for stream lists, graph memcpy/memset ranges, graph-exec memset sizes, synthetic pointer alignment, and pitched allocations.
  - Cleans up context failure paths so partially-created contexts and default memory pools are released consistently.


libhrx/src/binding/common/graph.c
```diff
@@ -736,13 +1131,326 @@ iree_status_t iree_hal_streaming_begin_capture(
...
+        graph, host_allocator, &additional_edge_index);
+  }
+  if (iree_status_is_ok(status)) {
+    status = iree_hal_streaming_capture_mark_frontier_reachable(
+        graph, origin_stream->capture_dependencies,
+        origin_stream->capture_dependency_count, &additional_edge_index,
+        reachable_nodes, stack);
+  }
+  if (!iree_status_is_ok(status)) {
+    iree_hal_streaming_graph_deinitialize_additional_edge_index(
```
zjgarvey, line 1649: I think there is a code path where `initialize_additional_edge_index` calls the `deinitialize` in an error path and returns an error status, so this additional `deinitialize` would trigger a double free. Maybe remove the deinit that occurs inside the init func error path since it's handled here.

libhrx/src/binding/common/stream.c
```diff
@@ -534,13 +542,37 @@ iree_status_t iree_hal_streaming_stream_wait_event(
...
+        stream->capture_mode = event->recording_stream->capture_mode;
+        stream->capture_id = event->recording_stream->capture_id;
+        stream->capture_owner_thread_id =
+            event->recording_stream->capture_owner_thread_id;
+      } else {
+        stream->capture_mode = IREE_HAL_STREAMING_CAPTURE_MODE_GLOBAL;
+        stream->capture_id = stream->capture_id + 1;
+        stream->capture_owner_thread_id = 0;
+      }
+      iree_hal_streaming_graph_retain(stream->capture_graph);
```
zjgarvey: These stream writes be racy against `graph.c` functions which look at capture participants (e.g., clear, and has_unjoined) and `api.c` invalidation helpers. Maybe need a mutex here.

libhrx/src/binding/hip/api.c
```diff
@@ -10628,12 +11002,28 @@ HIPAPI hipError_t
...
+  if (!pGraphNode || !graph || !event) {
+    IREE_TRACE_ZONE_END(z0);
+    HIP_RETURN_ERROR(hipErrorInvalidValue);
+  }
+  if (numDependencies > 0 && !pDependencies) {
+    IREE_TRACE_ZONE_END(z0);
+    HIP_RETURN_ERROR(hipErrorInvalidValue);
+  }
+
+  iree_hal_streaming_graph_node_t* node = NULL;
```
zjgarvey, line 11291: Why is this a NULL? Is the type enough or something?

zjgarvey: Yeah, I guess I just don't understand where the event is going here.

libhrx/src/binding/hip/api.c
```diff
@@ -522,6 +734,65 @@ static hipError_t iree_hip_ensure_context(
...
+      context, context->default_stream->flags, context->default_stream->priority,
+      context->host_allocator, &stream);
+  if (!iree_status_is_ok(status)) {
+    iree_status_free(status);
+    *out_stream = NULL;
+    return hipErrorOutOfMemory;
+  }
+
+  iree_hip_per_thread_stream_context = context;
+  iree_hip_per_thread_stream = stream;
```
zjgarvey: I'm a bit confused on how these get cleaned up when the thread exits. I only see the clear function called earlier in this function body, and again during device reset.

libhrx/src/binding/hip/api.c
```diff
@@ -459,23 +477,217 @@ static hipError_t hrx_status_to_hip_result(hrx_status_t status) {
...
+
+static bool iree_hip_status_to_capture_invalidation_failure(
+    iree_status_t status) {
+  if (iree_status_is_ok(status)) return false;
+  iree_status_fprint(stderr, status);
+  iree_status_free(status);
+  return true;
+}
+
+static bool iree_hip_context_invalidate_visible_captures(
```
zjgarvey, line 614: This is used in a lot of places, and seems expensive if there isn't any capturing going on. E.g., every hipMalloc needs to snapshot all streams just to see if any of them are capturing? 

Is there like a bool we can early return on here if no captures are going on?

libhrx/src/binding/common/memory.c
```diff
@@ -929,113 +1021,73 @@ iree_status_t iree_hal_streaming_memcpy_host_to_device(
...
-      if (!iree_status_is_ok(flush_status)) {
-        IREE_TRACE_ZONE_END(z0);
-        return flush_status;
-      }
+  const iree_device_size_t staging_threshold = 256 * 1024;
+  if (size >= staging_threshold) {
+    iree_hal_streaming_stream_t* copy_stream =
+        stream ? stream : context->default_stream;
+    iree_hal_streaming_buffer_t* staging = NULL;
+    iree_slim_mutex_lock(&context->mutex);
```
zjgarvey, line 1038: Why the context mutex and not the `copy_stream->mutex`?

libhrx/src/binding/common/memory.c
```diff
@@ -1061,72 +1150,70 @@ iree_status_t iree_hal_streaming_memcpy_device_to_host(
...
+  }
+  if (iree_status_is_ok(direct_status)) {
+    IREE_TRACE_ZONE_END(z0);
+    return iree_ok_status();
+  }
+  iree_status_ignore(direct_status);
 
-    // Record copy command.
+  iree_hal_streaming_stream_t* copy_stream =
+      stream ? stream : context->default_stream;
```
zjgarvey, line 1183: Does D2H not require any mutexes?

libhrx/src/binding/hip/api.c
```diff
@@ -4192,74 +4539,87 @@ HIPAPI hipError_t hipMemcpyAsync(void* dst, const void* src, size_t sizeBytes,
...
+  hipError_t range_result =
+      iree_hip_validate_memcpy_ranges(context, dst, src, sizeBytes, kind,
+                                      false);
+  if (range_result != hipSuccess) {
+    IREE_TRACE_ZONE_END(z0);
+    HIP_RETURN_ERROR(range_result);
   }
 
   iree_status_t status = iree_ok_status();
+  if (!stream || stream == hipStreamLegacy || stream == hipStreamPerThread) {
```
zjgarvey: Still trying to understand some of the per-thread stream stuff. Why do we need to synchronize contexts in the per-thread case here?

libhrx/src/binding/hip/api.c
```diff
@@ -10871,45 +11707,84 @@ HIPAPI hipError_t hipGraphAddChildGraphNode(hipGraphNode_t* pGraphNode,
...
                                             size_t numDependencies,
                                             hipGraph_t childGraph) {
-  (void)graph;
-  (void)pDependencies;
-  (void)numDependencies;
-  (void)childGraph;
-  if (pGraphNode) *pGraphNode = NULL;
-  HIP_RETURN_ERROR(hipErrorNotSupported);
+  IREE_TRACE_ZONE_BEGIN(z0);
+  if (!pGraphNode || !graph || !childGraph) {
```
zjgarvey, line 11795: Are there really no other checks on the graphs needed here? E.g., can `childGraph = graph`? Or `graph` is transitively a child graph of `childGraph`? I'm not sure what is allowed, or if that's all just up to callers to make sure they don't do some kind of unbounded nesting.

libhrx/src/binding/common/graph.c
AWoloszyn: Good catch.

libhrx/src/binding/hip/api.c
AWoloszyn: Good catch. Type alone was not enough. Event graph nodes now store and retain the event in the node attrs, and graph exec uses that retained event to
  signal/wait. The NULL is now only “no extra trailing node data.”

libhrx/src/binding/hip/api.c
AWoloszyn: Addressed with binding-local TLS. The per-thread stream state has a destructor, so thread exit releases the cached stream and frees the small per-thread
  state object. Explicit clears still handle device reset/context replacement.

libhrx/src/binding/common/memory.c
AWoloszyn: The context mutex is for the shared pageable H2D staging buffer, not for stream submission. Stream command-buffer mutation is separately protected by
  copy_stream->mutex.

libhrx/src/binding/common/memory.c
AWoloszyn: The direct D2H path is a blocking device transfer and does not mutate stream command-buffer state. The fallback command-buffer path does take
  copy_stream->mutex while recording.

libhrx/src/binding/hip/api.c
AWoloszyn: Agreed, the previous context-wide sync was too broad. hipStreamPerThread should enqueue like an ordinary stream and not synchronize with unrelated streams. I narrowed the NULL/hipStreamLegacy path so it only synchronizes streams that participate in legacy default-stream ordering, excluding hipStreamNonBlocking; per-thread default streams remain blocking streams and only participate in the legacy interop case.

libhrx/src/binding/hip/api.c
AWoloszyn: Added validation in the common graph helper. It rejects childGraph == graph, rejects graphs from a different context, and walks nested child graph nodes to reject recursive containment

zjgarvey (approved):
This looks good to me.

PR #142, opened 2026-06-24
Improve HIP graph, memory, and stream runtime coverage

AWoloszyn:
Add graph execution, pooled memory, stream timeline, and HIP API coverage work.

### DIFF
--- a/libhrx/src/binding/common/device.c
+++ b/libhrx/src/binding/common/device.c
@@ -262,7 +262,7 @@ iree_status_t iree_hal_streaming_device_primary_context_state(
   return iree_ok_status();
 }
 
-iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
+static iree_status_t iree_hal_streaming_device_ensure_default_mem_pool_locked(
     iree_hal_streaming_device_t* device) {
   IREE_ASSERT_ARGUMENT(device);
   if (device->default_mem_pool && device->current_mem_pool) {
@@ -291,6 +291,16 @@ iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
     hrx_mem_pool_retain(device->current_mem_pool);
   }
   return iree_ok_status();
+}
+
+iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
+    iree_hal_streaming_device_t* device) {
+  IREE_ASSERT_ARGUMENT(device);
+  iree_slim_mutex_lock(&device->primary_context_mutex);
+  iree_status_t status =
+      iree_hal_streaming_device_ensure_default_mem_pool_locked(device);
+  iree_slim_mutex_unlock(&device->primary_context_mutex);
+  return status;
 }
 
 iree_status_t iree_hal_streaming_device_get_or_create_primary_context(
@@ -334,7 +344,7 @@ iree_status_t iree_hal_streaming_device_get_or_create_primary_context(
 
   // Ensure runtime allocations have a backing pool for this device.
   if (iree_status_is_ok(status)) {
-    status = iree_hal_streaming_device_ensure_default_mem_pool(device);
+    status = iree_hal_streaming_device_ensure_default_mem_pool_locked(device);
   }
 
   if (iree_status_is_ok(status)) {
@@ -383,7 +393,7 @@ iree_status_t iree_hal_streaming_device_retain_primary_context(
 
     // Ensure runtime allocations have a backing pool for this device.
     if (iree_status_is_ok(status)) {
-      status = iree_hal_streaming_device_ensure_default_mem_pool(device);
+      status = iree_hal_streaming_device_ensure_default_mem_pool_locked(device);
     }
 
     if (!iree_status_is_ok(status)) {
--- a/libhrx/src/binding/common/device.c
+++ b/libhrx/src/binding/common/device.c
@@ -262,6 +262,37 @@ iree_status_t iree_hal_streaming_device_primary_context_state(
   return iree_ok_status();
 }
 
+iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
+    iree_hal_streaming_device_t* device) {
+  IREE_ASSERT_ARGUMENT(device);
+  if (device->default_mem_pool && device->current_mem_pool) {
+    return iree_ok_status();
+  }
+
+  if (!iree_hal_streaming_device_registry()) {
+    return iree_make_status(IREE_STATUS_FAILED_PRECONDITION,
+                            "device registry not initialized");
+  }
+
+  hrx_mem_pool_t default_mem_pool = NULL;
+  if (!device->default_mem_pool) {
+    hrx_mem_pool_props_t props = {
+        .alloc_handle_type = 0,
+        .location_type = 1,  // device
+        .location_id = (int)device->ordinal,
+    };
+    IREE_RETURN_IF_ERROR(HRX_CALL(hrx_mem_pool_create(
+        device->hrx_device, &props, &default_mem_pool)));
+    device->default_mem_pool = default_mem_pool;
+  }
+
+  if (!device->current_mem_pool) {
+    device->current_mem_pool = device->default_mem_pool;
+    hrx_mem_pool_retain(device->current_mem_pool);
+  }
+  return iree_ok_status();
+}
+
 iree_status_t iree_hal_streaming_device_get_or_create_primary_context(
     iree_hal_streaming_device_t* device,
     iree_hal_streaming_context_t** out_context) {
@@ -301,20 +332,9 @@ iree_status_t iree_hal_streaming_device_get_or_create_primary_context(
       device, device->primary_context_flags, device_registry->host_allocator,
       &device->primary_context);
 
-  // Create default memory pool via pyre.
+  // Ensure runtime allocations have a backing pool for this device.
   if (iree_status_is_ok(status)) {
-    iree_host_size_t device_ordinal = device - device_registry->devices;
-    hrx_mem_pool_props_t props = {
-        .alloc_handle_type = 0,
-        .location_type = 1,  // device
-        .location_id = (int)device_ordinal,
-    };
-    status = HRX_CALL(hrx_mem_pool_create(device->hrx_device, &props,
-                                          &device->default_mem_pool));
-    if (iree_status_is_ok(status)) {
-      device->current_mem_pool = device->default_mem_pool;
-      hrx_mem_pool_retain(device->current_mem_pool);
-    }
+    status = iree_hal_streaming_device_ensure_default_mem_pool(device);
   }
 
   if (iree_status_is_ok(status)) {
@@ -361,20 +381,9 @@ iree_status_t iree_hal_streaming_device_retain_primary_context(
         device, device->primary_context_flags, device_registry->host_allocator,
         &device->primary_context);
 
-    // Create default memory pool via pyre if context was created successfully.
-    if (iree_status_is_ok(status) && !device->default_mem_pool) {
-      iree_host_size_t device_ordinal = device - device_registry->devices;
-      hrx_mem_pool_props_t props = {
-          .alloc_handle_type = 0,
-          .location_type = 1,  // device
-          .location_id = (int)device_ordinal,
-      };
-      status = HRX_CALL(hrx_mem_pool_create(device->hrx_device, &props,
-                                            &device->default_mem_pool));
-      if (iree_status_is_ok(status)) {
-        device->current_mem_pool = device->default_mem_pool;
-        hrx_mem_pool_retain(device->current_mem_pool);
-      }
+    // Ensure runtime allocations have a backing pool for this device.
+    if (iree_status_is_ok(status)) {
+      status = iree_hal_streaming_device_ensure_default_mem_pool(device);
     }
 
     if (!iree_status_is_ok(status)) {
--- a/libhrx/src/binding/common/device.c
+++ b/libhrx/src/binding/common/device.c
@@ -281,8 +281,8 @@ iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
         .location_type = 1,  // device
         .location_id = (int)device->ordinal,
     };
-    IREE_RETURN_IF_ERROR(HRX_CALL(hrx_mem_pool_create(
-        device->hrx_device, &props, &default_mem_pool)));
+    IREE_RETURN_IF_ERROR(HRX_CALL(
+        hrx_mem_pool_create(device->hrx_device, &props, &default_mem_pool)));
     device->default_mem_pool = default_mem_pool;
   }
 
--- a/libhrx/src/binding/common/graph.c
+++ b/libhrx/src/binding/common/graph.c
@@ -44,6 +44,21 @@ static void iree_hal_streaming_graph_node_deinitialize_attrs(
   }
 }
 
+static bool iree_hal_streaming_graph_contains_graph_memory_nodes(
+    const iree_hal_streaming_graph_t* graph) {
+  for (iree_hal_streaming_node_block_t* block = graph->node_blocks; block;
+       block = block->next) {
+    for (iree_host_size_t i = 0; i < block->count; ++i) {
+      iree_hal_streaming_graph_node_t* node = block->nodes[i];
+      if (node->type == IREE_HAL_STREAMING_GRAPH_NODE_TYPE_MEM_ALLOC ||
+          node->type == IREE_HAL_STREAMING_GRAPH_NODE_TYPE_MEM_FREE) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
 static iree_atomic_uint64_t iree_hal_streaming_next_graph_debug_id =
     IREE_ATOMIC_VAR_INIT(1);
 static iree_atomic_uint64_t iree_hal_streaming_next_node_debug_id =
@@ -1922,6 +1937,9 @@ iree_status_t iree_hal_streaming_graph_destroy_node(
                             "node must belong to an active graph");
   }
   iree_hal_streaming_graph_t* graph = node->graph;
+  const bool removed_graph_memory_node =
+      node->type == IREE_HAL_STREAMING_GRAPH_NODE_TYPE_MEM_ALLOC ||
+      node->type == IREE_HAL_STREAMING_GRAPH_NODE_TYPE_MEM_FREE;
   IREE_TRACE_ZONE_BEGIN(z0);
 
   iree_hal_streaming_graph_remove_dependency_refs(graph, node);
@@ -1940,6 +1958,10 @@ iree_status_t iree_hal_streaming_graph_destroy_node(
   }
   if (removed_from_nodes) {
     iree_hal_streaming_graph_node_deinitialize_attrs(node);
+    if (removed_graph_memory_node) {
+      graph->has_graph_memory_nodes =
+          iree_hal_streaming_graph_contains_graph_memory_nodes(graph);
+    }
     iree_hal_streaming_graph_renumber_nodes(graph);
     node->graph = NULL;
     node->dependency_count = 0;
--- a/libhrx/src/binding/common/graph.c
+++ b/libhrx/src/binding/common/graph.c
@@ -2458,7 +2458,8 @@ iree_status_t iree_hal_streaming_end_capture(
   }
   if (has_unjoined_participant) {
     const bool capture_graph_owned = stream->capture_graph_owned;
-    stream->capture_status = IREE_HAL_STREAMING_CAPTURE_STATUS_NONE;
+    iree_hal_streaming_stream_set_capture_status(
+        stream, IREE_HAL_STREAMING_CAPTURE_STATUS_NONE);
     stream->capture_graph = NULL;
     stream->capture_graph_owned = false;
     stream->capture_origin = false;
--- a/libhrx/src/binding/common/graph.c
+++ b/libhrx/src/binding/common/graph.c
@@ -31,6 +31,11 @@ static void iree_hal_streaming_graph_node_deinitialize_attrs(
   }
 }
 
+static iree_atomic_uint64_t iree_hal_streaming_next_graph_debug_id =
+    IREE_ATOMIC_VAR_INIT(1);
+static iree_atomic_uint64_t iree_hal_streaming_next_node_debug_id =
+    IREE_ATOMIC_VAR_INIT(1);
+
 iree_status_t iree_hal_streaming_graph_create(
     iree_hal_streaming_context_t* context,
     iree_hal_streaming_graph_flags_t flags, iree_allocator_t host_allocator,
@@ -56,6 +61,10 @@ iree_status_t iree_hal_streaming_graph_create(
   graph->current_node_block = NULL;
   graph->node_count = 0;
   graph->child_graph_node_count = 0;
+  graph->debug_id =
+      iree_atomic_fetch_add(&iree_hal_streaming_next_graph_debug_id, 1,
+                            iree_memory_order_relaxed);
+  graph->clone_source_graph_debug_id = 0;
   graph->next_clone_source_node_index = 0;
   graph->root_blocks = NULL;
   graph->current_root_block = NULL;
@@ -63,6 +72,9 @@ iree_status_t iree_hal_streaming_graph_create(
   graph->additional_edges = NULL;
   graph->additional_edge_count = 0;
   graph->owned_host_allocations = NULL;
+  graph->user_object_refs = NULL;
+  graph->has_graph_memory_nodes = false;
+  graph->active_graph_memory_exec_count = 0;
   graph->flags = flags;
   graph->context = context;
   iree_hal_streaming_context_retain(context);
@@ -92,6 +104,14 @@ static void iree_hal_streaming_graph_destroy(
           graph->context, owned_allocation->host_ptr));
     }
     owned_allocation = owned_allocation->next;
+  }
+
+  for (iree_hal_streaming_graph_user_object_ref_t* user_ref =
+           graph->user_object_refs;
+       user_ref; user_ref = user_ref->next) {
+    if (user_ref->count > 0) {
+      user_ref->release(user_ref->object, user_ref->count);
+    }
   }
 
   // Reset the arena - this frees all nodes and arrays at once.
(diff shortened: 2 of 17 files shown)


libhrx/src/binding/common/graph.c
```diff
@@ -1726,7 +2435,20 @@ iree_status_t iree_hal_streaming_end_capture(
     return joined_status;
   }
   if (has_unjoined_participant) {
+    const bool capture_graph_owned = stream->capture_graph_owned;
+    stream->capture_status = IREE_HAL_STREAMING_CAPTURE_STATUS_NONE;
```
zjgarvey: **M1 — MEDIUM · confidence high**

This abort branch resets `stream->capture_status` to `NONE` with a raw assignment instead of `iree_hal_streaming_stream_set_capture_status()`. The helper is the only path that calls `iree_hal_streaming_context_leave_capture()` on the non-NONE->NONE edge; the sibling branches use it (the INVALIDATED branch above and the success branch below). The stream is necessarily `ACTIVE` here, so the matching `enter_capture` is never balanced and `context->capture_stream_count` stays permanently elevated by 1 each time this path runs. `iree_hal_streaming_context_has_capture_streams()` then returns true for the life of the context, forcing every op onto the capture-detection slow path. Triggerable when a participant stream is not joined back before end.

Fix: `iree_hal_streaming_stream_set_capture_status(stream, IREE_HAL_STREAMING_CAPTURE_STATUS_NONE)` here.

libhrx/src/binding/common/memory.c
```diff
@@ -1018,16 +1592,36 @@ iree_status_t iree_hal_streaming_memcpy_host_to_device(
...
-  // Host allocations registered through hipHostMalloc are not guaranteed to be
-  // backed by AMDGPU command-buffer-compatible allocations. Use the blocking
-  // transfer path for host sources until registered host buffers carry the
-  // required lifetime and backing-storage contract.
+  if (dst_ref.buffer->host_ptr &&
+      iree_any_bit_set((iree_hal_memory_type_t)dst_ref.buffer->memory_type,
+                       IREE_HAL_MEMORY_TYPE_HOST_LOCAL)) {
+    iree_status_t status = iree_hal_streaming_context_synchronize_all();
+    if (iree_status_is_ok(status)) {
+      memcpy((uint8_t*)dst_ref.buffer->host_ptr + dst_ref.offset, src, size);
```
zjgarvey, line 1627: **M3 — MEDIUM · confidence high**

This host-local fast path resolves the buffer with `iree_hal_streaming_memory_lookup` (which returns only the offset and does not validate that `offset + size` fits the allocation), then `memcpy`s `size` bytes into `host_ptr + offset`. The path this replaced fell through to `iree_hal_device_transfer_h2d`, where the HAL bounds-checks the range. With no check here, an out-of-range `size` is a host heap overflow instead of a returned error. The device-to-host fast path at line 1752 has the symmetric over-read. (The memset path is unaffected — it uses `lookup_range`.)

Fix: before the memcpy, reject when `dst_ref.offset > buffer->size || size > buffer->size - dst_ref.offset` (and symmetrically for the source at 1752), returning `IREE_STATUS_OUT_OF_RANGE`.

libhrx/src/binding/common/memory.c
```diff
@@ -797,27 +1324,44 @@ iree_status_t iree_hal_streaming_memory_memset(
...
+    iree_status_ignore(lookup_status);
+    iree_hal_streaming_context_t* owner_context = NULL;
+    lookup_status = iree_hal_streaming_memory_lookup_range_across_contexts(
+        dst, length, &owner_context, &dst_ref);
+    if (iree_status_is_ok(lookup_status)) {
+      if (!dst_ref.buffer->is_managed) {
+        iree_hal_streaming_context_release(owner_context);
+        lookup_status = iree_status_from_code(IREE_STATUS_NOT_FOUND);
+      } else {
+        iree_hal_streaming_context_release(owner_context);
```
zjgarvey: **M4 — MEDIUM · confidence medium**

`memory_lookup_range_across_contexts` returns a *retained* owner context plus a `buffer_ref` into that context's registry — the retained reference is what keeps the buffer alive while used. Here the owner context is released (line 1343) and then the borrowed buffer is dereferenced (`dst_ref.buffer->is_managed` / `->host_ptr`, lines 1350-1354). The global registry holds its own reference to each registered context, so a single-threaded caller is safe; if another thread destroys the owning context in this window it's a use-after-free. The same release-then-use shape appears at stream.c:786 (deref 791-798) and api.c:3953 (callers deref at 3968/3977/5063).

Fix: hold `owner_context` until after the last dereference of the borrowed buffer on every path. (Same lifetime theme as H2.)

libhrx/src/binding/common/device.c
```diff
@@ -262,6 +262,37 @@ iree_status_t iree_hal_streaming_device_primary_context_state(
   return iree_ok_status();
 }
 
+iree_status_t iree_hal_streaming_device_ensure_default_mem_pool(
+    iree_hal_streaming_device_t* device) {
+  IREE_ASSERT_ARGUMENT(device);
+  if (device->default_mem_pool && device->current_mem_pool) {
```
zjgarvey, line 268: **M7 — MEDIUM · confidence medium**

`iree_hal_streaming_device_ensure_default_mem_pool` does an unguarded check-create-assign on `device->default_mem_pool`/`current_mem_pool`. The callers in this file hold `device->primary_context_mutex`, but `hipCtxCreate` calls it (api.c:3225) with no lock held. Two threads creating contexts on the same device can both observe `default_mem_pool == NULL`, both `hrx_mem_pool_create`, and both store — leaking one pool and double-retaining `current_mem_pool` — plus the torn read/write the mutex is meant to serialize.

Fix: take `device->primary_context_mutex` inside this function (or hold it across the `hipCtxCreate` call).

libhrx/src/binding/common/graph_exec.c
```diff
@@ -166,17 +184,403 @@ typedef struct iree_hal_streaming_graph_exec_t {
...
+        status = iree_allocator_malloc(context->host_allocator, sizeof(*entry),
+                                       (void**)&entry);
+        if (iree_status_is_ok(status)) {
+          entry->next = context->graph_memory_reusable_size_entries;
+          entry->size = contribution->size;
+          entry->reference_count = 1;
+          context->graph_memory_reusable_size_entries = entry;
+          context->graph_memory_used_current += bytes;
+          context->graph_memory_reserved_current += bytes;
+          iree_hal_streaming_graph_memory_add_high_water(context, bytes);
```
zjgarvey: **M8 — MEDIUM · confidence high**

When a new *reusable* size class is created, the high-water mark is updated with this class's `bytes` rather than the running total `context->graph_memory_used_current` (used correctly for the orphan path a few lines below). When a reusable-only graph is instantiated while another size class is already live, `hipGraphMemAttrUsedMemHigh`/`ReservedMemHigh` report `max(high, S2)` instead of `max(high, S1+S2)`, under-reporting the true peak.

Fix: pass `context->graph_memory_used_current` (the total) after updating the counters, or update high-water from the total once at function end.

libhrx/src/binding/common/graph_exec.c
```diff
@@ -166,17 +184,403 @@ typedef struct iree_hal_streaming_graph_exec_t {
...
+  iree_slim_mutex_lock(&context->mutex);
+  context->graph_memory_used_high = 0;
+  context->graph_memory_reserved_high = 0;
+  iree_slim_mutex_unlock(&context->mutex);
+}
+
+void iree_hal_streaming_graph_memory_trim(
+    iree_hal_streaming_context_t* context) {
+  iree_slim_mutex_lock(&context->mutex);
+  context->graph_memory_used_current = 0;
```
zjgarvey: **M9 — MEDIUM · confidence medium**

`graph_memory_trim` zeroes `used/reserved_current` and frees the entire reusable size-entry list with no check of `reference_count`, even for classes still referenced by live execs. Because execs share classes by size lookup (not stored pointer), a later same-size instantiation recreates an entry, and a pre-trim exec's release then finds that new entry, decrements/frees it and subtracts its bytes — corrupting accounting and under-counting `used_current` while execs are live. No memory-unsafety (entries are looked up by size each call; subtraction is `iree_min`-clamped), but `hipGraphMemAttrUsedMemCurrent` becomes wrong and the 'freed when reference_count hits 0' invariant is violated.

Fix: free only entries with `reference_count == 0` and subtract only their bytes, leaving in-use classes and counters intact.

zjgarvey (commented):
> _Provenance: this is an initial pre-human review pass generated by Claude Code (Anthropic), which I have reviewed before posting; I'm also doing my own independent pass over the code and may amend or add comments. Each comment carries my read of severity and confidence — input to your judgment, not automated gates._

The items below are either in `hip/api.c` (its diff is too large for GitHub to anchor inline comments), in code this PR doesn't modify, or broader than a single line. Inline comments cover the rest of the changed files.

## Findings in `hip/api.c` (inline anchoring unavailable — its diff is too large for GitHub to resolve line comments)

- **H2 — HIGH · confidence high — graph memcpy-node update stores remote contexts without retaining (api.c:19067).** `iree_hip_graph_set_host_memcpy_callback_params` looks up `dst_context`/`src_context` via `memory_lookup_range_across_contexts` (which returns *retained* contexts), releases those retains at lines 19057/19062, then stores the now-unretained pointers into `callback_data->dst_context/src_context` at line 19067 — without retaining them for the graph's lifetime. Every other site that stores a remote context pairs the store with `iree_hip_graph_retain_remote_context` (e.g. lines 17338-17339, 17395-17396). The graph-launch host callback later dereferences these pointers (`iree_hip_graph_memcpy_callback` -> `memcpy_device_to_host`/`host_to_device`); if the application destroys the remote context between this update and graph launch, the callback reads freed context memory. Reachable from `hipGraphMemcpyNodeSetParams`, `hipGraphExecMemcpyNodeSetParams`, `hipDrvGraphMemcpyNodeSetParams`. Fix: after the validity check, `iree_hip_graph_retain_remote_context(node->graph, dst_context)` and the same for `src_context` (mirroring line 17339); keep releasing the lookup retains. See also M11.

- **M2 — MEDIUM · confidence medium — graphExec `is_live` gate has no lock/retain (api.c:13573).** `iree_hip_resolve_graph_exec` uses `iree_hal_streaming_graph_exec_is_live(exec)` as the sole validity gate and then returns the raw pointer for every `hipGraphExec*` entry point. `is_live` (graph_exec.c:794) reads `exec->is_destroyed` with no lock and takes no reference, while `graph_exec_destroy_handle` (graph_exec.c:803-822) sets `is_destroyed` under `exec->mutex` and then releases the handle reference, which can free the exec. A thread in `hipGraphLaunch` can pass this check while another completes `hipGraphExecDestroy` -> use of a freed exec; the unlocked `is_destroyed` read is itself a data race against the guarded write. Fix: retain the exec under `exec->mutex` in the resolve gate (gated on `!is_destroyed`) and release after the call.

- **M10 — MEDIUM · confidence medium — device graph-mem attribute/trim ignore the device argument (api.c:2436/2484/2502).** `hipDeviceGetGraphMemAttribute` (2436), `hipDeviceSetGraphMemAttribute` (2484) and `hipDeviceGraphMemTrim` (2502) validate `device` against the device count but then operate on `iree_hip_ensure_context` — the calling thread's *current* context (defaulting to device 0) — instead of the requested device's context. The graph-memory counters are per-context, so Get returns the wrong device's stats and Set/Trim mutate the wrong device whenever the current context's device differs from the argument. The correct resolver already exists: `iree_hip_primary_context_for_device_id` (api.c:7726-7749). Fix: resolve the target context from the `device` argument in all three.

- **M11 — MEDIUM · confidence medium — remote-context retain leaks on memcpy-node re-point / failed update (api.c:17339).** `iree_hip_graph_retain_remote_context` adds a graph-lifetime retain released only at graph destroy. The remote-memcpy update setters call it on every successful update (api.c:17339 and 17396) without releasing the previously-referenced context, and `iree_hip_graph_rollback_memcpy_node_update` (17253) does not undo a retain taken before a failed `iree_hip_graph_update_memcpy_metadata`. So re-pointing a cross-context D2D memcpy node from context A to B leaves A retained until graph destroy though nothing references it, and a failed update leaks the just-retained context. Fix: release the previously-retained remote context on re-point, and release any retain taken there when the subsequent metadata update fails.

- **L1 — LOW · confidence high — graph memory-node flag never cleared on node removal (api.c:16479).** `has_graph_memory_nodes` is set true when a memory node is added (api.c:16479 and 16532) but never re-evaluated when a node is removed: `iree_hal_streaming_graph_destroy_node` (graph.c) doesn't touch it. After adding then removing the last MEM_ALLOC/MEM_FREE node, the flag stays true, so `graph_clone` permanently returns UNIMPLEMENTED and instantiate stays pinned to the one-live-exec path. Also a layering smell that the invariant is maintained in the HIP layer rather than the common add/remove paths. Fix: maintain a memory-node count in the common add/remove paths and derive the flag.

- **L5 — LOW · confidence high — `hipGraphAddChildGraphNode` leaves `*pGraphNode` uninitialized on early error (api.c:16284).** The early validation failures return `hipErrorInvalidValue` without first writing `*pGraphNode = NULL`; a caller that reads the handle after the error gets garbage. Sibling add-node functions set `*pGraphNode = NULL` up front (e.g. `hipGraphAddMemAllocNode` at 16392, `hipGraphAddMemFreeNode` at 16510). Fix: set `*pGraphNode = NULL` immediately after the `!pGraphNode` check.

## Pre-existing issues surfaced during review
These live in code this PR does not modify; I'm flagging them because the review crossed them and a couple are serious. Happy to move any of these to a separate issue instead if you'd rather keep this PR thread focused on the diff.

- **H1 — HIGH · confidence high — capture-teardown deadlock.** `iree_hal_streaming_end_capture` (graph.c:2382) holds the origin `stream->mutex` and, still holding it, calls `has_unjoined_capture_participants` (graph.c:2429), which takes `context->stream_list_mutex` then each participant `stream->mutex`. Every other site orders these the other way (`stream_list_mutex` -> `stream->mutex`, e.g. `clear_capture_participants` graph.c:2124/2129). Two threads each running `hipStreamEndCapture` on different streams of one context can deadlock (T1 holds A's mutex wanting the list lock; T2 holds the list lock wanting A's mutex). The lock-order lines are unchanged by this PR, but the new multi-stream / event-fork capture support likely makes the triggering scenario more reachable. Fix: snapshot the origin's capture state under its mutex, release it, then run the participant scan so the global `stream_list_mutex` -> `stream->mutex` order holds everywhere.

- **M5 — MEDIUM · confidence high — `hipStreamWaitEvent` marks the stream complete early.** `stream_wait_event` (stream.c:647) sets `completed_value = signal_value` right after enqueuing the barrier that only signals once the event fires; `stream_synchronize` then sees `pending_value == completed_value` (stream.c:500) and skips the wait, so a `hipStreamWaitEvent` immediately followed by `hipStreamSynchronize` returns before the event fires. Fix: advance only `pending_value`/`submitted_value` in `wait_event`.

- **M6 — MEDIUM · confidence high — `stream_flush` wedges the stream on submit failure.** `stream_flush` (stream.c:371) increments `pending_value` before `queue_execute`/`flush`, but the failure branch (stream.c:402) doesn't roll it back and the command buffer is released anyway. If `queue_execute` fails, the timeline never reaches `pending_value`, yet later flush/synchronize wait on it forever. Fix: restore `pending_value` on the error path.

- **M12 — MEDIUM · confidence medium — `stream_destroy` touches capture state without the lock.** `stream_destroy` (stream.c:257-259) reads `capture_status` and calls `set_capture_status` with no `stream->mutex` held, and before `unregister_stream` — so the stream is still reachable while a concurrent capture-invalidation scan locks the same mutex and reads/writes `capture_status`. Fix: unregister first, or hold `stream->mutex` around the reset.

- **L3 — LOW · confidence medium — timeline values read lock-free.** `stream_query`/`synchronize`/`wait_submitted` read/update `pending/submitted/completed` (stream.c:451-453, 500, 539) with no lock, while the producers write them under `stream->mutex`. This PR widens exposure: the new `context_synchronize_all`/`context_query` call these from non-producer threads over a stream snapshot. Fix: access the timeline fields under `stream->mutex` (or make them atomics).

- **L2 — LOW · confidence medium (latent) — `end_capture` over-releases a borrowed capture graph.** The success path (graph.c:2480-2481) releases the capture graph when `out_graph == NULL` without checking `capture_graph_owned`, unlike the sibling branches. A `begin_capture_to_graph` capture (which borrows the caller's graph) ended that way would drop a ref it never owned. Not reachable via the present callers, but it's an ownership defect in an exported function.

- **L4 — LOW · confidence high — stale stream-list ownership comment.** internal.h:268-271 says the `streams[]` list is non-owning and streams must unregister before destruction, but `register_stream` retains and `unregister_stream` releases (context.c:641/693) — the list is owning. The comment predates this PR; noting it because it's the invariant a reader leans on when reasoning about the capture/snapshot lifetime. Fix: update the comment.

- **P1 — LOW · confidence medium — symbol-map fast-path race.** `context_symbol_map_lookup` reads the per-context table lock-free while `unregister_module`/`expunge_module` tombstone it under `registry->mutex`; concurrent `__hipUnregisterFatBinary` and symbol lookup race. Predates this PR; the new managed-variable surface widens the chance of concurrent registration/lookup.

## Design question — D1: concurrent event-record nodes
`submit_blocks_locked` (graph_exec.c:2094-2111) increments `event->signal_value`, swaps `event->recording_stream` (retain new / release old) and writes `record_time_ns` holding only the exec and stream mutexes. Two execs that each record the same event, launched concurrently on different streams, would race on these fields (lost increment; concurrent release of the same old `recording_stream` -> refcount double-decrement -> stream UAF). The non-graph `event_record` path has no event-level lock either, and the event has no mutex — so events seem to follow a 'caller serializes' contract. Is concurrent record of one event from two graph launches intended? If so, the event needs internal synchronization.

## Lower-confidence suspicions (worth a glance)
- `device.c` ~340-349: `get_or_create_primary_context`'s error path releases `default_mem_pool`/`current_mem_pool`; if `hipCtxCreate` already populated them (with `primary_context` still NULL), a later failed `get_or_create` could free pools owned by a live `hipCtxCreate` context.
- `memory.c` `free_device_async`: safe only because `hipFreeAsync` never passes a capturing stream; if `launch_host_function`'s capture branch were reached for a deferred free, the already-table-removed wrapper would be freed on every replay (double free). A defensive early-error on capture would harden it.
- A pending stream-ordered async free leaks the wrapper + free-op + retained owner context if the stream/device is torn down with the host call still pending.

libhrx/src/binding/common/graph.c
AWoloszyn: Fixed. The abort path now uses iree_hal_streaming_stream_set_capture_status(...) instead of writing stream->capture_status directly, so status transitions stay consistent with the rest of the capture code.

libhrx/src/binding/common/memory.c
AWoloszyn: Fixed. The cross-context host-local H2D/D2H fast paths now validate the requested range before using the host pointer, so out-of-range offsets cannot overflow/read past the allocation.

libhrx/src/binding/common/memory.c
AWoloszyn: Fixed. Cross-context lookup helpers now keep the owner context retained while returned buffer_refs are dereferenced, including pointer metadata queries, managed/advisable range handling, host device-pointer lookup, memset, and kernel parameter import.

libhrx/src/binding/common/device.c
AWoloszyn: Fixed. Default memory-pool creation is now serialized through the device primary-context mutex, with an internal locked helper used by callers that already hold the lock.

libhrx/src/binding/common/graph_exec.c
AWoloszyn: Fixed. Graph memory high-water accounting now updates from the aggregate current usage, not just the individual contribution size, so multiple active graph-memory entries cannot under-report the high-water value.

libhrx/src/binding/common/graph_exec.c
AWoloszyn: Fixed. hipDeviceGraphMemTrim no longer drops memory still referenced by active graph executables. Trim is limited to reusable/inactive graph-memory reservations, so active graph memory remains accounted for and valid.

AWoloszyn:
Also regarding your other findings:

H2
  Fixed. Graph memcpy callback params now retain remote contexts through the graph user-object ref table before storing them in callback data, so replay/update paths do not hold dangling context pointers.

  M2
  Fixed. Graph exec handles now resolve through a live exec registry and take a retained reference under the graph-exec live check. All graph-exec API paths  release that temporary reference after use, and destroy unregisters the handle only after successful destroy.

  M10
  Fixed. Graph memory get/set/trim paths now operate on the explicit device argument instead of implicitly using the current context/device.

  M11
  Fixed. Remote-context refs held by graph memcpy callback data are now balanced across update paths: new refs are released on rollback/failure, and old refs are released only after the update commits.

  L1
  Fixed. The graph memory-node flag is recomputed/cleared when memory alloc/free nodes are destroyed, so graphs do not remain marked as memory-node graphs after those nodes are removed.

  L5
  Fixed. hipGraphAddChildGraphNode now initializes *pGraphNode = NULL before later validation can fail, so early errors do not leave the output uninitialized.



The other sthat were not introduced by this PR we should flag for further fixes, since this PR is already getting incredibly large.

libhrx/src/binding/common/graph_exec.c
```diff
@@ -771,26 +1519,11 @@ static iree_status_t iree_hal_streaming_graph_record_partition(
...
-      // >1).
-      if (iree_hal_streaming_graph_node_has_recorded_dependency_hazard(
-              node, additional_edges, node_index_map, &barrier_index_set)) {
-        // This node has a dependency on one or more nodes issued between the
-        // last barrier and this. We have to insert an execution barrier to
-        // ensure they complete.
+      const bool has_dependency_hazard =
+          iree_hal_streaming_graph_node_has_recorded_dependency_hazard(
+              node, additional_edges, node_index_map, &barrier_index_set);
+      const bool preserve_sorted_order = true;
```
zjgarvey: **MED · confidence high — replay barrier is now unconditional (serializes independent nodes)**

`preserve_sorted_order = true` here makes the barrier unconditional (the `||` leaves `has_dependency_hazard` dead), where before it fired only on a real hazard. So independent nodes in a partition now serialize instead of overlapping, replayed on every launch. Intentional for the deterministic-order / graph-memory-lifetime reason in the comment above? If so, maybe scope it to partitions with alloc/free nodes (or drive the flag from a condition) so the rest keep the hazard-only path.

zjgarvey (commented):
_Provenance: a follow-up design/perf review pass (Claude Code, Anthropic), checked against the diff before posting. Severity/confidence are my read, not gates._

Two `api.c` items below (diff too large to anchor inline); three more are inline on the common-layer files.

- **MED · confidence high — batch-mem-op graph node returns success but no-ops at execution (api.c:16809).** `hipGraphAddBatchMemOpNode` now returns success, but a BATCH_MEM_OP node isn't recordable or an executed block type, so it falls through to a barrier (graph_exec.c:1838) and does nothing at runtime. It returned `hipErrorNotSupported` at base, like the immediate `hipStreamWriteValue32`/`WaitValue32` still do (api.c:9449/9469). So a graph wait/write-value builds fine then silently no-ops. Suggest rejecting it at add time until execution lands.

- **MED · confidence high — graph mem-alloc nodes allocate eagerly and share one backing across launches (api.c:16651).** `hipGraphAddMemAllocNode` allocates eagerly at construction (api.c:16651) and bakes a fixed `dptr` into the template (16671); MEM_ALLOC/FREE run as no-op barriers (graph_exec.c:1838). So the pointer is live pre-launch and one allocation is shared across executions — diverging from HIP's per-execution re-backing. The one-live-exec cap (graph.c:1992) and relaunch guard (graph_exec.c:2209) cover the worst cases; the residual is the shared `dptr` across re-instantiations and concurrent launches of an exec that frees internally or uses `AUTO_FREE_ON_LAUNCH` (which skips the guard). The `GraphMemAttribute` counters (graph_exec.c:452) are also size-keyed synthetic. Suggest documenting + serializing concurrent launches for now; a stream-ordered pool allocating at execution is the longer-term fix.

libhrx/src/binding/common/stream.c
```diff
@@ -467,6 +471,17 @@ iree_status_t iree_hal_streaming_stream_synchronize(
 
   int status = 0;
   uint64_t timing_step_ns = timing_enabled ? hrx_launch_timing_now_ns() : 0;
+  // HIP launches are logically submitted work even when HRX batches command
+  // buffer recording. Before waiting on one stream, submit other pending
+  // streams so cross-stream or cross-device spin/wait kernels can make forward
+  // progress.
+  iree_status_t flush_all_status = iree_hal_streaming_context_flush_all();
```
zjgarvey: **MED · confidence high — every synchronize triggers a process-wide cross-stream flush**

`stream_synchronize` calls `context_flush_all()` unconditionally here — walks every context under `context_list.mutex` and flushes every stream, with a snapshot alloc. Since `context_synchronize_streams` calls this per stream, a device sync re-runs the whole walk per stream (O(M·N)). The same `synchronize_all()` drain is on sync free (memory.c:792), host-local memcpy (memory.c:1624/1780), and `hipGraphExecDestroy` (graph_exec.c:851), so any of those waits on all contexts. Could hoist the flush to once per top-level sync and wait only the streams that can touch the target.

libhrx/src/binding/common/stream.c
```diff
@@ -630,6 +654,154 @@ iree_status_t iree_hal_streaming_stream_wait_event(
...
+  iree_hal_streaming_context_t* owner_context = NULL;
+  iree_status_t status = iree_hal_streaming_memory_lookup(
+      context, (iree_hal_streaming_deviceptr_t)(uintptr_t)device_ptr,
+      &stream_ref);
+  if (iree_status_is_ok(status) ||
+      iree_status_code(status) != IREE_STATUS_NOT_FOUND) {
+    if (!iree_status_is_ok(status)) return status;
+  } else {
+    iree_status_ignore(status);
+    status = iree_hal_streaming_memory_lookup_range_across_contexts(
```
zjgarvey, line 797: **MED · confidence high — kernel launch scans every context per pointer arg on a local miss**

On a local-table miss, per-pointer-arg resolution falls to `memory_lookup_range_across_contexts` (line 778), an O(contexts) scan under the global `context_list.mutex` (memory.c:441) — paid per imported arg, and a contention point across concurrent launches. (Raw/external pointers return on the first miss at line 854, so the raw-fallback case scans only up to the first.) Could gate the scan on >1 active context, or index pointers in a global map instead of iterating.

zjgarvey (approved):
Thanks, Andrew.

PR #143, opened 2026-06-24
[HAL/AMDGPU] Add device-side queue timestamps

zjgarvey:
`iree_hal_device_queue_timestamp` writes one 64-bit device tick into an 8-byte-aligned transfer target at the point the wait list resolves to in the queue affinity's timeline, signaling the signal list when that point is reached. The queue family specs and `iree_hal_device_spec_timing()` publish `timestamp_valid_bits` and `timestamp_frequency_hz`, but no queue operation wrote a tick from that domain into a caller-supplied buffer at a defined point in the timeline. The slot is mandatory: the HAL dispatches through it with no null check. AMDGPU implements the capture and the replay recorder records a new operation code as unreplayable and forwards the unwrapped target buffer; the CUDA, HIP, local-sync, local-task, Metal, Vulkan, WebGPU, null, and mock devices return `UNIMPLEMENTED`.

## AQL baseline and PM4 fast path

The baseline is a builtin device kernel storing the agent's steady counter to a caller-supplied address, registered at a 1x1x1 workgroup and dispatched over a 1x1x1 grid so a capture is one work-item at one instant. It needs no vendor packet support. Where the queue holds PM4 IB slots and its ISA selects a COPY_DATA timestamp strategy, which the selector supplies for gfx9 through gfx12, submission instead emits one COPY_DATA of the command processor's GPU clock counter, encoding the target address into the IB slot and reserving no kernarg block. Both read the same agent wallclock, so on one agent a tick from either path is interchangeable with the other.

Adding the kernel to the device kernel table makes device setup resolve its `.kd` symbol on every agent and fail on a miss, so all seven prebuilt device objects are regenerated. That rebuild re-emits the library rather than appending one kernel, so most of each object's bytes move.

## Per-agent clock facts

The tick is in the GPU agent's own timestamp domain, and its rate comes from the per-agent `HSA_AMD_AGENT_INFO_TIMESTAMP_FREQUENCY` rather than the system-scope `HSA_SYSTEM_INFO_TIMESTAMP_FREQUENCY`, which describes a different domain and stays where it belongs: converting host-queue epoch wait hints into `hsa_signal_wait` timeouts. Each queue family publishes the rate of the physical device it covers. A zero rate fails device creation at the query and again in spec parameter verification, and because the device-scope timing spec holds one rate for the whole logical device, verification also requires every physical device to agree. Equal rates give a common scale, not a common epoch: agents remain independent counters, so two ticks may only be differenced when both captures resolved to the same domain. Only the low `timestamp_valid_bits` of a tick are defined, so a delta must be reduced modulo that width before dividing by the frequency; AMDGPU publishes 64 at both scopes, which is what makes an unmasked subtraction wrap-correct there.

## Target publication and lifetime

Whether an operation's writes must reach the host follows from its target buffer, not from its semaphores: the buffer-derived release scope is SYSTEM for a `HOST_VISIBLE` target and NONE otherwise. The signal-list scope cannot carry this, since it classifies signal semaphores by their visibility contract and answers AGENT for an empty list, and an AGENT release need not publish outside the agent. The capture takes its submission minimum from the target on both paths; `queue_fill`, `queue_copy` and `queue_update` do the same on the blit dispatch and on the PM4 WRITE_DATA/COPY_DATA paths; the `queue_read` and `queue_write` staging copies take the stronger of it and the scope their transfer direction supplies. A final PM4-IB packet releases at the stronger of that minimum and the signal-list scope.

The target is retained for the operation: submitted inline it is an operation resource, and a capture deferred behind an unresolved wait, an unstaged `queue_alloca` target, or exhausted queue capacity retains it on the pending op until issue. Callers wait on the signal list before reading the tick back.

## Strategy and hardware coverage

A GPU suite is parameterized over both capture strategies, reaching the dispatch by clearing the PM4 strategy on every host queue during device initialization. Two cases pin which branch actually ran by reading the queue's kernarg ring write position, since the dispatch reserves one block and the PM4 submission none. Both take their PM4 expectation from the production selector on the reported ISA, so a queue that should hold the strategy and does not fails rather than skips; those two skip on an ISA with no PM4 timestamp packet. Three of the sixteen cases do not vary with the parameter and run once, and twelve run in both on any hardware; the sixteenth runs in both wherever the ISA provides the packet.

The advertised frequency is pinned against the agent attribute at device and queue-family scope, and against a rate measured from two captures 200ms apart, to within a factor of 1.05. The cross-strategy case requires a dispatch tick to land between two PM4 ticks and the two measured rate brackets to intersect. Capture targets cover a mapped allocation, a nonzero offset, a `queue_alloca` transient, and an agent-scope signal list; rejection cases cover misalignment, range, usage, flags, and an unstaged `queue_alloca` target.

GPU-free coverage pins the spec builder's per-family and device-scope rates and its rejection of zero and mismatched rates, the buffer release-scope policy including the large-BAR shape of device-local storage the host can still map, and the capture kernel's packet, kernargs, and single-work-item table registration.


runtime/src/iree/hal/device.h
```diff
@@ -1220,6 +1255,15 @@ typedef struct iree_hal_device_vtable_t {
   iree_status_t(IREE_API_PTR* queue_flush)(
       iree_hal_device_t* device, iree_hal_queue_affinity_t queue_affinity);
 
+  // Optional: may be NULL if the device does not support device-side timestamp
```
benvanik: no NULL vtable entries - it must be implemented in all targets, even if as a stub

also, keep this in order - it must follow queue_execute

runtime/src/iree/hal/device.h
```diff
@@ -939,6 +939,41 @@ IREE_API_EXPORT iree_status_t iree_hal_device_queue_barrier(
...
+// |target_buffer|.
+//
+// Returns IREE_STATUS_UNIMPLEMENTED if the device does not support device-side
+// timestamp capture; callers should fall back to a host-observed timestamp.
+// The IREE_HAL_DEVICE_TIMING_SPEC_FLAG_DEVICE_TIMESTAMPS device-spec flag is
+// only a hint (a necessary-but-not-sufficient signal): a specific queue may
+// still return IREE_STATUS_UNIMPLEMENTED, which is the authoritative per-queue
+// answer, so callers must handle UNIMPLEMENTED by falling back to a
+// host-observed timestamp even when the flag is set.
+IREE_API_EXPORT iree_status_t iree_hal_device_queue_capture_timestamp(
```
benvanik: rename to `iree_hal_device_queue_timestamp` (here and elsewhere)

we cannot use status as control flow - this has to be a capability query. status carries backtraces. remove the entire UNIMPLEMENTED commentary

runtime/src/iree/hal/drivers/amdgpu/logical_device.c
```diff
@@ -2931,6 +2931,22 @@ static iree_status_t iree_hal_amdgpu_logical_device_queue_copy(
                              target_offset, length, flags);
 }
 
+static iree_status_t iree_hal_amdgpu_logical_device_queue_capture_timestamp(
```
benvanik: ensure, as everywhere, that all methods follow the same exact ordering

benvanik (changes requested):

runtime/src/iree/hal/device.h
```diff
@@ -945,6 +945,36 @@ IREE_API_EXPORT iree_status_t iree_hal_device_queue_execute(
...
+// The written value is in the device timestamp domain. Convert a delta between
+// two timestamps to nanoseconds by dividing by the tick frequency reported in
+// iree_hal_device_spec_timing()->timestamp_frequency_hz. Callers must wait on
+// |signal_semaphore_list| before reading the tick back from |target_buffer|.
+//
+// Device-side timestamp support is a capability discovered via the device spec:
+// it is available only when iree_hal_device_spec_timing() reports the
+// IREE_HAL_DEVICE_TIMING_SPEC_FLAG_DEVICE_TIMESTAMPS flag. Callers must query
+// that flag and only enqueue this operation on devices that advertise it.
+IREE_API_EXPORT iree_status_t iree_hal_device_queue_timestamp(
```
benvanik, line 1002: The AMDGPU implementation is a useful backend foundation, and a queue-level
timestamp primitive is the right abstraction for fixing HIP event timing. The
remaining blocker is that the PR stops before the production contract that
motivated it. At this head, the only call to
`iree_hal_device_queue_timestamp` outside its direct HAL test is replay
forwarding. The shipping path remains:

`hipEventRecord` → `iree_hal_streaming_event_record` → `iree_time_now()` +
`iree_hal_device_queue_barrier`

`hipEventElapsedTime` then subtracts those host enqueue timestamps, and graph
`EVENT_RECORD` blocks still lower to `iree_hal_device_queue_barrier`.
Consequently MIOpen retains the existing near-zero host-enqueue spacing, and
the external HIP suite cannot exercise this operation through HIP.

The consumer developed in #170 is useful starting material. It needs these
corrections before it is the production witness for this API:

- A timing-enabled streaming event owns one host-visible, transfer-target
  timestamp slot; `DISABLE_TIMING` events retain the allocation-free barrier
  path.
- Direct records and graph-replayed event-record nodes enqueue
  `queue_timestamp` at the event's dependency point. Graph records must not
  silently fall back to host enqueue time.
- Elapsed-time conversion uses `timestamp_valid_bits` as well as the
  frequency, including counter wrap. The signed subtraction in #170 is not
  sufficient for a narrower timestamp domain.
- The event semaphore generation remains event-owned and monotonic across
  streams. The direct path currently derives `event->signal_value` from
  `stream->pending_value`, so re-recording an event after moving from a
  high-valued stream to a fresh stream can select an already-satisfied event
  value; graph replay already increments the event's own value.

The HAL primitive can remain its own reviewable change, but it should be
developed as a paired stack with this consumer and its external HIP acceptance
coverage. That keeps the HAL contract coherent while ensuring we do not land a
generic operation whose stated user-visible behavior remains unchanged.

runtime/src/iree/hal/device.h
```diff
@@ -945,6 +945,36 @@ IREE_API_EXPORT iree_status_t iree_hal_device_queue_execute(
...
+// |target_offset|. The target must be 8-byte aligned and be a transfer target
+// the device can write 8 bytes into; to read the tick back on the host the
+// buffer must additionally be host-visible (e.g. mappable).
+//
+// The written value is in the device timestamp domain. Convert a delta between
+// two timestamps to nanoseconds by dividing by the tick frequency reported in
+// iree_hal_device_spec_timing()->timestamp_frequency_hz. Callers must wait on
+// |signal_semaphore_list| before reading the tick back from |target_buffer|.
+//
+// Device-side timestamp support is a capability discovered via the device spec:
```
benvanik: This cannot be designed as a capability-gated operation.
`iree_hal_device_queue_timestamp` is a primary
`iree_hal_device_queue_*` method and a required device-vtable slot, so the HAL
contract is that every device implements it. Callers must not probe an
unrelated timing flag or call the method and interpret `UNIMPLEMENTED` as
feature detection.

The CUDA, HIP, local-sync, local-task, Metal, null, Vulkan, WebGPU, and mock
stubs in this PR are therefore temporary compilation scaffolding, not the
capability model for this API. We can complete those backend implementations
in follow-up changes; this PR does not need to take on all of that work.
Please remove the capability-query paragraph here so we do not establish the
wrong public contract in the meantime.

The blocking scope in this PR is AMDGPU, and it must implement the operation
for every AMDGPU queue. PM4 is an optional optimization strategy, not the
baseline implementation. The current code returns `UNIMPLEMENTED` whenever
`pm4_timestamp_strategy` is absent and then changes the device spec to hide
that implementation gap. That reverses the dependency: queue timestamp
semantics must exist independently, and PM4 may accelerate them when the
selected architecture supports the packet.

There is a fairly direct AQL path available:

- Add a one-workitem builtin beside `device/timestamp.c` that stores
  `iree_amdgpu_device_timestamp` (`__builtin_readsteadycounter`) into the
  target address.
- Register it in the existing AMDGPU builtin kernel table and load it with the
  other device blit/timestamp kernels.
- Make the host queue submit that AQL dispatch after the requested waits and
  complete the requested signals after the write. The target buffer must be
  retained across both immediate and deferred issue.
- Select the existing PM4 packet only as the fast path when a PM4 timestamp
  strategy is available; otherwise use the AQL builtin. Neither path returns
  `UNIMPLEMENTED`.

Please also restore the existing AMDGPU timing-domain facts. The HSA timestamp
frequency and `DEVICE_TIMESTAMPS` flag describe a real device clock already
used by profiling and semaphore timing; they must not be cleared merely
because one PM4 write-to-memory mechanism is unavailable.

The tests should make the strategy boundary explicit: exercise the AQL path
with PM4 timestamp support forced absent, exercise the PM4 fast path when
available, and run the same ordering/target/lifetime assertions against both.
That gives us a complete AMDGPU implementation now while leaving the other
backend implementations as well-scoped follow-up work.

runtime/src/iree/hal/drivers/amdgpu/host_queue_timestamp_test.cc
```diff
@@ -0,0 +1,513 @@
...
+      iree_hal_buffer_map_read(tick1, 0, &stop_tick, sizeof(stop_tick)));
+
+  EXPECT_NE(start_tick, 0u);
+  EXPECT_NE(stop_tick, 0u);
+  ASSERT_GT(stop_tick, start_tick);
+  // ASSERT_GT(stop_tick, start_tick) above already guarantees elapsed_ms > 0;
+  // only the upper-bound sanity ceiling adds information here.
+  const double elapsed_ms =
+      (double)(stop_tick - start_tick) * 1000.0 / (double)frequency_hz;
+  EXPECT_LT(elapsed_ms, 10000.0);
```
benvanik: Please remove the 10-second ceiling rather than increasing or relaxing it.
The operation already waits on explicit completion and asserts ordered ticks;
valid asynchronous work has no ten-second completion contract. This assertion
turns ordinary machine load into a low-probability failure and will gradually
poison the suite with flakiness without adding correctness evidence. Use an
infinite wait here and let the outer test harness diagnose a genuine hang.

benvanik (changes requested):

runtime/src/iree/hal/device.h
zjgarvey: Removing the paragraph. Worth knowing the model is already broken as written: HIP and Vulkan both advertise `DEVICE_TIMESTAMPS` (`hip/device_spec_builder.c:470`, `vulkan/device_spec_builder.c:472`) while their stubs return `UNIMPLEMENTED` and claim they do not advertise it. All the stub comments are corrected.

AQL baseline implemented, PM4 demoted to a fast path, neither returns `UNIMPLEMENTED`. Both were verified to read the same counter, and the tests run both paths. Timing facts restored.

Chasing the release scope on the AQL path turned up the same gap in `queue_fill` and `queue_copy`, on both their PM4 and dispatch paths: the scope was derived from the signal list, which answers what the semaphore edge needs rather than who reads the bytes the packet wrote. It is now derived once from whether the target is host-visible and applied at every target-writing blit path.

One question, since the flag is yours: should `DEVICE_TIMESTAMPS` mean "this device has a timestamp domain at a known rate" — its original sense, and still what HIP and Vulkan set it for — or "`queue_timestamp` is implemented"? Both meanings are live in the tree right now, and that is what produced the contradiction above.

runtime/src/iree/hal/device.h
zjgarvey: Opened as #210, stacked on this one, with the graph path included — the earlier version silently fell back to host time on replay, which is removed. Its acceptance test asserts a lower bound around known-duration device work, which is the only assertion shape that separates a device measurement from host enqueue spacing.

On valid bits: AMDGPU reports 64 at 100 MHz, so wrap is roughly 5800 years and modular arithmetic there would be speculative. The narrow-counter case now fails loud instead of proceeding, since HIP's own builder advertises 32.

The signal value point turned out worse than a timing concern. On main, re-recording an event across streams can also hang a stream indefinitely, and neither that nor the stale completion surfaces an error. Both are covered by regression tests that fail against main, and the fix is #211 against main rather than riding along here.

External coverage is running too: TheRock's prebuilt hip-tests execute against `libhrx`, with 27 of 29 of the CI smoke tier passing. It already found that `hipEventQuery` on an invalid handle segfaults where the contract requires `hipErrorInvalidHandle`; that is a separate fix.

runtime/src/iree/hal/drivers/amdgpu/host_queue_timestamp_test.cc
zjgarvey: Removed. Replaced with an assertion that pins the tick rate against a measured one, which is load-insensitive by construction. Worth doing because the ceiling was the only quantitative check in the file, and it was weak enough to hide the 10x frequency error raised separately.

zjgarvey:
Separate from the review, and worth flagging on its own: the advertised tick frequency was the wrong clock. We published `HSA_SYSTEM_INFO_TIMESTAMP_FREQUENCY`, which ROCr derives from the host clock (1 GHz here), while the tick the device writes runs at the per-agent wallclock (100 MHz on MI300X). Every duration computed per the documented conversion came out 10x short. It is now sourced per physical device from `HSA_AMD_AGENT_INFO_TIMESTAMP_FREQUENCY`, with a test that pins the advertised rate against a measured one — the old suite could not catch this because its only quantitative assertion was a one-sided ceiling.

benvanik (changes requested):
The AMDGPU mechanism is now on the right architecture: every AMDGPU queue has an AQL capture baseline, PM4 is only an optional fast path, both paths use the existing submission/reclaim machinery, and the target lifetime and host visibility requirements are handled correctly. I do not think this needs another implementation redesign.

There are still several contract and cleanup issues to resolve before this can land.

`iree_hal_device_queue_timestamp` is a required primary queue operation. Please remove the public `UNIMPLEMENTED` contract and the mock-device test that establishes unsupported devices as valid behavior. The non-AMDGPU stubs remain temporary compilation scaffolding; they must not become the feature-detection model.

`DEVICE_TIMESTAMPS` should mean only that the primary `timestamp_valid_bits` and `timestamp_frequency_hz` facts are meaningful. It must say nothing about whether a required queue operation exists.

The timestamp documentation also needs to describe the actual domain. AMDGPU emits an agent-domain tick, and ROCr performs per-agent frequency/offset conversion to the system domain. Two agents reporting the same frequency therefore have a common scale, not necessarily a common epoch. Timestamps may only be differenced when both captures resolve to the same timestamp domain; callers using multi-physical-device logical devices must constrain captures accordingly. Only the low `timestamp_valid_bits` bits are defined, and deltas must be computed modulo that width before conversion. Please update the shared-clock comments and test names to match. Keeping the equal-frequency validation is reasonable because it makes the single device-level rate usable as a scale for every same-agent delta, but it does not prove a shared clock.

Please remove `suppress_pm4_timestamps` from the public logical-device options, physical-device options, and CLI. It exists only to parameterize this test and changes unrelated profiling and command-buffer behavior. The internal test harness can force the initialized host queues’ timestamp strategy to `NONE` before their first submission, exercising the same production selection branch without creating shipping configuration.

The measured-rate tolerance also needs tightening. A multiplicative tolerance of `2.0` accepts a frequency anywhere from half to twice the measured bracket, including an exact 2× error. The bracket already accounts for queue and host scheduling latency; a 5% relative tolerance is ample.

Finally, please rebase on current main and rewrite the history around the final design. The current 29 commits preserve several rejected capability models, PM4-only behavior, repeated comment reversals, machine-local verification notes, and a session URL. The PR description likewise still describes capability-gated, PM4-only behavior and a valid-bits TODO. It should instead explain the required HAL operation, AQL baseline/PM4 fast path, per-agent clock facts, target publication/lifetime, and the strategy/hardware coverage now present. Please also drop the unrelated `physical_device_capabilities` build-target extraction from this change.

zjgarvey:
The history is rewritten around the final design and rebased on main. All of the last review comments should be addressed.

A separate finding from a last self-review: the AMDGPU capture now also defers when the target is an unstaged `queue_alloca` buffer. It previously deferred only on an unresolved wait, so a capture into a freshly allocated target could submit ahead of its allocation. That is in `[HAL] Require device-side queue timestamps` and the hardware suite covers it.

benvanik (approved):
The timestamp architecture and asynchronous ownership paths now look solid. I traced immediate and deferred submission, prerequisite failure and cancellation, target publication and reclaim, both capture strategies, clock-domain reporting, replay behavior, and the hardware coverage without finding another design blocker.

One landing gate remains: the branch no longer builds when merged with current main because `runtime/src/iree/hal/drivers/amdgpu/device_spec_builder_test.cc` and `runtime/src/iree/hal/drivers/amdgpu/host_queue_timestamp_test.cc` still reference the removed AMDGPU target-ID representation. Current main replaced that representation with structured target identity; GitHub reports the branch mergeable because these stale references do not textually conflict. Please rebase and port those two references before landing. This is the same current-main gate from the previous round: main moved shortly after the rewrite, so there is no new design work here, but we need to hold the line that the resulting merge builds.

I verified the correction is bounded. After applying only those two identity ports to the current-main merge, all 1,807 `//runtime/src/iree/hal/...` targets build under ASAN, the gfx1100 hardware suite exercises both PM4 and dispatch capture strategies, the focused device-spec, policy, builtin, and replay tests pass, Bazel-to-CMake regeneration is unchanged, and all seven embedded device objects contain the expected capture kernel and clock read.

Approving the design and implementation subject to that rebase correction and green CI.

zjgarvey:
I don't think I should wait for 1150 lest main advances again. 

The handfull of loom changes since the last rebase are unrelated, so I'm going to squash and merge now.

PR #150, opened 2026-06-26
hip: resolve kpack (HIPK) out-of-band code objects in-tree

zjgarvey:
kpack-packaged ROCm builds strip device code out of the host .so into sibling ".kpack" archives and hand the runtime a "HIPK" fat-binary wrapper whose binary pointer is msgpack metadata — a kernel name plus candidate .kpack search paths — rather than an inline code object.

A new `libhrx/src/binding/common/kpack.{c,h}` module resolves that metadata into the real AMDGPU code object at HIP module registration, with no dependency on the vendor librocm_kpack loader. It memory-maps each candidate .kpack archive and parses the 16-byte "KPAK" header and msgpack table of contents in place, so a lookup touches only the header, the table of contents, and the one requested kernel's bytes; the resolved code object is an independent copy. It provides a bounds-checked MessagePack reader, the "none" and "zstd-per-kernel" storage schemes, ISA feature-flag-subset target matching, binary-relative search-path resolution with "@GFXARCH@" expansion and /proc/self/maps discovery, and the ROCM_KPACK_PATH / _PATH_PREFIX / _DISABLE / _DEBUG environment overrides. `fat_binary.c` recognizes the HIPK magic and routes the resolved object back through the existing ELF / Clang-offload-bundle extraction path.

Only the runtime unpacking side is implemented; HRX is not a compiler and never produces kpack archives. The on-disk formats mirror the ROCm kpack reference.


libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1329 @@
...
+  }
+  *out_code_object = NULL;
+  *out_code_object_size = 0;
+
+  if (kpack_env_flag("ROCM_KPACK_DISABLE")) {
+    return iree_make_status(IREE_STATUS_UNAVAILABLE,
+                            "kpack disabled via ROCM_KPACK_DISABLE");
+  }
+
+  // 1. Parse the HIPK metadata blob (self-bounding within the cap).
```
AWoloszyn: One issue we can run into here is that we don't know if +64KB is actually valid (I am sure we have similar problems in fat binary parser). 
I think we can use msync() to make sure that all of the pages we are about to try to read are actually mapped into this memory space, alternatively we can read /proc/self/maps (although that's a bit more annoying).  Otherwise we might just read out of our memory spac.e

libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1329 @@
...
+  memcpy(state->arch[state->arch_count], target.data, target.size);
+  state->arch[state->arch_count][target.size] = '\0';
+  ++state->arch_count;
+  return false;  // continue collecting
+}
+
+// Maps and parses the archive at |path| into the resolve's open list
+// (deduplicated by canonical path). Missing or unparseable archives are skipped
+// (logged), not fatal.
+static void kpack_try_open_archive(kpack_resolve_state_t* state,
```
AWoloszyn: This treats any failure as skips rather than actual failures. Is this what we want? E.g. not being able to mmap a file (due to permissions) is different than it not being there at all. It could hide errors that are a bit annoying for the application to diagnose. (if the file exists but has the wrong permission bits set, the user wouldn''t know why a kernel is not found)

libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1329 @@
...
+  // cannot be resolved, and yields the canonical path used to dedup archives.
+  char canonical[KPACK_PATH_MAX];
+  if (!realpath(path, canonical)) {
+    KPACK_DBG("archive not found: %s", path);
+    return;
+  }
+  for (iree_host_size_t i = 0; i < state->open_count; ++i) {
+    if (strcmp(state->open[i].canonical, canonical) == 0) return;  // dedup
+  }
+  if (state->open_count >= KPACK_MAX_OPEN_ARCHIVES) {
```
AWoloszyn: Do we really want to ignore the 65th archive? If the user does not have kpack_debug turned on it will be hard to diagnose.

libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1329 @@
...
+// (deduplicated by canonical path). Missing or unparseable archives are skipped
+// (logged), not fatal.
+static void kpack_try_open_archive(kpack_resolve_state_t* state,
+                                   const char* path) {
+  // realpath() requires a buffer of at least PATH_MAX bytes; KPACK_PATH_MAX is
+  // sized to match (== PATH_MAX on Linux, the only platform that reaches here).
+  // realpath() also doubles as the existence check: it returns NULL if |path|
+  // cannot be resolved, and yields the canonical path used to dedup archives.
+  char canonical[KPACK_PATH_MAX];
+  if (!realpath(path, canonical)) {
```
AWoloszyn: This wont compile under msvc (I am sure we have other issues like this, but we should ideally be able to compile there even if we don't funciton correctly yet).

libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1589 @@
...
+    return iree_make_status(IREE_STATUS_OUT_OF_RANGE,
+                            "kpack base+relative path too long");
+  }
+  iree_host_size_t len = 0;
+  memcpy(joined + len, dir.data, dir.size);
+  len += dir.size;
+  joined[len++] = '/';
+  memcpy(joined + len, relative.data, relative.size);
+  len += relative.size;
+  return kpack_normalize_path(iree_make_string_view(joined, len), out,
```
AWoloszyn, line 970: I think there is an issue around here with symlinks.

if symlink points to /bar/baz:
`foo/symlink/../archive.kpack` would resolve to foo/archive.kpack, and if I am understanding correctly, it should really point to /bar/archive.kpack 
Which is somewhat insane to me, but IIUC that is the expected behavior.

libhrx/src/binding/common/kpack.c
```diff
@@ -0,0 +1,1589 @@
...
+  const iree_host_size_t n = feature_count;
+  if (n == 0) {
+    char buf[IREE_HAL_STREAMING_KPACK_TARGET_CAPACITY];
+    memcpy(buf, processor.data, processor.size);
+    // One candidate, so there is no later iteration for an early stop to skip.
+    (void)callback(iree_make_string_view(buf, processor.size), user_data);
+    return iree_ok_status();
+  }
+
+  // Power set of features, descending mask (most specific first). Bit
```
AWoloszyn: This enumerates descending numeric bitmasks. With three features, this produces cardinalities 3, 2, 2, 1, 2..., so a one-feature archive can win before a more-specific two-feature archive.

libhrx/src/binding/common/kpack.c
zjgarvey: Fixed, thanks for the catch.

AWoloszyn (approved):
LGTM Thanks for the fixes!

PR #151, opened 2026-06-27
[Loom] Fix AMDGPU FP8 WMMA blocker paths

benvanik:
This closes two FP8/BF16 WMMA correctness gaps in the AMDGPU path: compact FP8 RHS storage can now feed BF16 WMMA operands directly, and workgroup barriers after WMMA no longer allow unrelated VMEM hazards to be treated as drained.

`vector.fragment.load` is now symmetric with `vector.fragment.store`: fragment movement is a target-shaped boundary, so the storage element type in the view and the physical payload element type may differ when target lowering explicitly selects the conversion. Ordinary vector memory operations still keep their same-element contract. The verifier coverage now documents this load-boundary conversion contract instead of rejecting it generically before the AMDGPU target can make a legality decision.

AMDGPU fragment lowering has a new FP8-to-BF16 load payload form for LHS/RHS matrix operands. The selected route uses scalar FP8 byte loads from global/buffer memory, expands each byte through the low-level arithmetic packet vocabulary, rounds pairs of f32 decode lanes into packed BF16 registers, and feeds the existing BF16 WMMA operand layout. This gives authors a direct compact-weight shape such as `vector.fragment.load<rhs> ... view<...xf8E4M3> -> vector<...xbf16>` without hand-authored staging through LDS or scalar unpack sludge.

The wait planner now interprets wait-counter immediate values before applying packet counter effects. A wait such as `s_waitcnt {lgkmcnt = 0}` resets LDS/SMEM progress without pretending that VMEM loads are complete. That preserves required VMEM waits across workgroup barriers and fixes the stale-input failure mode where FP8 scratch/WMMA kernels could produce NaNs after a post-MMA barrier.

The regression surface covers the IR contract, the selected AMDGPU source-to-low sequence, barrier-adjacent fragment epilogues, and the wait-plan JSON that proves lgkm-only waits leave VMEM hazards live.

### DIFF
--- a/loom/py/loom/dialect/vector/defs.py
+++ b/loom/py/loom/dialect/vector/defs.py
@@ -1389,7 +1389,11 @@ vector_fragment_load = Op(
         "shape, view layout, and target legality; it is not an ordinary "
         "trailing-axis footprint of the view. The result carries fragment "
         "facts directly so vector.mma can consume it without a separate "
-        "vector.fragment wrapper."
+        "vector.fragment wrapper. When the view and result element types "
+        "differ, the operation represents a fragment-shaped numeric "
+        "conversion at the load boundary and target lowering must either "
+        "select that conversion explicitly or reject it with target "
+        "diagnostics."
     ),
     operands=[
         Operand("view", VIEW, doc="Typed source view holding logical matrix data."),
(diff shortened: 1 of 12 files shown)


benvanik:
Closing this as superseded by the implementation now on `main`. The wait-counter correctness fix landed separately in 057eab18c48, and the FP8-to-BF16 fragment path was subsequently replaced and refined by the compact/exact fragment lowering work beginning with 9c6a73d91dd and 878046ae538. Rebasing this combined branch would reintroduce the older lowering shape rather than preserve a useful independent delta.

PR #152, opened 2026-06-29
[hip] Track HIP stream and primary context state

AWoloszyn:
Store stream synchronization policy and CU-mask state on streaming streams, release owned CU-mask storage during stream teardown, and update primary context flags under the primary-context lock.

Report primary-context active state from the live context handle so implicit contexts and retained contexts share the same state contract.

### DIFF
--- a/libhrx/src/binding/common/device.c
+++ b/libhrx/src/binding/common/device.c
@@ -223,11 +223,12 @@ iree_status_t iree_hal_streaming_device_set_primary_context_flags(
                          "invalid device ordinal %" PRIhsz, device_ordinal));
   }
 
-  // Update the primary context flags.
+  iree_slim_mutex_lock(&device->primary_context_mutex);
   device->primary_context_flags = *flags;
-
-  // If the primary context exists, we might need to update it.
-  // For now, we just store the flags for new contexts.
+  if (device->primary_context) {
+    device->primary_context->flags = *flags;
+  }
+  iree_slim_mutex_unlock(&device->primary_context_mutex);
 
   IREE_TRACE_ZONE_END(z0);
   return iree_ok_status();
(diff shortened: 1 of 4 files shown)


libhrx/src/binding/common/device.c
```diff
@@ -247,16 +248,15 @@ iree_status_t iree_hal_streaming_device_primary_context_state(
...
   }
 
+  iree_slim_mutex_lock(&device->primary_context_mutex);
   if (out_flags) {
-    *out_flags = device->primary_context_flags;
+    *out_flags = device->primary_context ? device->primary_context->flags
+                                         : device->primary_context_flags;
   }
-
-  // Context is active if reference count > 0.
```
zjgarvey, line 254: nit: I think the callsite in api.c:2940 still has this comment (which isn't true anymore iiuc).

libhrx/src/binding/common/internal.h
```diff
@@ -466,9 +473,16 @@ typedef struct iree_hal_streaming_stream_t {
...
+  // HIP stream creation flags.
   iree_hal_streaming_stream_flags_t flags;
+  // HIP synchronization policy value for stream attribute queries.
+  iree_hal_streaming_synchronization_policy_t synchronization_policy;
+  // HIP stream scheduling priority hint.
   int priority;
+  // Number of 32-bit entries in |cu_mask|.
+  iree_host_size_t cu_mask_count;
+  // Optional HIP CU mask owned by this stream; NULL means default device mask.
+  uint32_t* cu_mask;
```
zjgarvey, line 488: Where are the cu mask attrs used? Are `hipExtStreamCreateWithCUMask` or `hipExtStreamGetCUMask` going to be modified in a follow up PR or something? Or are you just adding the attrs to avoid something crashing hard on some external test?

If the latter, I think maybe the docstring could indicate that these vars aren't actually wired up yet.

zjgarvey (approved):
Small nits, but I don't really get why the cu attrs are added if they are dead.

libhrx/src/binding/common/internal.h
AWoloszyn: There are scaffolding for future changes. The PR split that I had was a bit wonky, but everything should get used soon.

libhrx/src/binding/common/device.c
AWoloszyn: Fixed.

PR #153, opened 2026-06-29
Propagate AMDGPU kernel resource metadata

AWoloszyn:
Carry max workgroup size, fixed local/private memory usage, and uniform-workgroup metadata from HSACO messagepack into executable export info. Preserve custom-direct kernarg layouts with implicit argument offsets so launch paths can separate caller-provided payload from runtime-owned hidden arguments.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -1726,7 +1726,17 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
     uint8_t* tail_payload) {
   switch (kernarg_storage_mode) {
     case IREE_HAL_AMDGPU_COMMAND_BUFFER_KERNARG_STORAGE_MODE_CUSTOM_DIRECT: {
-      const iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
+      iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
+      if (explicit_bytes == 0) {
+        // Zero means dynamic explicit bytes; with an implicit suffix, user
+        // kernargs end where the runtime-owned hidden args begin.
+        explicit_bytes = layout->has_implicit_args
+                             ? layout->implicit_args_offset
+                             : layout->total_kernarg_size;
+        if (explicit_bytes == 0) {
+          explicit_bytes = constants.data_length;
+        }
+      }
       const iree_host_size_t copy_bytes = constants.data_length < explicit_bytes
                                               ? constants.data_length
                                               : explicit_bytes;
--- a/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
+++ b/runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
@@ -1729,7 +1729,10 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
       iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
       if (explicit_bytes == 0) {
         // Zero means dynamic explicit bytes; with an implicit suffix, user
-        // kernargs end where the runtime-owned hidden args begin.
+        // kernargs end where the runtime-owned hidden args begin. Executable
+        // loading rejects metadata where visible args resume after that suffix,
+        // and plan construction reserves enough bytes for the explicit prefix
+        // and any suffix written below.
         explicit_bytes = layout->has_implicit_args
                              ? layout->implicit_args_offset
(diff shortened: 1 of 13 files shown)


runtime/src/iree/hal/drivers/amdgpu/aql_command_buffer.c
```diff
@@ -1726,7 +1726,17 @@ static iree_status_t iree_hal_amdgpu_aql_command_buffer_write_dispatch_tail(
...
-      const iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
+      iree_host_size_t explicit_bytes = layout->explicit_kernarg_size;
+      if (explicit_bytes == 0) {
+        // Zero means dynamic explicit bytes; with an implicit suffix, user
+        // kernargs end where the runtime-owned hidden args begin.
+        explicit_bytes = layout->has_implicit_args
+                             ? layout->implicit_args_offset
+                             : layout->total_kernarg_size;
+        if (explicit_bytes == 0) {
+          explicit_bytes = constants.data_length;
```
zjgarvey: I don't really get all of this. 

Is there a guard somewhere else against memcpy-ing `copy_bytes > total_kernarg_size`? 

Is it possible that there are explicit args _after_ the hidden block? Like [visible][hidden][visible]? I don't even know how that gets passed into `constants`, but I can't figure out how that would work here. I looked around for how this is done without command buffers, and I wonder if it makes sense to have a shared helper between this and the packers in `device/dispatch.c`, since I assume those should have the same behavior.

AWoloszyn: Good call. I added a clarification commit that documents this as a layout invariant instead of leaving it implicit.

  The visible/hidden/visible case is rejected while loading executable metadata: metadata-derived fixed layouts require all visible arguments to end at or before the implicit suffix offset. The command-
  buffer writer can therefore treat the caller payload as a visible prefix and write the runtime-owned implicit suffix after it.

  The memcpy bound is guarded by that same prevalidated layout contract: plan construction reserves enough storage for the explicit prefix plus any implicit suffix, and the write path clamps the caller copy
  to the resolved explicit byte count.

  I also removed the stale comment in executable descriptor construction that implied visible args might appear after the first hidden record.

zjgarvey: I'm still trying to understand the full picture:

1. we call `hipModuleLaunchKernel(f, ..., kernelParams=NULL, /*extra=*/{<pointer-flag>, blob, <size-flag>, size_ptr, <end-flag>})`
2. we set `params_ptr=blob` and `params_size=size` and set the dispatch flags `PRE_PACKED`. 
3. in the stream launch path, where `is_pre_packed && params->buffer`, we have a nonzero `buffer_size`, but `parameters.constant_bytes = 0`, so we set `constants = params->buffer` and `size = params->buffer_size` (unclamped).
4. When preparing the plan in `aql_command_buffer.c`, under the custom direct path, the `required_explicit_bytes` is set to `explicit_kernarg_size` which is 0. So the check against `padded_constants_length < required_explicit_size = 0` is automatically false. Is this bound the one you are referring to in the following comment?
> "The memcpy bound is guarded by that same prevalidated layout contract: plan construction reserves enough storage for the explicit prefix plus any implicit suffix, and the write path clamps the caller copy
to the resolved explicit byte count."
5. in the `write_dispatch_tail` code, this case hits `explicit_bytes=implicit_args_offset=0` and falls through to `explicit_bytes = data_length = size` hitting `memcpy(tail_payload, blob, size)`. 

As far as I'm aware, this memcpy `size` isn't bounded above anywhere in this code path, and the `tail_payload` is set based on fixed kernel metadata.

AWoloszyn: The previous explanation missed the pre-packed path: with `explicit_kernarg_size == 0` and `implicit_args_offset == 0`, the AQL writer
  could still resolve the copy size to `constants.data_length`, so the memcpy was not bounded by the fixed metadata reservation.

  I fixed this by centralizing explicit-prefix size resolution and using it in AQL validation/copying, host-queue validation, and the device packer. For
  implicit-suffix layouts, caller-owned bytes are only `[0, implicit_args_offset)`, so offset 0 copies zero bytes and leaves the suffix runtime-owned.

  Added regression coverage for that case.

zjgarvey (approved):
I'm not sure if you should land on my approval alone, but I personally think this is good.

benvanik:
Superseded by #360, which carries forward the valid AMDGPU kernel resource metadata and custom kernarg bounds through the generic HAL executable-function API with the original authorship preserved. The uniform-workgroup flag is intentionally not carried forward because HAL does not yet have the exact-global-size dispatch contract required to represent it correctly.

PR #154, opened 2026-06-29
Implement HIP array allocation and descriptors

AWoloszyn:
Add process-local HIP array handles backed by device allocations, descriptor conversion for runtime and driver array APIs, and array-aware 2D/3D copy normalization. Add mipmapped-array handle management and batch array-copy validation using the existing IREE synchronization primitives.

### DIFF
--- a/libhrx/src/binding/hip/abi_stubs.c
+++ b/libhrx/src/binding/hip/abi_stubs.c
@@ -9,11 +9,15 @@
 // preserving a loud unsupported result if one of these paths is executed.
 
 #include <dlfcn.h>
+#include <limits.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "iree/base/api.h"
+#include "iree/base/threading/call_once.h"
+#include "iree/base/threading/mutex.h"
 #include "libhrx/src/binding/hip/api.h"
 
 // Local compatibility declarations for unsupported ABI entries. These symbols
@@ -31,7 +35,10 @@ typedef struct HIP_LAUNCH_CONFIG_st HIP_LAUNCH_CONFIG;
 typedef struct HIP_RESOURCE_DESC HIP_RESOURCE_DESC;
 typedef struct HIP_RESOURCE_VIEW_DESC HIP_RESOURCE_VIEW_DESC;
 typedef struct HIP_TEXTURE_DESC HIP_TEXTURE_DESC;
-typedef struct hipArrayMemoryRequirements hipArrayMemoryRequirements;
+typedef struct hipArrayMemoryRequirements {
+  size_t alignment;
+  size_t size;
+} hipArrayMemoryRequirements;
 typedef struct hipDeviceProp_tR0000 hipDeviceProp_tR0000;
 typedef hipDeviceProp_t hipDeviceProp_tR0600;
 typedef struct ihipDevResourceDesc_t* hipDevResourceDesc_t;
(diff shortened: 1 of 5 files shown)


libhrx/src/binding/hip/abi_stubs.c
```diff
@@ -207,50 +467,22 @@ static void hrx_hip_stream_callback_host_fn(void* user_data) {
...
+      return "hipMalloc";
+    case 4:
+      return "hipFree";
+    case 5:
+      return "hipMemcpy";
+    case 6:
+      return "hipLaunchKernel";
+    default:
+      return "unknown";
+  }
```
zjgarvey: Is this standard somewhere? I don't know if it is related to the other changes here either.

libhrx/src/binding/hip/abi_stubs.c
```diff
@@ -681,20 +905,10 @@ HIPAPI hipError_t hipGetDriverEntryPoint(
   return hipErrorNotSupported;
 }
 
-HIPAPI hipError_t hipGetFuncBySymbol(hipFunction_t* functionPtr,
```
zjgarvey, line 684: This is removed, but did you mean to reimplement this in api.c? I don't see it there.

libhrx/src/binding/hip/api.c
```diff
@@ -6852,8 +7259,12 @@ static hipError_t iree_hip_array_create(hipArray_t* array,
...
 
   new_array->magic = IREE_HIP_ARRAY_MAGIC;
+  iree_atomic_ref_count_init(&new_array->ref_count);
   new_array->desc = *desc;
+  new_array->public_extent = public_extent;
   new_array->extent = extent;
+  new_array->format = format;
+  new_array->num_channels = num_channels;
   new_array->flags = flags;
   new_array->context = context;
```
zjgarvey, line 7288: It looks like the arrays are retained below via a separate counter than the context (which is stored but isn't retained here). I'm not sure if the array/context decoupling is intended or not- I think hip/cuda tie arrays to their owning context.

libhrx/src/binding/hip/abi_stubs.c
```diff
@@ -1088,8 +1302,6 @@ HIPAPI hipError_t hipMallocMipmappedArray(
     hipMipmappedArray_t* mipmappedArray,
     const struct hipChannelFormatDesc* desc, struct hipExtent extent,
     unsigned int numLevels, unsigned int flags) {
-  (void)mipmappedArray;
```
zjgarvey, line 1091: Did you mean to implement this API? I guess I don't understand the removal of the void casts.

zjgarvey:
I forgot to leave a comment for the review:

In general, there is a lot of redundant scanning of the live arrays (I'm not sure how many of these we end up tracking in practice, so not sure if it's important to optimize). 

I also don't really get the mimapped array registry uses a different ownership model than the corresponding array registry. Is there a reason for this? Would it make sense to roll these together?

libhrx/src/binding/hip/abi_stubs.c
AWoloszyn: Removed, these got pulled in from elsewhere in the change.

libhrx/src/binding/hip/api.c
AWoloszyn: Agreed. There were two issues here: some array copy normalization paths looked up the same array more than once, and mipmapped arrays had a different registry/lifetime model than normal arrays. I updated
  the copy normalization path to reuse the retained array object for byte-range and extent checks, and changed mipmapped arrays to use a refcounted lookup/remove/release model like normal arrays.

AWoloszyn:
I pushed a cleanup pass addressing the review comments. The PR is now scoped back to HIP array allocation/descriptors/copy behavior: unrelated `hipGetFuncBySymbol`, `hipApiName`, peer-copy validation, and
  pitched-allocation refactor pieces were removed or restored to their previous unsupported behavior. I also tightened array and mipmapped-array ownership so handles retain the objects they depend on while
  in use.

zjgarvey (approved):

PR #155, opened 2026-06-29
Fix HIP graph launch and runtime edge semantics

AWoloszyn:
Allow no-argument native HIP kernel launches to be represented consistently across graph and direct stream paths when host stubs provide an empty args-array. Non-empty args-array launches without parameter metadata still fail explicitly instead of inventing an ABI layout.

Tighten graph memcpy and runtime pointer semantics around pooled logical allocation sizes, zero-byte allocation behavior, host-pointer attributes, worker-thread stream copies, event barriers, host-visible allocation backing, and native argument ABI offsets.

### DIFF
--- a/libhrx/src/binding/common/module.c
+++ b/libhrx/src/binding/common/module.c
@@ -255,6 +255,7 @@ static iree_status_t iree_hal_streaming_module_extract_metadata(
   for (iree_host_size_t i = 0, parameter_base = 0;
        iree_status_is_ok(status) && i < module->symbol_count; ++i) {
     iree_hal_streaming_symbol_t* symbol = &module->symbols[i];
+    memset(symbol, 0, sizeof(*symbol));
     symbol->module = module;
     symbol->name = export_infos[i].name;
     symbol->type = IREE_HAL_STREAMING_SYMBOL_TYPE_FUNCTION;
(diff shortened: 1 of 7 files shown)


libhrx/src/binding/common/stream.c
```diff
@@ -998,6 +1057,13 @@ static iree_status_t iree_hal_streaming_pack_raw_argument_list(
   return iree_ok_status();
 }
 
+static bool iree_hal_streaming_parameter_info_is_empty(
```
zjgarvey: Can this get lifted together with the duplicate in graph.c? Also, I think there may be other places, like the parameter unpack functions in this file, which could also be refactored to rely on the shared `is_empty` helper.

libhrx/src/binding/common/module.c
```diff
@@ -314,25 +337,28 @@ static iree_status_t iree_hal_streaming_module_extract_metadata(
...
         // backends, not byte offsets, so use the packed source offset we
         // calculate from the full parameter sequence. AMDGPU BUFFER_PTR
         // parameters already carry native kernarg byte offsets.
-        op->dst_offset =
-            is_buffer_binding_parameter ? parameter->offset : src_offset;
-        src_offset += parameter->size;
+        op->dst_offset = (uint16_t)native_dst_offset;
+        src_offset = (uint16_t)source_extent;
         buffer_size = src_offset;
+        ++src_ordinal;
```
zjgarvey: What was the point of adding this counter instead of using j?

zjgarvey (commented):
Looking at some other callers for `iree_hip_validate_memcpy_ranges`, I noticed that the `hipMemcpy` and `hipMemcpyAsync` pass `false` for `use_hip_visible_pool_size`, and wasn't sure if those direct stream paths should use this logical size bound, too.

AWoloszyn:
Thanks, these were valid cleanups. I added follow-up commit `6d437bceb`.

  - `hipMemcpy` and `hipMemcpyAsync` now use the HIP-visible logical allocation bound, matching the graph memcpy path. That matters for pooled allocations where the backing allocation can be larger than the
  allocation visible through HIP.
  - The duplicated parameter-info empty predicate is now shared. I used it in the pack/unpack paths, but kept the separate “requires parameter storage” check because direct-argument-only metadata is a valid
  state and should not require a packed/list parameter buffer.
  - `src_ordinal` was redundant; the metadata loop does not skip parameters, so it now uses `j` directly.

zjgarvey (approved):

PR #159, opened 2026-07-02
Fix topology storage max alignment

jtuyls:
iree_max_align_t is already the numeric maximum host alignment value, not a type. Passing iree_alignof(iree_max_align_t) asks for the alignment of the expression produced by that value, which is conceptually the wrong operation even on compilers that accept it.

Pass iree_max_align_t directly to iree_host_align so the topology storage layout is padded to the intended maximum host alignment.

On MSVC the old expression expands to __alignof(16) and fails with: topology_builder.c(66): error C2059: syntax error: 'constant'.


benvanik (approved):
This correction is the right representation. iree_max_align_t is already the numeric maximum host alignment, so passing it directly both preserves the intended 16-byte alignment and avoids forming invalid __alignof(16) under MSVC. The previous expression also evaluated to 8 on the reviewed Linux configuration, so this fixes a real under-alignment rather than only a compiler portability issue.

PR #160, opened 2026-07-03
Expose native ABI offsets from the HAL and use it in bindings.

AWoloszyn:
This is so that we can have a consistent approach to passing down bindings. Some times we have to pass down raw bindings, and some times we were extracting metadata/etc. Now at the streaming layer we just pass down native args all of the time. Because with native hip itself, we cannot trust that a pointer parameter is actually that pointer parameter (it could be shifted, encoded, or otherwise) we can't trust that we can actually introspect ANY parameters coming through the ABI. So instead of that we just always pack everything natively.


runtime/src/iree/hal/drivers/amdgpu/device/dispatch.c
```diff
@@ -72,20 +72,18 @@ void iree_hal_amdgpu_device_dispatch_emplace_custom_kernargs(
...
-                                  : explicit_bytes;
-    if (copy_bytes > 0) {
-      iree_amdgpu_memcpy(kernarg_ptr, custom_kernarg_ptr, copy_bytes);
-    }
+  const size_t explicit_bytes = layout->explicit_kernarg_size
+                                    ? layout->explicit_kernarg_size
+                                    : custom_kernarg_length;
+  const size_t copy_bytes = custom_kernarg_length < explicit_bytes
+                                ? custom_kernarg_length
+                                : explicit_bytes;
```
zjgarvey: Somewhere in here, I think we lost a bounding on `copy_bytes` to `total_kernarg_size`. I think the command_buffer path needs a clamp, too.

runtime/src/iree/hal/drivers/amdgpu/executable_metadata_hsaco_test.cc
```diff
@@ -261,6 +261,7 @@ TEST(ExecutableMetadataHsacoTest, PopulatesSparseInterleavedKernelLayout) {
   ExpectRebasedView(loaded_code_object_data, args[2].name,
                     metadata->parameters[2].name);
   EXPECT_EQ(metadata->parameters[2].offset, 2);
+  EXPECT_EQ(metadata->parameters[2].native_abi_offset, 20);
```
zjgarvey, line 268: I feel like the new path needs a bit of testing. Nothing in the current tests exercises the new packers, `iree_hip_parse_launch_extra`, `iree_hip_validate_launch_configuration`, or the new launch branches, and the one `emplace_custom_kernargs` test uses `layout = {}` so it never runs the changed branch.

If we are changing some of these kernarg read/writes, adding a test case with `data_length > total_kernarg_size` to avoid regressing on unbounded memcpy would be helpful.

runtime/src/iree/hal/drivers/amdgpu/device/dispatch.h
zjgarvey: I think you missed updating this stale comment, too.

libhrx/src/binding/common/internal.h
```diff
@@ -563,47 +563,44 @@ typedef enum iree_hal_streaming_symbol_type_e {
...
 } iree_hal_streaming_parameter_op_t;
 
-// Function parameter information used for unpacking.
+// Function parameter information used for argument packing.
 // Kernel launch parameters may arrive as a pointer array or packed argument
-// buffer and need to be converted into IREE constants and bindings for
-// dispatch. Bindless constant-only parameters are accepted by directly copying
-// buffer pointers to constant storage. Resolving buffer pointers to HAL
-// bindings is preferred because it gives the runtime explicit resource
-// ownership and is required by IREE async allocation support.
```
zjgarvey, line 593: Was the prior comment incorrect? Or do we lose async alloc support by using native device pointer values?

zjgarvey (commented):
I think more tests and checking out the unbounded caller controlled memcopy would be good before merging (the parallel PR).

libhrx/src/binding/common/internal.h
AWoloszyn: The old comment was describing the previous binding-resolution path, but it overclaimed the model after this change. For native HIP launches we cannot rely on reflected pointer slots as a complete residency/lifetime model because device pointers can be nested inside copied structs or otherwise appear in opaque argument bytes. This path preserves native device pointer values and uses metadata for ABI placement/validation; any conservative lifetime tracking has to live in the stream/runtime contract rather than pointer-slot introspection.

zjgarvey: That makes sense, thanks.

benvanik (changes requested):
Requesting changes. The native ABI offset is useful executable metadata, and
the cold-path reflection analysis is the right direction. I also verified that
ordinary Loom/HAL dispatch does not gain a new loop, allocation, lookup, or
branch from this patch: the changed AMDGPU hot code is reached only after
`IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS` has already selected the
libhrx/HIP-specific path.

The current implementation does not yet complete that abstraction boundary,
though, and it has two deterministic correctness failures: a zero-length
pre-packed argument span causes an out-of-bounds read, and a failed graph update
mutates the node it was supposed to leave unchanged. The replay record can also
accept a replacement executable with a different native layout, and another HAL
metadata producer leaves the new field indeterminate.

This review is intentionally strict about representation and failure behavior.
This code joins a public HIP ABI, graph capture/replay, executable reflection,
and the HAL dispatch path. In that part of the system, a partially migrated
representation tends to become silent argument corruption rather than a local
feature gap. A smaller explicit contract that rejects unsupported shapes is
safer than compatibility heuristics that make both representations look valid.

## What can remain as designed

The following parts do not need redesign based on this review:

- Adding an optional 16-bit native ABI offset to cold executable metadata is
  acceptable. A few bytes of metadata storage are not a dispatch-latency
  concern.
- Analyzing reflection once during libhrx module loading and caching native copy
  operations is the correct place for that work.
- `kernelParams` packing should copy the caller-provided argument values as
  bytes. In particular, it should preserve pointers nested inside by-value
  structs instead of trying to discover or translate them.
- The `extra` form should pass the caller's pre-packed native byte image through
  unchanged.
- Native packing remains explicitly opt-in through
  `IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS`. Loom plus ordinary HAL must
  continue to work without HIP and without native-ABI work on its dispatch path.
- The custom-direct copy clamps and canary tests correctly limit
  writes to the native argument bytes and required pre-implicit gap.
- The free/unregister ordering is a coherent allocation-lifetime contract:
  synchronous free synchronizes active contexts, asynchronous free is queued as
  a host action, and graph nodes privately copy argument bytes while referenced
  device allocations remain the application's responsibility.
- The installed HIP test-label and shared-library-name changes appear
  mechanically sound.

The requested changes below are about completing and proving that design, not
replacing it with pointer introspection or moving HIP policy into core HAL.

## 1. Finish separating the HAL parameter layout from the native ABI layout

`iree_hal_executable_function_parameter_t` now gives the two offsets distinct
meanings:

- `offset` is the ordinary HAL dispatch offset or binding ordinal;
- `native_abi_offset` is the target ABI byte offset and is meaningful only when
  its flag is present.

HSACO reflection does not follow that contract yet. A `global_buffer` is still
reported as `IREE_HAL_EXECUTABLE_FUNCTION_PARAMETER_BUFFER_PTR`, and its native
byte offset is written into both fields. libhrx then assumes that the first
`binding_count` `BUFFER_PTR` records are really HAL bindings.

That heuristic preserves the older HRX workaround instead of completing the
new representation. It also means a generic HAL reflection consumer sees a
normal AMDGPU binding as a raw pointer in the constants table.

The completed representation should be:

1. Reflect an HSACO `global_buffer` as
   `IREE_HAL_EXECUTABLE_FUNCTION_PARAMETER_BINDING`.
2. Store its dense HAL binding ordinal in `offset`.
3. Set the native-offset flag and store the HSACO byte offset in
   `native_abi_offset`.
4. Have libhrx classify the record from `BINDING`, then consult the optional
   native offset only while constructing the explicitly custom-direct HIP image.
5. Delete the `BUFFER_PTR && resolve_count < binding_count` heuristic and its
   compatibility-shaped fallback.

Please add metadata coverage that asserts both values independently, including
an export with padding between native parameters. Add a matching libhrx test
that proves the binding is ordinary HAL state in normal dispatch and lands at
the native byte offset only in custom-direct packing.

This is the main architecture requirement: HIP compatibility may consume extra
metadata, but it must not redefine the core HAL representation or impose work on
normal Loom/HAL dispatch.

## 2. Preserve the caller-declared length of a pre-packed argument span

`iree_hip_parse_launch_extra` accepts a non-null pre-packed buffer with a
declared size of zero. The direct and graph paths later replace that zero with
the reflected HAL `constant_bytes`. Graph add/set then copy bytes the caller
never declared, while direct launch passes the substituted length into HAL.

Under ASAN, a one-byte allocation declared as a zero-byte span with 16 reflected
constant bytes produces:

```text
AddressSanitizer: heap-buffer-overflow
READ of size 16
iree_hal_streaming_graph_set_kernel_node_params
  libhrx/src/binding/common/graph.c:1332
```

Zero is a real length, not a sentinel for “infer another representation's
size.” Keep the native span length exactly as supplied by the caller:

- a nonempty reflected kernel with a zero-length native span must fail before
  any read;
- an actually empty kernel may accept a zero-length span;
- a short span must reach the custom-direct validator unchanged and fail there;
  and
- omitted trailing ABI padding must remain legal when no parameter read reaches
  it.

Direct launch, graph add, and graph set should share one length contract instead
of independently substituting reflected sizes.

## 3. Make graph parameter replacement failure-atomic

`iree_hal_streaming_graph_set_kernel_node_params` clears the node's existing
storage and writes the replacement incrementally. If a later argument is null,
out of range, or otherwise invalid, the function returns an error after already
changing the node. `hipGraphKernelNodeSetParams` therefore corrupts a previously
valid graph node on failure.

A node initialized with eight `0xA5` bytes and updated with one valid 32-bit
argument followed by a null argument returns the expected error but leaves:

```text
44 33 22 11 00 00 00 00
```

instead of the original payload.

Every source pointer, source ordinal, destination range, capacity, mode, and
integer conversion must be validated before the first mutation. If complete
prevalidation is awkward, pack into temporary storage and commit only after the
entire replacement succeeds. The ordinary graph and graph-exec update paths
should call the same transactional implementation; maintaining a backup/restore
path for only one wrapper leaves the underlying contract inconsistent.

The regression test must assert the old bytes, bindings, and argument mode are
bit-for-bit unchanged after failures at the first, middle, and final operation.

## 4. Carry native layout identity through replay substitution validation

The replay executable-parameter record stores `type`, `flags`, `offset`, and an
8-bit `size`, but not `native_abi_offset`. Reloading the original executable
normally recovers the original native offset, so this does not manifest as
metadata loss during an ordinary replay. It does make substitution validation
incomplete: a replacement executable with identical HAL metadata and a
different native layout passes validation, after which custom-direct replay
writes arguments to the wrong native offsets.

Extend the replay schema and comparison to include `native_abi_offset` whenever
the flag is set. While changing that record, make `size` 16-bit to match the
public parameter representation; the current 8-bit field silently truncates
sizes the runtime API already permits.

Add a replay substitution test where the original and replacement differ only
in native offset. Validation must reject the replacement before any dispatch.
Also cover a parameter size above 255 so the serialization width remains
intentional.

## 5. Initialize every field from every HAL metadata producer

`iree_hal_executable_library_export_parameters` assigns the legacy output
fields individually and never initializes `native_abi_offset`. Its current v0
flags are zero, so this branch does not consume the indeterminate value today,
but the public record is still only partially initialized. A later flag
forwarding change or a consumer that copies/compares the complete record turns
that into observable undefined state.

Zero each output record before translating v0 metadata, then populate supported
fields. Add CTS assertions for flags and optional-field zero values rather than
testing only that the query returned OK. Every producer of a public aggregate
must establish the complete representation, including inactive optional fields.

## 6. Apply one launch-validation contract to every HIP entry point

The new shared configuration validator is used by
`hipModuleLaunchKernel` and the graph add/set APIs, but `hipLaunchKernel`—and
therefore `hipExtLaunchKernel`—bypasses it. That path also narrows
`size_t sharedMemBytes` into the 32-bit streaming representation without an
explicit range check.

All entry points need the same validation and HIP error mapping before capture
or dispatch. Backend validation should remain as a final invariant check, not as
a substitute for a consistent public API contract. Include at least:

- zero and unsupported workgroup dimensions;
- shared memory at `UINT32_MAX` and `UINT32_MAX + 1` where representable by the
  host API;
- identical failures from `hipLaunchKernel`, `hipExtLaunchKernel`,
  `hipModuleLaunchKernel`, graph add, and graph set; and
- proof that a validation failure does not allocate, enqueue, or mutate graph
  state.

## 7. Add production-shaped coverage for the changed behavior

The new unit test exercises `iree_hip_parse_launch_extra` and
`iree_hip_validate_launch_configuration` in isolation. The existing libhrx
binding suite does not execute an actual `hipLaunchKernel`,
`hipModuleLaunchKernel`, custom-direct HAL dispatch, graph kernel capture, graph
parameter replacement, or asynchronous pointer-lifetime scenario. That is why
all existing tests pass while the two deterministic reproducers above fail.

The minimum integrated matrix should include:

- an args-array launch with native inter-argument padding and a pointer nested
  inside a by-value struct;
- direct and graph pre-packed launches with exact, short, zero, and padded
  lengths;
- failed graph replacement preserving the prior node representation;
- repeated graph execution after the caller's temporary argument array and
  pre-packed source storage have gone out of scope;
- concurrent/repeated launches of multiple entry points with different native
  layouts; and
- ordinary HAL constants-plus-bindings dispatch of the same HSACO export,
  proving the native metadata addition does not alter its representation,
  dependency graph, or dispatch cost.

Tests should exercise the public HIP entry points and the real custom-direct HAL
path. Mocking the parser or directly calling only the packing helper cannot prove
the ownership, replay, or queue behavior at issue.

## Minimal reproducer for the two graph failures

The following can be saved as `/tmp/pr160_graph_argument_contract_probe.cc` and
run against this PR. It requires no GPU:

<details>
<summary>Probe source</summary>

```cpp
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <string_view>

#include "common/graph.h"
#include "common/internal.h"

static void InitializeKernelNode(
    iree_hal_streaming_graph_t* graph,
    iree_hal_streaming_graph_node_t* node,
    iree_hal_streaming_symbol_t* symbol, void* storage,
    iree_host_size_t storage_capacity) {
  *graph = {};
  *node = {};
  *symbol = {};
  node->graph = graph;
  node->type = IREE_HAL_STREAMING_GRAPH_NODE_TYPE_KERNEL;
  node->attrs.kernel.constants =
      iree_make_const_byte_span(storage, storage_capacity);
  node->attrs.kernel.constants_capacity = storage_capacity;
  symbol->type = IREE_HAL_STREAMING_SYMBOL_TYPE_FUNCTION;
}

static int ProbeZeroLength() {
  iree_hal_streaming_graph_t graph;
  iree_hal_streaming_graph_node_t node;
  iree_hal_streaming_symbol_t symbol;
  uint8_t destination[16] = {};
  InitializeKernelNode(&graph, &node, &symbol, destination,
                       sizeof(destination));
  symbol.parameters.constant_bytes = sizeof(destination);

  uint8_t* source = static_cast<uint8_t*>(std::malloc(1));
  source[0] = 0x5A;
  iree_hal_streaming_dispatch_params_t params = {};
  params.buffer = source;
  params.buffer_size = 0;
  params.flags = IREE_HAL_STREAMING_DISPATCH_FLAG_PRE_PACKED;
  iree_status_t status = iree_hal_streaming_graph_set_kernel_node_params(
      &node, &symbol, &params);
  iree_status_ignore(status);
  std::free(source);
  return 0;
}

static int ProbeAtomicity() {
  iree_hal_streaming_graph_t graph;
  iree_hal_streaming_graph_node_t node;
  iree_hal_streaming_symbol_t symbol;
  uint8_t destination[8];
  std::memset(destination, 0xA5, sizeof(destination));
  InitializeKernelNode(&graph, &node, &symbol, destination,
                       sizeof(destination));

  iree_hal_streaming_parameter_op_t ops[2] = {};
  ops[0].copy.size = 4;
  ops[0].copy.native_abi_destination_offset = 0;
  ops[0].copy.source_ordinal = 0;
  ops[1].copy.size = 4;
  ops[1].copy.native_abi_destination_offset = 4;
  ops[1].copy.source_ordinal = 1;
  symbol.parameters.buffer_size = sizeof(destination);
  symbol.parameters.direct_arg_bytes = sizeof(destination);
  symbol.parameters.copy_count = 2;
  symbol.parameters.ops = ops;

  uint32_t first_argument = 0x11223344u;
  void* arguments[2] = {&first_argument, nullptr};
  iree_hal_streaming_dispatch_params_t params = {};
  params.buffer = arguments;
  params.flags = IREE_HAL_STREAMING_DISPATCH_FLAG_ARGS_ARRAY;
  iree_status_t status = iree_hal_streaming_graph_set_kernel_node_params(
      &node, &symbol, &params);
  if (iree_status_is_ok(status)) return 2;
  iree_status_ignore(status);

  uint8_t expected[sizeof(destination)];
  std::memset(expected, 0xA5, sizeof(expected));
  if (std::memcmp(destination, expected, sizeof(destination)) != 0) {
    std::fprintf(stderr, "failed update mutated captured bytes:");
    for (uint8_t byte : destination) std::fprintf(stderr, " %02x", byte);
    std::fprintf(stderr, "\n");
    return 3;
  }
  return 0;
}

int main(int argc, char** argv) {
  if (argc != 2) return 1;
  if (std::string_view(argv[1]) == "zero-length") return ProbeZeroLength();
  if (std::string_view(argv[1]) == "atomicity") return ProbeAtomicity();
  return 1;
}
```

</details>

```bash
build_tools/bin/iree-bazel-try \
  /tmp/pr160_graph_argument_contract_probe.cc \
  --config=asan \
  --dep //libhrx/src/binding/common:common -- zero-length

build_tools/bin/iree-bazel-try \
  /tmp/pr160_graph_argument_contract_probe.cc \
  --config=asan \
  --dep //libhrx/src/binding/common:common -- atomicity
```

The first command currently terminates in the 16-byte ASAN read. The second
returns 3 and prints the mutated node bytes.

## Required verification before the next revision

Please install the repository's paranoid hook once and run the full profile
before pushing:

```bash
```

At minimum, the revised branch should also pass the focused sanitizer surface:

```bash
  //runtime/src/iree/hal/drivers/amdgpu/device:dispatch_test \
  //runtime/src/iree/hal/drivers/amdgpu:executable_metadata_hsaco_test \
  //libhrx/src/binding/hip:launch_params_test \
  //libhrx/src/binding/...
```

The two reproducers above must become checked-in regression tests rather than
remaining standalone probes. Please also add the integrated launch, graph,
replay-substitution, and normal-HAL coverage described above.

For reference, the reviewed branch already passes the three focused tests and
all six existing libhrx binding tests under ASAN. That evidence is useful but
does not exercise the failing behaviors.

Once these changes are in place, the main design should be considerably easier
to review: ordinary HAL parameters will have one honest representation, native
packing will remain an explicit HIP/libhrx strategy, graph replacement will be
transactional, and replay will validate the same ABI identity that dispatch
uses.

AWoloszyn:
That should now cover the comments.

benvanik (approved):
Approved after the latest repair pass.

The blocking representation and failure-path contracts from my earlier review are now in place:

- Ordinary HAL parameters retain their normal binding/constant representation, while the target ABI byte offset is separate optional metadata consumed only by the custom-direct libhrx/HIP path.
- Caller-declared prepacked spans keep their actual length, including zero and short spans, and are rejected before an out-of-bounds read.
- Graph parameter replacement validates before mutation and therefore preserves the prior node on failure.
- Replay records carry and compare the native layout identity and the full parameter size.
- Public HIP entry points share the launch-configuration checks and reject narrowing or resource-limit failures before capture or submission.
- The added native argument work remains behind `IREE_HAL_DISPATCH_FLAG_CUSTOM_DIRECT_ARGUMENTS`; normal Loom/HAL dispatch does not acquire another loop, allocation, lookup, or branch.

I pushed three narrowly scoped finishing commits onto this branch. They keep the Linux-only HIP shim artifact and validation test out of Windows CMake, construct non-copyable and variable-sized test objects in a form accepted by native MSVC, replace a transient edit-history comment with the public size contract, and remove one remaining C++20 designated initializer from the AMDGPU hardware test. The complete branch passes the paranoid presubmit profile, including runtime/libhrx/Loom tests, Semgrep, and clang-tidy.

I also exercised the production-shaped public HIP path locally with an actual custom-direct kernel: args-array packing with a pointer nested in a by-value struct, padded prepacked arguments, graph capture followed by overwriting caller-owned argument storage, repeated graph launches, and output readback. That passed both the local gfx1100 fixture and the multi-target embedded fixture selection under ASAN.

The cache/reload and executable-image lifetime proof belongs with #173, which replaces cache-shaped loading with a target-explicit API. After this lands, the public HIP regression should be adapted to #173's target-aware fixture API and checked in there so it covers the final loading contract instead of immediately being rewritten around that API change.

The remaining native-MSVC issue is older than this branch: the HIP API header spells `nodiscard` with a GNU annotation, and the existing Linux-only shim hides that from MSVC today. I verified a small portable spelling plus removal of the warning suppressions locally against all seven registered libhrx tests. Per the agreed merge boundary, that cleanup will be submitted as a separate follow-on from the merged `main`; it is not a reason to rebase or hold #160.

PR #161, opened 2026-07-06
[CI] Add Windows CMake package flow

jtuyls:
Adds a Windows CMake product and packaging flow for the core runtime. The
MSVC lane is the canonical Windows product build: it configures, builds, and
installs the tree, runs the installed tests, creates the release archives,
relocates the extracted package, reruns its installed tests, and publishes the
artifacts. A separate clang-cl lane exercises the same build and installed-test
surface as a second-frontend diagnostic without producing release packages.

Moves cross-platform artifact acquisition, phase reporting, command
construction, and packaging metadata into `ci_core_common.py`. The Linux and
Windows drivers retain their platform-specific environment and toolchain
policy, with tests covering the shared contract and each driver's generated
commands.

Makes the libhrx dependency boundary explicit. AMDGPU no longer implicitly
enables libhrx, and `LIBHRX_BUILD_HIP_BINDING` independently owns the HIP
binding. The Windows core artifact can therefore build and test libhrx without
acquiring HIP- or AMDGPU-only dependencies; Loom, AMDGPU, CTS, passthrough, the
HIP binding, and benchmarks remain outside this core Windows product surface.

Completes the runtime portability needed by both compiler lanes. Installed
Python tests use the configured interpreter, stdio descriptor wrapping maps to
the native MSVC API, and timestamp scaling uses checked 64-by-64-bit arithmetic
without relying on compiler-specific 128-bit integer syntax.

## Reviewer notes

The highest-value review surfaces are the MSVC package relocation contract,
the MSVC-versus-clang-cl policy split, the shared/common CI boundary, and the
libhrx/HIP option ownership.


benvanik (changes requested):
Requesting changes on two product-contract issues: the only Windows lane uses
ROCm `clang-cl` rather than MSVC, and the packaged CTest tree embeds the build
worker's absolute Python path. The common CI factoring, portable timestamp
math, and libhrx/HIP build decoupling are all reasonable and do not need to be
reworked based on this review.

The distinction between the two compilers is important here. MSVC is the
baseline Windows product compiler; clang-cl is a valuable supplemental compiler
because it gives us better sanitizer and diagnostic coverage. Entering a Visual
Studio environment does not make a clang-cl build an MSVC build. Likewise, a
package is relocatable only if its tests and consumers run after extraction on
a different machine/root—not merely because the original install tree passed.

The bar is deliberately high because this workflow will become the evidence
for Windows support and will publish the artifacts users consume. A green job
must name and prove the exact compiler and package contract it represents.

## What can remain as designed

The following parts have supporting evidence and should not be churned while
addressing the two blockers:

- `iree_math_round_mul_div_u64` preserves the full 128-bit intermediate,
  performs the intended half-up rounding, and diagnoses zero denominators and
  quotient overflow. Its focused MSVC test passes, and the portable path matched
  a native `unsigned __int128` oracle for two million deterministic random
  inputs.
- Consolidating the profile timestamp conversions onto that helper is the right
  factorization.
- Using `iree_time_now()` for the already environment-gated HRX timing path is
  the correct Windows portability change and does not add work to the normal
  launch path.
- `LIBHRX_BUILD` no longer implicitly forces AMDGPU/HIP, and
  `LIBHRX_BUILD_HIP_BINDING` is explicit. That is the dependency shape we want:
  libhrx core plus HAL must build without HIP, and the HIP compatibility binding
  remains opt-in.
- The common/Linux/Windows CI driver factoring is large, but the reviewed unit
  tests pass under ASAN. The workflow uses read-only permissions, pinned
  actions, argument-vector subprocesses, and checked extraction destinations;
  no privilege or path-traversal blocker was found.

## 1. Add an explicit mandatory MSVC product lane

The reusable workflow currently defines one matrix entry named `Default`.
`build_tools/ci_core_windows.py` then selects ROCm's `clang-cl.exe`,
`clang++.exe`, `llvm-lib.exe`, and `lld-link.exe` unless every tool path is
overridden. `msvc_build_env()` imports the Visual Studio and Windows SDK
environment, but it does not select `cl.exe`.

The successful live job therefore proves clang-cl compatibility and package
orchestration. It does not establish that the repository builds with MSVC, even
though the job and artifacts are labeled only as Windows.

Please make the toolchain a coherent named input, with exactly supported values
such as `msvc` and `clang-cl`. It should resolve a tuple rather than exposing
four independent path overrides:

- `msvc`: `cl.exe` for C/C++ and C-preprocessed assembly, plus `lib.exe`,
  `link.exe`, and `rc.exe` from the initialized developer environment;
- `clang-cl`: ROCm `clang-cl.exe`/`clang++.exe`, `llvm-lib.exe`, and
  `lld-link.exe`, while retaining the Windows SDK resource compiler.

The minimum useful matrix is:

| Lane | Contract |
| --- | --- |
| `windows-msvc` | Required product lane. Build the narrow no-HIP base configuration with real `cl.exe`; run build-tree and installed tests; build consumer smoke projects; create the canonical Windows package; scan it; relocate it; and rerun the installed tests/consumers. |
| `windows-clang-cl` | Supplemental diagnostic lane. Keep the broader assertions/sanitizer/fanout that ROCm LLVM supports. It does not need to duplicate every packaging step unless we deliberately ship a distinct clang-cl artifact. |

One MSVC build/test/package lane is sufficient as the baseline; the larger
fanout can remain on clang-cl. The MSVC support itself is not optional.

Job names, required-check names, build directories, cache keys, and artifact
identities must include the compiler. A green clang-cl job must not be able to
satisfy the MSVC required check, and two matrix entries must not race to upload
the same immutable artifact name. If only one lane publishes the canonical
Windows package, it should be the MSVC lane.

The MSVC CMake configuration needs `CMAKE_ASM_COMPILER=cl.exe`; this repository
contains C-preprocessed assembly as well as MASM, and leaving that selection
implicit chooses an unsuitable assembler in this configuration.

## 2. Resolve Python when the installed test tree runs

The PR changes an installed test command from portable `python3` to the
configure-time absolute `${Python3_EXECUTABLE}`. That fixes command discovery on
the build worker but serializes the worker's filesystem into the package.

The successful artifact from Windows run `28864732869` contains this command at
lines 361 and 363 of
`share/hrx-system/tests/ctest/CTestTestfile.cmake`:

```cmake
add_test(... "C:/Python313/python.exe" ...)
```

This can be reproduced directly from the live artifact:

```bash
mkdir -p /tmp/pr161-windows-artifact
gh run download 28864732869 \
  --repo ROCm/hrx-system \
  --name hrx-tests-windows-x86_64 \
  --dir /tmp/pr161-windows-artifact

unzip -p \
  /tmp/pr161-windows-artifact/hrx-tests-windows-x86_64-2026-07-07.zip \
  share/hrx-system/tests/ctest/CTestTestfile.cmake \
  | rg -n 'python(3|\.exe)|Python[0-9]+'
```

A user does not necessarily have Python at the build worker's path, so the
package is not relocatable as produced.

Resolve the interpreter at installed-test runtime. One suitable generated CTest
contract is:

```cmake
if(DEFINED ENV{HRX_TEST_PYTHON} AND
   NOT "$ENV{HRX_TEST_PYTHON}" STREQUAL "")
  set(HRX_TEST_PYTHON "$ENV{HRX_TEST_PYTHON}")
else()
  find_program(HRX_TEST_PYTHON NAMES python3 python REQUIRED)
endif()
```

All installed Python tests should invoke `${HRX_TEST_PYTHON}`. The environment
override gives hermetic CI and users an explicit interpreter selection; the
`find_program` fallback keeps an ordinary unpacked package usable.

The package lane needs both of these checks:

1. Scan every generated CMake/CTest file and reject absolute source-root,
   build-root, install-root, and configure-time interpreter paths.
2. Extract the package into a different root, make the original build and
   install trees unavailable, select Python only through `HRX_TEST_PYTHON` or
   the installed lookup contract, and run the installed tests and consumer
   smoke projects from that relocated tree.

Running CTest from the original install directory before zipping remains useful,
but it cannot prove relocation because all of the accidentally captured build
state is still present.

## Mainline MSVC repairs we will land separately

A real MSVC 19.44 build exposed several current-`main` defects. These are not
introduced by this PR and should not be folded into it. We will pre-land them as
a focused MSVC-enablement series, after which this branch can rebase and add the
actual CI lane without carrying unrelated fixes.

For transparency, the required series is:

1. Land PR 159: pass numeric `iree_max_align_t` directly rather than forming
   `iree_alignof(iree_max_align_t)`/`__alignof(16)`.
2. Remove the unnecessary raw `-Wno-unused-variable` from
   `libhrx/src/binding/common/BUILD.bazel` and regenerate CMake. MSVC interprets
   it as the invalid `/Wno-unused-variable` option.
3. Replace the direct `<stdatomic.h>` global in
   `libhrx/src/binding/common/memory.c` with IREE atomics
   (`iree_atomic_uint64_t`, `IREE_ATOMIC_VAR_INIT`, and relaxed
   `iree_atomic_fetch_add`). Do not impose `/experimental:c11atomics` on the
   product build.
4. Capture `messages_per_producer` by value in all four producer lambdas in
   `runtime/src/iree/base/internal/mpsc_queue_test.cc`.
5. Include `<numeric>` directly in
   `libhrx/src/binding/common/buffer_table_benchmark.cc` before using
   `std::iota`.
6. Make build-tree libhrx CTest provide the directory containing `hrx.dll` to
   `buffer_table_test.exe` and `buffer_table_benchmark.exe`, or place the DLL
   next to the executables. Both currently fail at loader startup with Windows
   status `0xc0000135` and pass when the DLL directory is on `PATH`.
7. Regenerate or replace
   `runtime/src/iree/hal/local/elf/testdata/elementwise_mul_x86_64.so`. The
   checked-in fixture implements local executable-library ABI v0.6 while main
   now requests v0.7.

With diagnostic versions of items 1–5, a completely clean Release build
completed all 1,361 compile/link steps under MSVC 19.44. CTest initially passed
234 of 237 tests. Supplying the DLL path made both libhrx tests pass; a complete
rerun with that environment and only the proven-stale ELF fixture excluded
passed all 236 selected tests. This demonstrates that a narrow real-MSVC lane is
practical once the seven baseline repairs land.

The product-shaped reproduction after those repairs, from an x64 Visual Studio
developer PowerShell, is:

```powershell
cmake -S . -B build-msvc -G Ninja `
  -DCMAKE_BUILD_TYPE=Release `
  -DCMAKE_C_COMPILER=cl.exe `
  -DCMAKE_CXX_COMPILER=cl.exe `
  -DCMAKE_ASM_COMPILER=cl.exe `
  -DIREE_BUILD_TESTS=ON `
  -DLOOM_BUILD=OFF `
  -DLIBHRX_BUILD=ON `
  -DLIBHRX_BUILD_CTS=OFF `
  -DLIBHRX_BUILD_HIP_BINDING=OFF `
  -DLIBHRX_BUILD_PASSTHROUGH=OFF `
  -DIREE_HAL_DRIVER_AMDGPU=OFF

cmake --build build-msvc --parallel 8

ctest --test-dir build-msvc --output-on-failure -j 8
```

The CI lane should encode the same no-HIP dependency shape and select and report
the toolchain directly.

## Required package and workflow tests

The revised driver tests should prove the named contract rather than only the
generated command list:

- `msvc` and `clang-cl` resolve distinct coherent compiler/linker/library
  tuples;
- unknown toolchain names fail before configuration;
- check names, build roots, caches, and artifacts contain the compiler name;
- the mandatory MSVC entry cannot be removed while a clang-cl entry remains;
- only the intended lane publishes the canonical Windows package;
- generated installed-test files contain no original absolute roots or
  configure-time interpreter path;
- the relocated package runs with `HRX_TEST_PYTHON` pointing to a different
  interpreter location; and
- consumer smoke projects discover only the extracted package, not the build
  tree.

Please retain the existing archive traversal tests and argument-vector command
construction. Those parts already model the right security boundary.

## Required presubmit before the next revision

Install the repository's paranoid hook once and run the complete profile before
pushing:

```bash
```

The focused CI orchestration tests should also remain green under ASAN:

```bash
  //build_tools:ci_core_linux_test \
  //build_tools:ci_core_windows_test \
  //build_tools/devtools:ci_test
```

Once the separate mainline repairs land, the remaining work in this PR is
conceptually narrow: name the compiler contract honestly, add one required MSVC
base lane while keeping the broader clang-cl lane, and make the installed test
package resolve its dependencies after relocation. The portable runtime and
dependency-decoupling changes can remain intact.

benvanik:
Thanks — I reviewed the two commits pushed after the change request. The two substantive blockers appear to be resolved.

The workflow now has a real native-MSVC product lane and a distinct clang-cl diagnostic lane. The MSVC lane selects the coherent `cl.exe`/`lib.exe`/`link.exe` toolchain tuple, uses `cl.exe` for C-preprocessed assembly, is the only lane that publishes the canonical package, and passes alongside the broader clang-cl lane.

The installed-test package now resolves Python when CTest runs through `HRX_TEST_PYTHON`, with a `find_program(python3 python)` fallback. Package creation scans generated CMake/CTest files for captured source, build, install, and interpreter paths. The relocation test then moves the original build and install trees out of the way, creates a new Python environment under the relocated root, extracts the packages into new directories, and reruns the installed tests and consumer projects. That is the relocation evidence requested in the review rather than another test of the original install tree.

The remaining blocker is branch ancestry, not the implementation of those two contracts. Commit `2f70286f` is a squashed copy of the then-unmerged #172 baseline. #172 is now on `main` with its real history, so this branch contains a second representation of that baseline and GitHub reports it as conflicting. Please rebase onto current `main` while dropping `2f70286f`, retaining the packaging/workflow change from `a7aa7528`. For example:

```bash
git fetch origin main
git rebase --onto origin/main 2f70286f
```

Please preserve the current mainline no-HIP core dependency shape during conflict resolution. Once the rebased Windows MSVC, clang-cl, relocation, presubmit, and focused orchestration checks are green, I will do one final review of the true delta. Assuming the rebase does not distort the implementation, the requested changes are satisfied and I expect to approve it.

benvanik (approved):
The two requested product contracts are now satisfied on current main. MSVC owns the canonical build, installed-test, package, relocation, and retest lane; clang-cl remains a separate diagnostic lane. Installed tests resolve Python at execution time, and the relocated package passes without the original build or install roots. The shared CI factoring and explicit libhrx/HIP dependency boundary remain coherent.

PR #163, opened 2026-07-07
[hip] Fix hipGetProcAddress symbol resolution under RTLD_LOCAL

### LINKED ISSUE #162: hipGetProcAddress resolves against process-global scope; fails when libamdhip64 is dlopen'd RTLD_LOCAL (breaks Triton)
## Summary

HRX's `hipGetProcAddress()` resolves symbols against the **process-global**
symbol scope (`dlsym(dlopen(NULL), symbol)`) instead of against the HIP runtime
library itself. Any consumer that `dlopen`s `libamdhip64.so` with `RTLD_LOCAL`
therefore cannot resolve **any** HIP entry point through `hipGetProcAddress()`,
because an `RTLD_LOCAL` library never exports its symbols into the global scope.

This breaks **Triton** (and any similar `hipGetProcAddress`-based loader). Triton's
AMD backend does exactly this: `dlopen(libamdhip64.so, RTLD_LAZY | RTLD_LOCAL)`
then resolves the whole HIP API via `hipGetProcAddress`. The first required
symbol (`hipGetLastError`) fails, so Triton dies at driver-init with:

```
RuntimeError: cannot get address for 'hipGetLastError' from libamdhip64.so
```

Stock ROCm resolves against its own module, so it works either way.

## Environment

- Branch: `users/awoloszyn/native-bindings` (PR #160), tip `4dc14be63`.
- MI300X (`gfx942`), ROCm 7.2.1, torch `2.14.0.dev+rocm7.2`, triton `3.8.0`.
- Pure-HIP launch probes (static / module / graph_add / graph_setparams /
  graph_execsetparams) all pass under HRX; only the `hipGetProcAddress` path is
  affected.

## Root cause

`libhrx/src/binding/hip/api.c`, `hipGetProcAddress()`:

```c
void* process = dlopen(NULL, RTLD_LAZY);   // process-GLOBAL scope
...
*pfn = dlsym(process, symbol);             // misses RTLD_LOCAL libraries
HIP_RETURN_ERROR(*pfn ? hipSuccess : hipErrorNotFound);
```

Triton's loader (`triton/backends/amd/driver.c`):

```c
dlopen(hipLibSearchPaths[i], RTLD_LAZY | RTLD_LOCAL);   // RTLD_LOCAL
...
hipGetProcAddress("hipGetLastError", &fn, hipVersion, 0, &status);  // -> hipErrorNotFound
```

## Minimal reproducer (no GPU needed)

```c
// cc repro.c -o repro -ldl ; ./repro /path/to/libamdhip64.so
#include <dlfcn.h>
#include <stdint.h>
#include <stdio.h>
typedef int (*gpa_t)(const char*, void**, int, uint64_t, void*);
static int go(const char* p, int flag, const char* label) {
  void* lib = dlopen(p, RTLD_LAZY | flag);
  gpa_t gpa = (gpa_t)dlsym(lib, "hipGetProcAddress");
  void* pfn = 0; int st = -1;
  int rc = gpa("hipGetLastError", &pfn, 60000000, 0, &st);
  printf("[%s] rc=%d pfn=%p -> %s\n", label, rc, pfn, (rc==0&&pfn)?"OK":"FAIL");
  dlclose(lib); return (rc==0&&pfn)?0:1;
}
int main(int c, char** v){ const char* p=c>1?v[1]:"libamdhip64.so";
  go(p, RTLD_LOCAL, "RTLD_LOCAL"); go(p, RTLD_GLOBAL, "RTLD_GLOBAL"); return 0; }
```

Buggy HRX output:

```
[RTLD_LOCAL ] rc=500 pfn=(nil) -> FAIL      # 500 = hipErrorNotFound
[RTLD_GLOBAL] rc=0   pfn=0x... -> OK
```

Stock ROCm: both `OK`.

## Fix

Resolve against **this** module. Use `dladdr()` on an owned symbol to find our
own path, then reopen it `RTLD_NOLOAD` to get a handle scoped to the HIP runtime
library; fall back to the global scope only if that fails. (Requires
`_GNU_SOURCE` for `dladdr`/`RTLD_NOLOAD`.)

After the fix the reproducer prints `OK` for both `RTLD_LOCAL` and
`RTLD_GLOBAL`, and both Triton workloads (`triton_add.py`, minimal
`add.py`) run through HRX with `VERDICT=FULLY_WORKS`.

## Test

Added `libhrx/src/binding/hip/hip_get_proc_address_test.cc`
(`hrx_cc_test`): `dlopen(..., RTLD_LOCAL)` then assert
`hipGetProcAddress("hipGetLastError", ...)` returns `hipSuccess` with a
non-null pointer. Verified it **fails** on the pre-fix library and **passes**
on the fixed one.

amd-vivekag:
## Summary

`hipGetProcAddress()` resolved symbols against the **process-global** symbol
scope (`dlsym(dlopen(NULL), symbol)`) instead of against the HIP runtime library
that defines them. Any consumer that `dlopen`s `libamdhip64.so` with
`RTLD_LOCAL` therefore cannot resolve **any** HIP entry point through
`hipGetProcAddress()` — an `RTLD_LOCAL` library never exports its symbols into
the global scope.

This breaks **Triton**: its AMD backend does
`dlopen(libamdhip64.so, RTLD_LAZY | RTLD_LOCAL)` and then binds the whole HIP API
via `hipGetProcAddress`. The first required symbol (`hipGetLastError`) fails, so
Triton dies at driver-init with:

```
RuntimeError: cannot get address for 'hipGetLastError' from libamdhip64.so
```

This is a pre-existing bug on `main` (independent of any in-flight work).

Fixes #162.

## Fix

Resolve against **this** module: `dladdr()` on an owned symbol yields our own
path, which is reopened once with `RTLD_NOLOAD` to get a handle scoped to the
HIP runtime library (resolved lazily via `iree_call_once`). Falls back to the
global scope only if the self-handle cannot be established. `_GNU_SOURCE` is
defined for `dladdr`/`Dl_info`/`RTLD_NOLOAD`.

## Test

Adds `hip_get_proc_address_test` (`hrx_cc_test`): `dlopen(..., RTLD_LOCAL)` then
asserts `hipGetProcAddress("hipGetLastError", ...)` returns `hipSuccess` with a
non-null pointer, reproducing the consumer (Triton) scenario. It **fails** on
the old implementation and **passes** with the fix. The generated
`CMakeLists.txt` was regenerated with `bazel_to_cmake` and is in sync with
`BUILD.bazel`.

## Verification

- GPU-free C reproducer: `RTLD_LOCAL` goes `FAIL` (hipErrorNotFound) → `OK`
  after the fix; `RTLD_GLOBAL` stays `OK`.
- New gtest passes.
- Both Triton workloads (`triton_add.py`, minimal `add.py`) go from failing at
  `hipGetProcAddress` to `VERDICT=FULLY_WORKS` end-to-end through HRX.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -52,10 +52,17 @@ hrx_cc_shared_library(
 hrx_cc_test(
     name = "hip_get_proc_address_test",
     srcs = ["hip_get_proc_address_test.cc"],
-    # The test dlopen()s the built shared object with RTLD_LOCAL to reproduce
-    # the consumer scenario (e.g. Triton), so it needs the artifact at runtime.
-    data = [":amdhip64"],
+    # The test dlopen()s the built libamdhip64 with RTLD_LOCAL to reproduce the
+    # consumer scenario (e.g. Triton). Under CMake the exact artifact path is
+    # injected via the HRX_TEST_LIBAMDHIP64_PATH compile definition and its
+    # build dependency (see CMakeLists.txt below the preserve marker). Under
+    # Bazel, point the test at the artifact with HRX_TEST_LIBAMDHIP64, e.g.
+    # `bazel test --test_env=HRX_TEST_LIBAMDHIP64=$(bazel cquery ...)`; without
+    # it the test skips (GTEST_SKIP) rather than failing.
     deps = [
+        # The test itself calls dlopen()/dlsym()/dlclose(); link libdl so the
+        # executable resolves them (glibc that ships dl separately, ld.lld).
+        "//build_tools:dl",
         "//runtime/src/iree/testing:gtest",
         "//runtime/src/iree/testing:gtest_main",
     ],
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -7,6 +7,7 @@
 load(
     "//libhrx/build_tools/bazel:cc.bzl",
     "hrx_cc_shared_library",
+    "hrx_cc_test",
 )
 
 package(
@@ -42,5 +43,17 @@ hrx_cc_shared_library(
         "//runtime/src/iree/base",
         "//runtime/src/iree/base/threading",
         "//runtime/src/iree/hal",
+    ],
+)
+
+hrx_cc_test(
+    name = "hip_get_proc_address_test",
+    srcs = ["hip_get_proc_address_test.cc"],
+    # The test dlopen()s the built shared object with RTLD_LOCAL to reproduce
+    # the consumer scenario (e.g. Triton), so it needs the artifact at runtime.
+    data = [":amdhip64"],
+    deps = [
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
     ],
 )
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -22,7 +22,10 @@ hrx_cc_shared_library(
         "abi_stubs.c",
         "api.c",
     ],
-    hdrs = ["api.h"],
+    hdrs = [
+        "api.h",
+        "binding_internal.h",
+    ],
     additional_linker_inputs = ["amdhip64.map"],
     copts = [
         "-DIREE_HAL_STREAMING_HIP_EXPORTS",
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -18,6 +18,7 @@ iree_cc_library(
     "-Wno-unused-variable"
   HDRS
     "api.h"
+    "binding_internal.h"
   SRCS
     "abi_stubs.c"
     "api.c"
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -41,7 +41,30 @@ iree_cc_library(
   PUBLIC
 )
 
+iree_cc_test(
+  NAME
+    hip_get_proc_address_test
+  SRCS
+    "hip_get_proc_address_test.cc"
+  DATA
+    ::amdhip64
+  DEPS
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+)
+
 ### BAZEL_TO_CMAKE_PRESERVES_ALL_CONTENT_BELOW_THIS_LINE ###
+
+# Point hip_get_proc_address_test at the shared object built in this tree so it
+# can dlopen() the exact artifact under test. Guarded because the test target
+# only exists when tests are enabled.
+if(TARGET libhrx_src_binding_hip_hip_get_proc_address_test)
+  target_compile_definitions(libhrx_src_binding_hip_hip_get_proc_address_test
+    PRIVATE
+      "HRX_TEST_LIBAMDHIP64_PATH=\"$<TARGET_FILE:libhrx_src_binding_hip_amdhip64>\""
+  )
+endif()
 
 # The symbol version script and --undefined-version are applied through the
 # generated LINKOPTS above (from linkopts in BUILD.bazel). Keep them out of
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -47,9 +47,8 @@ iree_cc_test(
     hip_get_proc_address_test
   SRCS
     "hip_get_proc_address_test.cc"
-  DATA
-    ::amdhip64
   DEPS
+    ${CMAKE_DL_LIBS}
     iree::testing::gtest
     iree::testing::gtest_main
     libhrx_defs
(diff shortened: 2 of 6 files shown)


amd-vivekag:
## Design notes: options considered, in plain words

### What's broken

`hipGetProcAddress("hipMalloc")` is supposed to hand back a pointer to *our*
`hipMalloc`. The old code found it by asking the **global** symbol table of the
whole process (`dlsym(dlopen(NULL), name)`).

That only works if our library's symbols are *in* the global table. When a
program loads us with `RTLD_LOCAL` (Triton's AMD backend does exactly that), our
symbols are kept **private** to that load and never go global — so the lookup
finds nothing and returns `hipErrorNotFound`. Triton asks for `hipGetLastError`
first, so it dies immediately at startup.

### The options I weighed

**A. Look ourselves up directly (what this PR does).**
Ask the loader "which file am I in?" (`dladdr`), then get a handle to *that*
file (`dlopen(..., RTLD_NOLOAD)`) and search it. This looks in our own library
every time, so it doesn't matter whether we were loaded global or local.
- Why it makes sense: tiny change; behaves like real ROCm (returns pointers to
  *this* implementation); reuses the symbols we already publish, so there's no
  second list to keep in sync; and it's the safe choice when a *different*
  `libamdhip64` (e.g. the one PyTorch bundles) is also in memory — we never
  accidentally hand out the other library's function.
- Downside: uses Linux/glibc-only helpers (`dladdr`, `RTLD_NOLOAD`), and it can
  only return symbols we actually export. Both are fine here (the shim is
  already Linux-only and dlopen-based), but noted.

**B. Use a built-in "search flag" like `RTLD_DEFAULT` / `RTLD_SELF`.**
Tempting one-liner, but it doesn't actually fix the problem: `RTLD_DEFAULT` is
just the same global table (still empty for us), and `RTLD_SELF` starts
searching *after* our own library, so it skips exactly what we want. Rejected —
doesn't work.

**C. Keep a hand-built table: name -> function pointer.**
Store `{"hipMalloc", &hipMalloc}, ...` for all ~540 entry points and look up in
that. This is basically how upstream AMD's real HIP does it.
- Why it's attractive: no dependence on the dynamic loader at all; works under
  any load mode and even static linking; and it can precisely honor the
  `flags`/stream-per-thread request.
- Why not now: it's a big (~540-entry) generated table to add and keep in sync,
  i.e. a much larger change for the same user-visible result. Better as a
  possible future refactor than as the fix for this bug. (More on this below.)

**D. Force our symbols to be global / require callers to use `RTLD_GLOBAL`.**
We can't: *the caller* (Triton, upstream code we don't control) decides how to
load us, and it chooses `RTLD_LOCAL`. Forcing ourselves global would also risk
clashing with PyTorch's bundled `libamdhip64`. Wrong layer — rejected.

**E. Try global first, then fall back to ourselves.**
Almost A, but backwards. If some other `libamdhip64` is in the global table, the
global-first order could return *its* function pointer and mix two runtimes.
Looking at ourselves first is safer, so I used A's order (self first, global only
as a last-resort fallback).

**F. Capture the self-handle in a load-time constructor instead of on first
use.** Same effect as A; I just used a lazy one-time init, which is equivalent
and does no work if the function is never called. Minor.

### What shipped

Option **A**, plus one thing I found while implementing it: the exact same bug
also lived in `hrx_hip_spt_lookup()` (behind the exported `hipGetProcAddress_spt`
and `hipGetDriverEntryPoint_spt`). It was latent (Triton uses the plain entry
point) but real, and the two copies had already drifted. So I put the
self-lookup in **one shared helper** (`iree_hip_self_dl_handle()`) that both call,
which fixes the second entry point and stops them from diverging again. Added a
gtest case for the `_spt` path too.

### Is option C (the hipamd-style table) better?

For *this* bug, no — A fixes it correctly with a fraction of the code and matches
real HIP's observable behavior. C is worth considering later, and only if the
team wants one of these specific things: drop the Linux/loader dependency, honor
the `flags` (stream-per-thread) argument exactly, or resolve symbols we
implement but don't export. If those aren't goals, A is the better long-term
shape too. My recommendation: ship A now; treat C as an optional future
follow-up, not a blocker.

amd-vivekag:
Pushed a follow-up commit fixing the two CMake CI failures (both in the newly added `hip_get_proc_address_test`, not in the shim fix itself):

1. **Link error (Default job):** `ld.lld: undefined symbol: dlopen/dlsym/dlerror/dlclose`. The test calls `dlopen()`/`dlsym()`/`dlclose()` directly but wasn't linking `libdl`; that only works on toolchains where glibc folds `dl` into libc. Added `//build_tools:dl` (→ `${CMAKE_DL_LIBS}`) to the test deps. (The shim `.so` itself was unaffected — shared objects allow undefined symbols, resolved at load.)

2. **Install error (ASAN/TSAN/UBSAN jobs):** `file INSTALL cannot find .../::amdhip64`. `data = [":amdhip64"]` becomes `DATA ::amdhip64`, and the relocatable test-dist installer (`hrx_installed_tests_install_data`) tests `if(TARGET "::amdhip64")` on the raw label — which is not the real target name `libhrx_src_binding_hip_amdhip64` — so it fell through to the file branch and tried to install a file literally named `::amdhip64`. Dropped the target-labeled data edge: under CMake the test already receives the exact artifact path via the `HRX_TEST_LIBAMDHIP64_PATH` compile definition, and an explicit `add_dependencies()` (below the preserve marker) keeps the build order. Under Bazel the artifact can be supplied via `HRX_TEST_LIBAMDHIP64`; absent it the test skips instead of failing.

Verified locally: `-ldl` now on the test link line, both test cases pass, and the previously-failing `--component HrxTestsDist` install of the hip subtree now succeeds (no `::amdhip64`). `CMakeLists.txt` regenerated with `bazel_to_cmake`.

AWoloszyn (approved):

PR #166, opened 2026-07-08
[HIP] Fix host-visible memory and managed globals

AWoloszyn:
This allows us to run some of the hip specific runtime bindings in hrx. For example printf and kernel-side memory allocations.


libhrx/src/binding/common/registry.c
```diff
@@ -782,6 +993,57 @@ static void iree_hal_streaming_context_symbol_map_expunge_module(
...
+    iree_hal_streaming_context_symbol_map_t* map) {
+  IREE_ASSERT_ARGUMENT(map);
+
+  for (iree_hal_streaming_context_module_entry_t* module_entry = map->modules;
+       module_entry; module_entry = module_entry->next) {
+    iree_hal_streaming_module_registration_t* registration =
+        module_entry->registration;
+    for (iree_host_size_t i = 0; i < registration->symbol_count; ++i) {
+      iree_hal_streaming_symbol_registration_t* symbol_registration =
+          &registration->symbols[i];
```
zjgarvey: Might want to hold the registry lock here.

runtime/src/iree/hal/drivers/amdgpu/physical_device.c
```diff
@@ -49,6 +50,34 @@ static_assert(IREE_HAL_AMDGPU_PHYSICAL_DEVICE_GROUP_SEGMENT_MAX_SIZE_DEFAULT !=
                   0,
               "group segment max size default must be non-zero");
 
+// ROCm device-libs initialize hidden_heap_v1 with a 128KiB heap header and
+// 2MiB initial slabs. These constants mirror __ockl_dm_init_v1's heap layout.
```
zjgarvey: Is this stable? E.g., if `__ockl_dm_init_v1` changes or adds a new code, I'd like to be able to at least run a test to make sure these magic numbers here will be compatible.

libhrx/src/binding/common/registry.c
```diff
@@ -47,6 +47,126 @@ static inline uint64_t iree_hal_streaming_symbol_pointer_hash(void* ptr) {
...
+  IREE_ASSERT_ARGUMENT(out_size);
+  if (IREE_UNLIKELY(!iree_host_size_checked_mul(count, element_size,
+                                                out_size))) {
+    return iree_make_status(IREE_STATUS_RESOURCE_EXHAUSTED,
+                            "%s allocation size overflow", kind);
+  }
+  return iree_ok_status();
+}
+
+static iree_status_t iree_hal_streaming_managed_device_name(
```
zjgarvey: This is identical to `module_managed_device_name` in `module.c:1005`.

libhrx/src/binding/common/module.c
```diff
@@ -842,6 +1002,310 @@ iree_status_t iree_hal_streaming_module_global_symbol(
...
+                            "managed storage `%.*s` was declared but not "
+                            "resolved by the executable",
+                            (int)storage_name.size, storage_name.data);
+  }
+
+  char* pointer_name_string = NULL;
+  IREE_RETURN_IF_ERROR(iree_hal_streaming_module_copy_cstring(
+      module, pointer_name, &pointer_name_string));
+  iree_status_t status =
+      iree_hal_streaming_module_initialize_managed_pointer(
```
zjgarvey: This init is also called in `prepare_module` around `registry.c:861`

libhrx/src/binding/common/module.c
```diff
@@ -852,9 +1316,37 @@ iree_status_t iree_hal_streaming_module_global(
...
+  bool managed_found = false;
+  iree_hal_streaming_symbol_t* managed_symbol = NULL;
+  iree_status_t status = iree_hal_streaming_module_try_lookup_global_symbol(
+      module, managed_device_name, &managed_found, &managed_symbol);
+  iree_allocator_free(module->host_allocator, managed_device_name);
+  IREE_RETURN_IF_ERROR(status);
+
+  if (managed_found && managed_symbol) {
+    if (found && symbol) {
+      IREE_RETURN_IF_ERROR(iree_hal_streaming_module_initialize_managed_pointer(
```
zjgarvey: And here (calling init, which does a blocking D2H), on every lookup. This also allocates and frees a "<name>.managed" string on every call. 

Might be worth considering whether to hoist a single `pointer-init` routine, initialize once at load, and cache + lookup. I'm not sure how important this optimization is in practice, but it the blocking transfers seem bad to throw around if not strictly necessary.

libhrx/src/binding/common/memory.c
```diff
@@ -45,7 +45,105 @@ static void iree_hal_streaming_host_memcpy_callback(void* user_data) {
...
+  const iree_device_size_t d2h_chunk_size = 4 * 1024 * 1024;
+  uint8_t* staging_ptr = (uint8_t*)transfer->staging;
+  iree_device_size_t remaining = transfer->size;
+  iree_device_size_t chunk_offset = 0;
+  iree_status_t status = iree_ok_status();
+  iree_slim_mutex_lock(&transfer->context->direct_transfer_mutex);
+  while (remaining > 0 && iree_status_is_ok(status)) {
+    iree_device_size_t this_chunk =
+        remaining < d2h_chunk_size ? remaining : d2h_chunk_size;
+    status = iree_hal_device_transfer_d2h(
```
zjgarvey: Does this get routed through `queue_execute`? If so, I'm concerned that if the d2h queue is the same as the one we put the parent function-call on, we might hit a deadlock.

libhrx/src/binding/hip/api.c
```diff
@@ -613,6 +613,10 @@ static hipError_t iree_hip_get_per_thread_stream_state(
   return hipSuccess;
 }
 
+// HIP reports a successful zero-size malloc by returning NULL. Keep just enough
+// thread-local state for hipMemPtrGetInfo(NULL, &size) to report that result.
+static iree_thread_local bool iree_hip_zero_size_allocation_pending = false;
```
zjgarvey: I looked at o.g. `hipMemPtrGetInfo` and it just unconditionally returns `{0, hipSuccess}` for `NULL`. Is there like a stricter version of this we need where we reject `NULL`?

libhrx/src/binding/common/memory.c
```diff
@@ -45,7 +45,105 @@ static void iree_hal_streaming_host_memcpy_callback(void* user_data) {
...
+        transfer->context->device, transfer->src_buffer,
+        transfer->src_offset + chunk_offset, staging_ptr + chunk_offset,
+        this_chunk, IREE_HAL_TRANSFER_BUFFER_FLAG_DEFAULT,
+        iree_infinite_timeout());
+    chunk_offset += this_chunk;
+    remaining -= this_chunk;
+  }
+  iree_slim_mutex_unlock(&transfer->context->direct_transfer_mutex);
+  if (iree_status_is_ok(status)) {
+    memcpy(transfer->dst, transfer->staging, transfer->size);
```
zjgarvey: I'm not sure I understand why the transfer needs staging. The staging memory is just normal unpinned host memory, right? Why not just do the d2h directly into `transfer->dst`?

libhrx/src/binding/common/registry.c
```diff
@@ -782,6 +993,57 @@ static void iree_hal_streaming_context_symbol_map_expunge_module(
...
+      if (byte_length > (size_t)symbol->size_bytes) {
+        return iree_make_status(
+            IREE_STATUS_OUT_OF_RANGE,
+            "managed symbol `%s` is too small (%zu requested, %" PRIu64
+            " available)",
+            symbol_registration->device_name, byte_length,
+            (uint64_t)symbol->size_bytes);
+      }
+
+      IREE_RETURN_IF_ERROR(iree_hal_device_transfer_d2h(
```
zjgarvey: Do we need to hold the `direct_transfer_mutex` here?

zjgarvey (commented):
Some extra tests might be good.

E.g., managed-global roundtrip, or malformed ELF parsing, or heap-header-population test, or a test checking that a dispatch with a non-zero heap pointer ends up in `implicit_args` would be useful. If these are covered by the hip tests, feel free to skip them (but I'd like to know which tests this PR fixes so I can look at them).

benvanik:
Closing this older dev-base carrier as an exact duplicate of #198. Both PRs point to exact head `15341e0f9f81f689ca5d9c5e9f9bfc8f489d72df`; #198 targets `main` and is the current review surface for this change.

PR #167, opened 2026-07-08
[HIP] Record kernel attributes on module symbols

AWoloszyn:
Populate streaming module symbols from executable metadata, including workgroup limits, fixed memory requirements, cache/carveout state, and uniform-workgroup requirements.

Keep parameter unpack metadata bounded with checked allocation and field-width validation so launch argument layouts fail cleanly when reflected metadata exceeds the stored representation.

Validation: fixed 0 failing tests directly in the isolated suite.


libhrx/src/binding/common/module.c
```diff
@@ -376,28 +444,36 @@ static iree_status_t iree_hal_streaming_module_extract_metadata(
...
-      }
-      parameter_info->direct_arg_bytes = (uint16_t)this_kernel_direct_arg_size;
+    if (!iree_status_is_ok(status)) break;
+    parameter_info->buffer_size = buffer_size;
+    if (IREE_UNLIKELY(export_infos[i].constant_byte_length > UINT16_MAX)) {
+      status = iree_make_status(IREE_STATUS_OUT_OF_RANGE,
+                                "kernel constant layout exceeds metadata "
+                                "field width");
+      break;
+    }
```
zjgarvey, line 556: Redundant with exact same check earlier on (lines 343-348)

https://github.com/ROCm/hrx-system/pull/167/changes#diff-4511cae78c75cccbaa3333fdd41f2313b55bfb60353502304466f0e86df02017R343-R348

libhrx/src/binding/common/module.c
zjgarvey, line 381: The whole `module->symbols` was memset earlier, so this is now redundant.

libhrx/src/binding/common/module.c
```diff
@@ -262,14 +315,28 @@ static iree_status_t iree_hal_streaming_module_extract_metadata(
...
+      symbol->max_threads_per_block =
+          module->context->device_entry->max_threads_per_block;
+    }
+    if (symbol->max_threads_per_block == 0) {
+      symbol->max_threads_per_block = 1024;
+    }
+    symbol->shared_size_bytes = export_infos[i].workgroup_local_memory_size;
+    symbol->local_size_bytes = export_infos[i].private_memory_size;
+    // Register usage is not currently exposed by the executable metadata.
+    symbol->num_regs = 0;
```
zjgarvey, line 403: This was changed from being hard-set at 32 to being hard-set to 0. I assume the change is correct, but I just want to verify.

libhrx/src/binding/common/module.c
```diff
@@ -262,14 +315,28 @@ static iree_status_t iree_hal_streaming_module_extract_metadata(
...
+    symbol->max_dynamic_shared_size_bytes = 0;
+    symbol->preferred_shared_memory_carveout = 0;
```
zjgarvey, line 405: Apparently, `preferred_shared_memory_carveout = 0` means "0% shared / prefer max L1" in HIP, and `-1` means "no preference". Not wired up yet, but might cause issues if we miss this. I'd seed at -1 just in case.

zjgarvey (approved):
A few nits, otherwise looks good to me.

I assume the users for `local_size_bytes` and `uniform_workgroup_size` are just not wired up yet? A few of these symbols are basically write only right now. `cache_mode_ca` I think is actually just not used in ordinary HIP (they just return hardcoded 0, iiuc).

benvanik:
Closing this older dev-base carrier as an exact duplicate of #199. Both PRs point to exact head `be7b8376edab60b9099ff736c56ba63266ff0108`; #199 targets `main` and is the current review surface for this change.

PR #168, opened 2026-07-08
[HIP] Validate pointer ranges and graph memory accounting

AWoloszyn:
Use owner-aware range validation for legacy array copy entry points, report mapped pointer state from the allocation metadata, and keep reusable graph-memory high-water counters tied to current aggregate usage.

Validation: fixed 1 failing test directly in the isolated suite.


libhrx/src/binding/hip/api.c
```diff
@@ -13129,9 +13147,17 @@ HIPAPI hipError_t hipPointerGetAttribute(void* data,
...
+          buffer_ref.buffer->owns_host_ptr ||
+          buffer_ref.buffer->imported_host_allocation;
+      const bool mapped_host_allocation = iree_all_bits_set(
+          buffer_ref.buffer->host_register_flags,
+          IREE_HAL_STREAMING_HOST_REGISTER_FLAG_MAPPED);
+      *(unsigned int*)data =
+          (buffer_ref.buffer->is_managed || !host_allocation ||
+           mapped_host_allocation)
+              ? 1
+              : 0;
```
zjgarvey: Can you explain this change?

I'm looking at how this is handled in ROCm, and I think there it's true whenever a pointer resolves to a memory object. E.g., for device-visible host memory, I think the above logic will resolve to false, but HIP would normally say it's mapped.

https://github.com/ROCm/rocm-systems/blob/3544def6abd4f62d0b1c95be96776b465ffed8df/projects/clr/hipamd/src/hip_memory.cpp#L4218-L4226

libhrx/src/binding/hip/api.c
```diff
@@ -13263,7 +13289,6 @@ HIPAPI hipError_t hipPointerSetAttribute(const void* value,
     case HIP_POINTER_ATTRIBUTE_MAPPED:
     case HIP_POINTER_ATTRIBUTE_BUFFER_ID:
     case HIP_POINTER_ATTRIBUTE_P2P_TOKENS:
-    case HIP_POINTER_ATTRIBUTE_ACCESS_FLAGS:
```
zjgarvey, line 13266: I think the docstring needs to be updated since it says this attr can be set. (~13208).

If the other cases are read-only anyway, shouldn't they also go into the default case (`hipErrorInvalidValue`)?

libhrx/src/binding/hip/api.c
```diff
@@ -7665,6 +7665,15 @@ HIPAPI hipError_t hipMemcpyDtoA(hipArray_t dstArray, size_t dstOffset,
...
   if (result != hipSuccess) HIP_RETURN_ERROR(result);
+  iree_hal_streaming_context_t* context = NULL;
+  result = iree_hip_ensure_context(&context);
+  if (result != hipSuccess) HIP_RETURN_ERROR(result);
+  iree_hal_streaming_buffer_ref_t src_ref = {0};
+  iree_hal_streaming_context_t* owner_context = NULL;
+  result = iree_hip_lookup_streaming_range_with_owner(
+      context, srcDevice, ByteCount, &owner_context, &src_ref);
+  if (result != hipSuccess) HIP_RETURN_ERROR(result);
+  iree_hal_streaming_context_release(owner_context);
```
zjgarvey: I don't really understand this change. I thought the `hipMemcpy` call checks the same thing.

The only difference being that, for a completely unknown device pointer, this block returns `hipErrorInvalidValue` instead of the `hipErrorNotFound` returned from the memcpy's validation. Is that error reporting behavior specific to this old api, or should this just get fixed in a common location?

zjgarvey (commented):
A few questions, since each change seems kinda random to me.

benvanik:
Closing this carrier as subsumed by the newer main-target HIP branches in #198 and #199. Exact head `a3cbb3ae3ca4f93842b98109d2ac593f74a8e546` is an ancestor of both, so every commit in this PR is already carried by those newer review surfaces.

PR #180, opened 2026-07-18
[HAL/Remote] Add remote HAL device execution

benvanik:
Remote HAL turns an IREE HAL device into a distributed object: applications use the normal HAL resource and queue model in a client process, while `iree-serve-device` owns the native device and executes work in a server process. This separates applications, models, weights, and executable selection from the machine that physically hosts the accelerator without introducing a second execution API.

The primary deployment is a trusted private network or compute fabric. A server listens on the network by default so another machine can connect without hidden bind configuration, and startup prints the actual listener plus a directly usable client URI. TCP is plaintext and the protocol deliberately provides no peer authentication, authorization, confidentiality, or active integrity; deployments that require those properties establish them outside IREE with fabric isolation, IPsec, WireGuard, or an equivalent network boundary. SHM relies on operating-system endpoint permissions, RDMA relies on fabric admission, and exposing any transport directly to an untrusted network is outside the deployment contract.

The implementation is a transport-independent networking substrate plus a remote HAL client and server:

```text
client HAL API
  |
  v
remote HAL client
  |
  |  iree/net session bootstrap and immutable device metadata
  |
  +-- control channel: typed resource and lifecycle RPCs
  +-- queue channel: frontier-ordered HAL queue operations
  +-- bulk channel: bounded large payload, file, map, and profile transfer
  |
  v
remote HAL server
  |
  v
native HAL device (AMDGPU, Vulkan, or another server-side driver)
```

Remote HAL maps HAL ownership and synchronization onto these channels instead of emulating every C function with a synchronous RPC. Buffers, executables, command buffers, semaphores, events, files, and virtual-memory reservations have explicit remote identities and release rules. Queue operations retain normal HAL semaphore dependencies, server failures advance or fail the affected signal frontiers, and disconnect converges outstanding work on one terminal session state.

The detailed code and debugging map lives in [`runtime/src/iree/hal/remote/README.md`](runtime/src/iree/hal/remote/README.md).

## Native GPU proof

`iree-remote-check` is the smallest complete proof of the intended product shape. The client embeds AMDGPU HSACO and Vulkan SPIR-V artifacts, connects to the server, selects a target advertised by the native device, uploads only the selected executable, performs host-to-device and device-to-host transfers, dispatches real GPU work, waits on a HAL semaphore, and verifies the exact result. The client links the remote HAL and driver-neutral target-selection utilities but no AMDGPU, HIP, or Vulkan HAL driver; the server is the only process that loads a native driver.

Run the AMDGPU server and client on machines in the same trusted network:

```bash
# Server: expose AMDGPU device 0 on all IPv4 interfaces.
iree-bazel-run \
  --//runtime/config/hal:drivers=amdgpu \
  //tools:iree-serve-device -- \
  --device=amdgpu://0

# Client: select, upload, dispatch, and verify a matching embedded HSACO.
iree-bazel-run \
  --//runtime/config/hal:drivers=remote \
  --//runtime/config/hal:executable_artifacts=amdgpu \
  //tools:iree-remote-check -- \
  --device=remote-tcp://server-host:5000
```

Server startup prints the bound listener and a client device flag. For the wildcard default, replace `<server-host>` with a hostname or interface address reachable from the client; `0.0.0.0` is a bind address, not a client destination. A same-host client can use `127.0.0.1:5000` against the same default listener.

The same client can target a Vulkan server:

```bash
# Server: expose the default Vulkan device.
iree-bazel-run \
  --//runtime/config/hal:drivers=vulkan \
  //tools:iree-serve-device -- \
  --device=vulkan://

# Client: select and dispatch the embedded SPIR-V module remotely.
iree-bazel-run \
  --//runtime/config/hal:drivers=remote \
  --//runtime/config/hal:executable_artifacts=vulkan \
  //tools:iree-remote-check -- \
  --device=remote-tcp://server-host:5000
```

The smoke client is independent of IREE compiler and VM artifacts. Its AMDGPU artifact set follows the configured target selectors, and its Vulkan artifact is assembled independently as Vulkan 1.3 BDA SPIR-V. This keeps the proof focused on native HAL execution, executable compatibility, transfer, dispatch, synchronization, and process-boundary dependency isolation.

### 1. Exact asynchronous lifecycle contracts

The networking stack is built on production async primitives for socket operations, event sources, receive pools, proactor registration, cancellation, and notification-driven shutdown. io_uring, IOCP, POSIX, and fallback paths obey the same ownership contract: every accepted operation has exactly one terminal completion, callback storage remains alive until the backend proves release, and shutdown joins real completion state rather than sleeping for a guessed drain interval.

The highest-pressure invariant is teardown. Listener cancellation, event-source unregistration, multishot receive emulation, nested callback submission, io_uring registration, and proactor destruction all have explicit terminal states. Polling distinguishes internal progress and explicit wakeups from user callback count so completion-driven loops do not mistake a productive poll turn for failure.

### 2. Message carriers and portable TCP

[`iree_net_carrier_t`](runtime/src/iree/net/carrier.h) defines transport-independent framed message movement, send reservations, direct operations, capabilities, and terminal failure behavior. The begin/commit/abort reservation model allows storage to be prepared without copying and permits out-of-order producer commit while preserving completion ownership. Backpressure is represented explicitly instead of becoming an unbounded allocation path.

Loopback provides an in-process semantic reference and TCP provides the default portable cross-machine path. Peer departure is terminal and observable, valid frames may wait for local stream activation during endpoint setup, send storage survives until completion, and listener shutdown cancels pending accepts. Carrier CTS applies the same lifecycle, reliability, pressure, reservation, direct-operation, and zero-copy contracts across implementations.

### 3. Sessions and semantic channels

[`iree_net_session_t`](runtime/src/iree/net/session.h) turns one carrier connection into a negotiated set of endpoints. Bootstrap exchanges protocol versions, topology, frontier axes, transport capabilities, and immutable application payloads before endpoints become visible. The parser bounds every variable-length field and is fuzzed independently; incompatible protocol revisions fail instead of entering a compatibility mode.

Control channels carry typed request, response, and notification traffic. Queue channels carry commands and advances ordered by explicit frontier values rather than carrier arrival or FIFO assumptions. Bulk channels carry transfer-ID-keyed START, DATA, COMPLETE, ABORT, and CREDIT frames. Endpoint detach, GOAWAY, and carrier failure converge on one session failure so a semantic channel cannot remain half-alive after its transport has become terminal.

### 4. Remote HAL resources and queue execution

The remote client exposes an ordinary `iree_hal_device_t`; the server retains one or more ordinary local devices. Session-scoped resource IDs encode type, generation, server proactor placement, and table slot. Client-created resources may begin with provisional IDs so queue work can reference them before a control response returns; the server parks dependent work until the provisional ID resolves and reports the canonical mapping on the corresponding response or queue advance.

The protocol covers allocation, mapping, semaphores, events, executable upload, reusable and one-shot command buffers, queue allocation and deallocation, fill, update, copy, dispatch, file operations, profiling, extension operations, and batched release. Release messages carry the greatest submission epoch that could still reference a resource, preventing control traffic from freeing an object ahead of previously submitted queue work. Queue failures are serialized into ADVANCE messages and surface through the client HAL semaphore path used by local devices.

### 5. Device specifications and executable selection

The server serializes immutable device specifications, memory heaps, and executable target tables into the session bootstrap payload. The remote device is not published until that catalog has been validated and the queue and bulk endpoints are ready. Clients can therefore cache device facts, choose a compatible executable locally, and upload only the selected artifact without paying one synchronous network round trip per query.

Executable loads carry the stable target-table ordinal, queue affinity, load flags, specialization constants, and backend-native bytes. AMD target parsing and family compatibility live in driver-neutral HAL executable support so remote clients, HRX, HIP, and AMDGPU share vocabulary without linking each other's driver implementations. Wire sizes use explicit fixed-width representations, protocol peers must be little-endian, and mismatched revisions are rejected.

### 6. Bounded bulk transfer, files, and profiling

Large payloads move independently of latency-sensitive control and queue traffic. Transfer tables, chunk descriptors, peer credits, receive windows, staging pools, and owned upload/download components keep memory bounded under pressure. DATA chunks may arrive out of order, but each transfer reservation covers a precise byte range and completion, cancellation, or failure retires every descriptor exactly once.

Remote file operations support both client-local files bridged over bulk transfer and server-local files exposed through an explicit logical allow-list. Clients never submit arbitrary server paths; final path and access validation happen against the server-owned namespace. Profiling relays server sink callbacks into a client-owned sink while retaining session and callback state through terminal completion, so profile backlog cannot outlive its connection.

### 7. Virtual-memory operations

Remote allocator support carries virtual address reservation, physical allocation, map, unmap, and access changes across the protocol. Virtual and physical resources have distinct message namespaces, queue affinity and access-agent selection are preserved, and the server returns opaque HAL addresses rather than exposing a native driver object to the client.

AMDGPU implements the corresponding allocator operations and advertises only capabilities it can execute. Partial failures unwind mappings, reservations, and physical resources through the same ownership graph. This is the generic pointer-stability mechanism; higher-level sparse pool policy and HIP pointer semantics are intentionally not encoded into the remote HAL allocator contract.

### 8. Cross-process shared memory

The SHM carrier provides same-host IPC with shared rings, adaptive progress, a cross-process wake primitive, direct registered-buffer reads and writes, and endpoint bootstrap over Unix domain sockets or Windows named pipes. Unix descriptor and Windows HANDLE transfer support remote file sidebands without copying file contents through control messages.

Operating-system endpoint permissions define admission. Shared mappings and transferred handles have one owner, wakeups cannot be lost across the arm/sleep boundary, peer death wakes stranded waiters, and in-process carrier tests are supplemented by real cross-process multi-endpoint coverage on both POSIX and Windows implementations.

### 9. RDMA transport and remote memory windows

The Linux RDMA carrier uses librdmacm and libibverbs through dynamic loaders, so an RDMA-enabled binary can still start on a system without those libraries. The implementation owns CM event handling, device and protection-domain context, completion queues, queue pairs, receive credits, send windows, work-request retirement, registered staging, direct read/write, endpoint negotiation, and exported memory windows. It supports RoCE and InfiniBand deployment shapes and includes DMA-BUF registration probing for GPU-direct paths.

CM and CQ failures remain terminal, every work request retires once, queue depth and explicit credits bound outstanding traffic, and registration lifetime spans every DMA that references it. Requesting RDMA is a hard capability requirement and never silently downgrades to ordinary TCP bulk semantics. RDMA is Linux-only and opt-in in both build systems; disabled `//...` builds do not fetch its headers or pull its libraries.

### 10. Deployment and dependency boundaries

`iree-serve-device` defaults to `tcp://0.0.0.0:5000` because the intended use is another trusted machine, not an accidental loopback-only demo. Readiness is printed only after the listener starts, wildcard addresses become a usable client template, explicit loopback binds are classified as same-host, and SHM reports its operating-system permission boundary. Network-visible startup reports the surrounding trust boundary and the absence of protocol authentication or encryption; any peer admitted through that boundary can submit arbitrary HAL work to the exposed device.

Build configuration lives under `//runtime/config/hal` and `//runtime/config/net`, with TCP enabled by default and SHM/RDMA explicit transport choices. Bazel is the source of build graph truth and generated CMake metadata carries equivalent conditions. Remote-only tools link the remote HAL client and neutral executable-target support but no native HAL driver; HIP never links AMDGPU, AMDGPU never links HIP, and native drivers remain on the server side of the process boundary.

## Test architecture

Carrier CTS is reusable across loopback, TCP, SHM, and RDMA and exercises reservation ordering, backpressure, reliability, lifecycle, direct operations, and terminal errors. Remote HAL CTS wraps ordinary HAL CTS backends behind an in-process server/client pair so the same queue and resource contracts are exercised through serialization. Dedicated integration coverage exercises real TCP sessions, cross-process SHM, server slot reuse, file transfer, bounded bulk pressure, profiling relay, virtual memory, disconnect, bootstrap parsing, malformed frames, and native AMDGPU/Vulkan executable selection. Bootstrap, framing, control-channel, and carrier parsers have fuzz targets at their untrusted length boundaries.

## HRX scope boundary

This branch includes the narrow HRX device-selection and lifecycle bridge used to prove that an HRX process can create a remote HAL device, derive its AMDGPU architecture from the bootstrapped device spec, and run ordinary stream, buffer, and transfer paths without loading a native GPU driver in the client. That probe exposed generic session startup ordering, allocator capability, and receive-pool lifetime bugs that are fixed at their owning layers here.

It does not claim production HIP-on-HRX remoting. That consumer requires a stronger native device-pointer identity contract for nested pointers, sparse virtual-memory-backed pools that do not commit most of device memory on the first small allocation, post-connect failure propagation through HIP synchronization, and a remote-only HRX dependency graph without local-task or VM implementations. Those concerns form a separate consumer review surface and are not hidden behind compatibility fallbacks in this PR.

## Reviewer notes

The highest-value review questions are:

- Does every accepted asynchronous operation and relay reach exactly one terminal completion before caller-owned storage can be released?
- Does every cross-machine ordering edge pass through an explicit frontier, with exactly one advance or failure path?
- Does every remote resource have a session-local owner, a stable provisional-to-canonical resolution path, and a release condition tied to observed queue progress?
- Is every wire length, count, offset, and padded command size validated before allocation, pointer arithmetic, or copy?
- Can every large payload move without blocking control/queue traffic or allowing credits, descriptors, staging memory, or callbacks to grow without bound?
- Do SHM descriptors/HANDLEs and RDMA registrations/work requests remain owned through the final kernel or peer-visible completion?
- Does a remote-only client stay free of native HAL driver code across both Bazel and CMake dependency closures?
- Is the trusted-fabric boundary visible at build, startup, documentation, and failure diagnostics without implying that the protocol itself authenticates or protects peers?

The branch history is organized by these ownership and invariant boundaries so the individual mechanisms remain useful archaeology, but the review unit is the integrated system: lifecycle, transport, protocol, HAL semantics, native execution, and teardown must agree at the branch head.


zjgarvey:
This review focused on the networking and remote-HAL stack (`iree/async`, `iree/net`, `iree/hal/remote`), the AMDGPU virtual-memory additions, and the serve/check tooling. The loopback CTS exercises the happy-path queue/resource contracts well, and the AMDGPU/GPU-execution path and the remote-only client's driver isolation are clean in the reviewed paths. The findings are all in the transport/protocol/lifecycle layers, and they cluster in one place: the error/teardown and cross-thread completion paths that loopback's inline, same-thread completions never reach. Most only surface on a real out-of-process carrier (TCP/SHM/RDMA), where a completion lands on a different thread and a peer can die mid-operation. They are ordered below by how likely they are to actually fire.

## Memory safety and liveness

**1. Backpressured control frame concatenates with the next → GOAWAY silently lost.** In `net/channel/util/frame_sender.c`, `flush` intentionally retains the batch on a `RESOURCE_EXHAUSTED` submit (`:535`, "DO NOT release batch_lease") with `batch_used` unchanged, so the next `frame_sender_queue` appends at the same offset and the following flush emits `[retained frame][next frame]` as one carrier message. The control wire format is one-frame-per-message with no intra-message length prefix, so the receiver parses only the first header and treats the trailing frame as its payload. Concretely: `handle_ping` (`net/channel/control/control_channel.c:256`) queues a PONG and flushes; under backpressure the PONG is retained and the auto-PONG path swallows the `RESOURCE_EXHAUSTED` and returns OK (`:266-271`), so nothing compensates, and a subsequently sent GOAWAY is fused into the retained PONG and lost while `send_goaway` still transitions the local channel to DRAINING (`:763-766`). `send_ping`/`send_goaway` share the same retain-on-error path, so a caller retry also re-queues and concatenates. Triggers: carrier backpressure, a normal state under load. Fix: rewind the just-queued frame on any flush failure, or send each protocol frame via its own `send_copy` rather than sharing the batch.

**2. Backpressured queue submit orphans an epoch → every resource release stalls permanently (session-fatal).** In `iree/hal/remote/client/queue.c`, the submission epoch is assigned by an unconditional `fetch_add` (`:769`) before the COMMAND frame is sent (`send_queue_op`, `:794`). `iree_net_message_endpoint_begin_send` can return `RESOURCE_EXHAUSTED` on a healthy channel — ordinary backpressure — and the failure path (`:805-812`) fails only the signal semaphores; it does not roll back `next_submission_epoch` or set an error state, so the channel keeps running with epoch N consumed but never sent. The server advances `observed_submission_window` only when a submitted command observes its epoch, so the contiguous prefix pins at N-1 forever; every later queue-ordered resource release is deferred against `>= N` (`server/session.c:5254`) and never runs — buffers, executables, command buffers, events, and files strand in the deferred list, the fixed-size resource table fills, and all further allocations/creates fail `RESOURCE_EXHAUSTED`, recoverable only by reconnect. Triggers: a single transient send backpressure event on a real carrier under load, assuming the app treats the retryable submit status as retryable and keeps submitting (the conventional contract). Fix: publish the submission epoch only after the COMMAND frame is admitted to the transport, or emit a compensating server-observable skip for an abandoned epoch.

**3. Use-after-free of a stack-owned pending RPC on disconnect.** `iree_hal_remote_client_device_fail_pending_rpcs` (`iree/hal/remote/client/device.c:1261-1276`) detaches the pending-RPC list under `rpc_mutex`, unlocks (`:1266`), then stores the status and posts each waiter's notification outside the lock (`:1270-1273`). The response path `on_control_data` (`:1636-1667`) deliberately does its store-and-post while holding `rpc_mutex` and leaves the entry linked, precisely so a woken waiter's unlink-and-deinitialize (which re-takes the lock at `:1390`) cannot free the stack-owned entry (`:1332`) mid-post. `fail_pending_rpcs` breaks that ordering: once it unlocks, a waiter woken by a concurrent `on_control_data` can unlink, deinitialize, and reclaim its stack frame before the loop dereferences the same entry → write-after-free plus a post on a freed notification word. Triggers: a control response landing as a disconnect/error runs `fail_pending_rpcs` (from `deactivate` on an app thread, or `notify_bulk_transport_error` on a bulk-transport thread) — the exact "response races teardown" case it exists to handle; because `on_control_data` never unlinks, the race spans its unlock through the waiter's re-lock for every response still in flight at teardown. Fix: do the per-entry status store and post while holding `rpc_mutex`, mirroring `on_control_data`.

**4. Bulk map-read deadlocks forever on transport error / teardown.** A mapped read larger than the inline control capacity registers a completion callback bound to a stack-local wait (`iree/hal/remote/client/buffer.c:303`), sends BUFFER_MAP, and on OK blocks the app thread on `iree_notification_await(..., iree_infinite_timeout())` (`:383`). That callback is the only waker and is fired only on DATA/COMPLETE/ABORT. But `notify_bulk_transport_error` (`device.c:1278-1289`) fails only the control RPCs and invokes the error callback — it never walks the bulk transfer table (the download receiver has no fail-all-in-flight path; `cancel_transfer` handles only FILE_WRITE) — so a hard bulk/peer failure after the BUFFER_MAP response returns OK but before the data frames arrive never wakes the waiter. Triggers: ordinary peer/network failure during the large-payload window. Fix: give the download receiver a terminal "fail all in-flight transfers" path, invoked from `notify_bulk_transport_error` and from teardown, that completes each transfer with the error status.

**5. Profile transfer freed while still linked in the sequence window → heap UAF / double-free from untrusted frames.** In `iree/hal/remote/client/bulk_profile_receiver.c`, an out-of-order profile COMPLETE is deferred into `profile_sequence_window` via the transfer's embedded `sequence_node` (`:623-627`, with `profile_transfer_transferred=true` so it is not released), but the transfer is left in the bulk transfer table with `profile_receive` set — two owners. `release_transfer` (`:45`) frees the object and drops the table entry but never unlinks the window node (the window has no per-node removal; `release_pending_locked` shows the safe teardown that `on_abort` bypasses), and `on_abort` (`:660-681`) calls `release_transfer` on the still-present table entry. This is the most gated item here — it needs an active profiling session — but it is heap corruption from untrusted input: triggers are a server sending profile transfers out of order (the case the window exists for) followed by an ABORT or a duplicate COMPLETE, with no terminal-frame dedup at the bulk layer; the window's next `observe`/`take_pending` then dereferences the freed node (UAF) or re-frees it (double-free). Fix: add a window remove-node call before freeing a deferrable transfer, or drop the table entry at defer time so the window is the sole owner.

**6. NULL-deref in the server-create error path.** `server->sessions` is assigned only inside the `if (iree_status_is_ok(status))` block at `iree/hal/remote/server/server.c:341`, which runs after `create_device_catalog` (`:327`). On catalog failure the assignment is skipped and `sessions` stays NULL (from the `memset` at `:286`), but the error path calls `destroy`, whose cleanup loop indexes `&server->sessions[i]` for `i < max_connections` (`:375-378`, and `max_connections` is set at `:293` before the catalog build) and dereferences it. Triggers: `create_device_catalog` failing — a device that exposes no HAL device spec, or allocation pressure during catalog build — crashes at construction instead of returning the error. Fix: assign `server->sessions` right after the `memset` (the trailing storage is already zeroed and every deinit tolerates a zeroed slot), or guard the destroy loops on a non-NULL `sessions`.

**7. Double-close of a caller-owned fd on SHM bootstrap failure.** In `net/carrier/shm/factory_bootstrap.c`, `prepare` copies the caller's channel fd into `bootstrap->channel` (`:224`) but clears it back to none on failure only inside `if (bootstrap->worker_thread)` (`:289-290`). Any failure before the worker thread is created (shared-wake, `handshake_results` malloc, notification/event create, all at `:236-255`) leaves the caller's fd in `bootstrap->channel`, and `iree_net_shm_bootstrap_destroy` closes it (`:79`) — while the header contract (`factory_bootstrap.h:62-63`) states the caller retains and closes the channel on failure. Triggers: resource exhaustion during SHM connect (a cold path); under concurrency the fd number can be reused between the two closes, so the second close hits an unrelated descriptor. Fix: set `bootstrap->channel` to none unconditionally on the failure path before destroy, or defer copying the fd until the success point.

**8. Unbounded stack write in `send_ping` / `send_goaway` (latent).** `iree_net_control_channel_serialize_frame` (`net/channel/control/control_channel.c:163-189`) memcpys header + sub-header + trailing into the destination with no capacity argument and no clamp. `send_ping` (`:689`) and `send_goaway` (`:738`) pass fixed 256-byte-tail stack buffers with caller-supplied `payload.data_length` / `message.size`, so a string over 256 bytes overruns the stack. Triggers: no in-tree call site passes a long string today, but the sole caller `iree_net_session_shutdown` (public `IREE_API_EXPORT`, `session.c:2108`) forwards an application-supplied reason string straight to `send_goaway`, so any consumer passing a message over 256 bytes overruns. Latent for in-tree use, but an unbounded write reachable through the public API. Fix: give `serialize_frame` a destination capacity and validate length up front (return `OUT_OF_RANGE`), or route oversized frames through the pooled send path.

## Correctness and robustness on real-carrier paths

**Server frontier window not closed on failure → stranded and leaked completions.** In `iree/hal/remote/server/session.c`, `submit_command` reserves both `observed_submission_window` and `completed_signal_window` for the signal epoch. The former is observed on every failure path, but `completed_signal_window` is observed only on success (`:989`, under the `if (ok)` guard at `:976`); on execution failure `on_command_complete` skips it, and on setup failure `submit_command` closes only `observed_submission_window` (`:1780`). The failed epoch is reserved but never closed, so its hole pins the completed prefix and every later completion is parked and leaked (`:1001-1003`). For a correct client this is a bounded leak released at teardown; a client that keeps submitting after an axis failure grows `sequence_window` without bound. Fix: observe/close the `completed_signal_window` hole on the execution-failure and setup-failure paths, mirroring what is already done for `observed_submission_window`. (This is the server-side companion to item 2 — same frontier, different fix.)

**SHM client trusts peer-advertised ring geometry → OOB.** In `net/carrier/shm/handshake.c`, the client maps using `offer_header.region_size` and checks only the region magic and `ring_capacity` equality (`:454-468`); it never re-validates `ring_capacity` (power-of-two, `>= MIN_CAPACITY`) or that the mapping is large enough for the derived ring extent the way the server does at create (`:234-254`). `init_rings` (`:110-137`) then derives ring B's base purely from `ring_capacity` and opens rings there; the internal size check uses the same peer-derived value, so it cannot catch a mapping overflow. Triggers: a buggy, version-skewed, or compromised same-host server advertising a small `region_size` with a large `ring_capacity` → OOB read / SIGBUS. Fix: validate `ring_capacity` and require the mapping to cover `OFFSET_RINGS + 2 * required_size(ring_capacity)` (overflow-checked) before `init_rings`.

**Large `update_buffer` / `dispatch` truncate the 16-bit command length → server stream misparse.** In `iree/hal/remote/client/command_buffer.c`, `update_buffer` (`:719-758`, store at `:736`) and `dispatch` (store at `:913`) compute an overflow-checked `total_size` that can legitimately exceed 65535, then store `cmd->header.length = (uint16_t)total_size` with no `cb_validate_command_length` call (unlike `execution_barrier` and `wait_events`, which validate before the store); the server iterator advances by the truncated `length`. Triggers: a legal HAL `update_buffer` with a >64KB inline payload, or a large-constant dispatch (constants can reach ~256KB). The server's outer loop bounds-checks `offset + length`, so the outcome is a stream misparse / wrong result rather than a server OOB — but it is silent. Fix: call `cb_validate_command_length` before writing `header.length`, and reject or chunk oversized payloads.

**Cross-type release through FILE_CLOSE.** In `iree/hal/remote/server/resource_table.c`, `detach`/`release` (`:88-108`) validate slot and generation but not resource type (the header at `resource_table.h:63-64` says they validate type, and `lookup` does). `handle_file_close` (`session.c:5733`) resolves a client-supplied id and calls `resource_table_release` directly (`:5748`), so a client passing a resolved BUFFER id to FILE_CLOSE releases that buffer through the generic path — which for a virtual-reservation buffer bypasses `iree_hal_allocator_virtual_memory_release` (the code warns about exactly this at `session.c:458`) and the epoch-gated ordering. Same-session scope only. Fix: add the type check to `detach`, and route file-close through the virtual-buffer-aware release helper.

**Server file allow-list is lexical-only (symlink escape).** In `iree/hal/remote/server/file_index.c`, `resolve_host_path` (`:396-419`) enforces containment only with the lexical `iree_file_path_is_portable_relative` (`base/internal/path.c:179-255`, which inspects the string with no filesystem access), then string-joins and opens (`:462-473`) with `open()` using no `O_NOFOLLOW`/`realpath`. A symlink inside an exposed directory whose target is outside it is followed at open. Triggers: a directory allow-list entry whose tree contains a symlink pointing out of the root (model snapshot directories are commonly symlink farms) — the client then reads, or overwrites in place, the target. Bounded by the trusted-fabric model and the granted R/W bits, and no new files can be created (`O_CREAT`/`O_TRUNC` are never set), but it is an independent host-filesystem capability that "arbitrary GPU work" does not imply, and the error text claims the suffix cannot "escape the allowed logical directory." Fix: `realpath`-canonicalize and check prefix-containment, or open with `O_NOFOLLOW` / `openat2(RESOLVE_BENEATH)`.

## Testing

The common thread across the items above is that loopback completes inline on the submitting thread, so the failure, teardown, and cross-thread completion edges are largely unexercised — items 1 through 4 and the drain/credit lost-wakeup paths are not reachable with same-thread inline completions. An out-of-process carrier test (TCP or SHM) that induces peer death mid-transfer and mid-RPC, plus a backpressured-send test, would surface most of this class. The intricate surfaces with the least coverage are the partial-range virtual-memory unmap (only full-range removal is tested) and the network-facing session bootstrap parser negative paths (no fuzzer or negative test — the one bootstrap fuzzer targets the device-catalog parser, and even that only exercises single-entry catalogs).

## Lower priority

Beyond the above there are a number of lower-severity items, mostly doc/impl divergences where a comment or header describes behavior the code does not implement: DATA/DATA_END queue fragmentation is documented (`frame.h`, `queue_channel.h`) but returns `UNIMPLEMENTED` and terminates the session (`queue_channel.c:342`) — no in-tree sender emits it, so it is latent interop rather than a live bug; the auto-PONG "response is omitted" comment (`control_channel.c:267`, the frame is actually retained — see item 1); and the `resource_table` "validates type" header contract (see the cross-type item above). There are also three oversized translation units that tangle several state machines (`rdma/carrier.c` ~3787, `client/queue.c` ~2953, `io_uring/proactor.c` ~2407), a few dead branches around always-OK `acquire_timepoint` rollbacks, and assorted narrow validation gaps (reserved-field and unknown-flag acceptance, an alloc-before-validate in `PROFILING_BEGIN`). Exact line references for these are available on request.

Reviewed at 210c880e.

benvanik:
Closing this historical merge carrier. The remote branch `users/benvanik/remoting` is intentionally retained at exact head `499c600a713f87337e390b9cbd7c03926a2957c8` for architecture archaeology and focused re-extraction; closing this PR does not delete it. Current remoting work should return as reviewable units on the current runtime rather than rebasing this 539-file snapshot.

PR #183, opened 2026-07-20
[Loom] Analyze final native artifacts before publication

benvanik:
Make final native bytes an explicit compiler transaction point. A compiler embedding can synchronously analyze the exact artifact produced by the selected target provider, publish structured results into the compile report, and reject the candidate before any executable artifact becomes observable.

## Analyzer contract

The target-neutral callback borrows the final byte span, artifact format, target key and bundle, diagnostic sink, reporting limit, and candidate-owned compile report for one synchronous call. The borrow lasts for that call; candidate preparation retains ownership of request storage and compiler state.

Candidate preparation invokes analysis only after native emission succeeds and before publication. Analyzer failure leaves the candidate unpublished. Candidate teardown remains the single owner of targets, artifact bytes, and report storage, and analyzer-written report data remains valid through teardown for both clean and rejected outcomes.

## Structured results

Compile reports distinguish no analysis, a complete clean result, a complete rejection, and an analysis error. Text and JSON include analyzer identity and ABI version, independently observed artifact target, decoded instruction and memory-event counts, discovered and analyzed functions, findings observed and reported, truncation, early termination, completeness, and the final verdict.

AMDGPU wait hazards have a typed diagnostic bridge carrying the stable hazard code, kernel attribution, section and file offsets, producer and consumer instructions, register class and span, counter identity, required progress, and explanation. Final-byte analyzers therefore use Loom's normal diagnostic and report surfaces while preserving machine-readable finding data alongside human explanations.

## Embedding model

The compiler embedding owns analyzer discovery, dynamic-library loading, ABI negotiation, selection policy, and required-versus-advisory behavior. The core compiler owns callback invocation and the structured result schema. Target providers own native artifact production. These boundaries let analyzer implementations and loading policy evolve independently of candidate ownership and target emission.

## Reviewer Notes

The highest-value review surfaces are transactional ordering in `hal/candidate.c`, request and report lifetime ownership, rejection before publication, and the stability of the target-neutral report and typed-hazard schemas.

### DIFF
--- a/loom/py/loom/error/amdgpu.py
+++ b/loom/py/loom/error/amdgpu.py
@@ -1027,6 +1027,49 @@ ERR_AMDGPU_044 = ErrorDef(
     ),
 )
 
+# ERR_AMDGPU_045: AMDGPU final artifact has an insufficient wait.
+ERR_AMDGPU_045 = ErrorDef(
+    domain=ErrorDomain.AMDGPU,
+    code=45,
+    severity=Severity.ERROR,
+    summary="AMDGPU final artifact has an insufficient wait.",
+    message=(
+        "AMDGPU final-artifact wait hazard '{diagnostic_code}' for kernel "
+        "'{kernel_name}' (present={has_kernel}, entry .text+{kernel_entry_offset}) "
+        "at {section_name}+{consumer_section_offset} "
+        "(file+{consumer_file_offset}): instruction '{consumer_instruction}' "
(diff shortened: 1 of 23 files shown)


benvanik:
Closing for now. This will return with the rocJitsu integration so the analyzer seam, structured findings, publication rejection, and lifetime behavior land as one real production path with end-to-end artifact tests instead of standalone plumbing.

PR #186, opened 2026-07-20
[Loom] Require speculation safety for LICM

benvanik:
Loop-invariant code motion now treats moving work before an enclosing loop as speculation, requiring every moved non-terminator to declare that it is safe to execute under a broader dynamic predicate.

A loop body can execute zero or many times, and hoisting may execute a candidate on a new path or reorder a trap ahead of observable body work even for an exact single-trip loop. PURE establishes deterministic, effect-free computation; it does not establish that broader execution and ordering are safe.

The pass delegates to the shared subtree speculation analysis, preserving its SSA, type, attribute, and nested-region availability checks. Coverage distinguishes pure-only work that remains nested from total comparison and selection chains that hoist through loops and nested conditionals.

### DIFF
--- a/loom/src/loom/transforms/loop/licm.c
+++ b/loom/src/loom/transforms/loop/licm.c
@@ -23,7 +23,7 @@ LOOM_PASS_STATISTICS_DEFINE(loom_licm_statistics, loom_licm_statistics_t,
 
 static const loom_pass_info_t loom_licm_pass_info_storage = {
     .name = IREE_SVL("licm"),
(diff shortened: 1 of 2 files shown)


benvanik:
CI triage: the MSAN, UBSAN, and GCC jobs all failed while Bazel fetched the same SPIRV-Tools v2026.1 archive from GitHub (`SocketTimeoutException: Read timed out`), before this change compiled. PR #187 and the later stack heads contain this commit and passed those CPU configurations.

PR #189, opened 2026-07-20
[Loom] Describe scalar view movement exactly

benvanik:
Ordinary scalar view loads and stores now enter the target-neutral movement model as exact one-element byte transfers. Each endpoint retains its storage root, alias scope, nullability, memory space, symbolic byte interval, alignment, typed schema, and cache policy.

Async legality and configured interleaving consume one shared endpoint-to-region projection instead of reconstructing partial view summaries. Configured interleaving can therefore commute stores to explicitly disjoint noalias roots while preserving order for unrelated external roots and unresolved byte ranges.

The SCF dialect no longer carries its stale dependency on motion analysis, keeping the dependency graph acyclic as movement analysis becomes a shared legality substrate.

### DIFF
--- a/loom/src/loom/analysis/BUILD.bazel
+++ b/loom/src/loom/analysis/BUILD.bazel
@@ -537,6 +537,7 @@ loom_cc_test(
         "//loom/src/loom/ops/kernel",
         "//loom/src/loom/ops/test",
         "//loom/src/loom/ops/vector",
+        "//loom/src/loom/ops/view",
         "//loom/src/loom/util:fact_table",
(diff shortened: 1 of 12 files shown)


benvanik:
CI triage: the gfx110X job ran on a host whose HSA runtime reports `HSA_AMD_SYSTEM_INFO_VIRTUAL_MEM_API_SUPPORTED == false`; the AMDGPU driver therefore rejected device creation and the resource lane failed broadly. PR #190 contains this commit and passed gfx110X on a VMM-capable runner.

PR #192, opened 2026-07-20
[Loom] Place invariants against residency cliffs

benvanik:
Introduce target-informed loop-invariant placement as a first-class source-to-low planning phase. Legal invariants are moved as dependency closures only when the resulting structured pressure envelope satisfies the enclosing loop's residency contract, allowing reuse to be exposed without accidentally crossing a target occupancy or residency cliff.

## Author-facing residency contract

`scf.for` gains one optional residency policy with three behaviors:

| Source form | Contract |
| --- | --- |
| No clause | Select legal placements automatically. A target pressure model protects the current cumulative projected tier; targets without a model still receive semantics-preserving invariant motion. |
| `residency(preserve)` | Preserve the tier represented by the authored placement under the selected target's source-pressure model, and retain the authored boundary as the recovery point when exact post-allocation evidence disagrees with that projection. |
| `residency(%minimum)` | Resolve an exact SSA-valued minimum through value facts and require at least that tier. This form may intentionally trade down from the authored projected tier when the requested minimum permits it. |

The categorical and SSA-valued forms are mutually exclusive. Parsing disambiguates their shared `residency(...)` spelling by payload kind, and text, bytecode, builders, verification, canonicalization, and structured loop rewrites preserve the contract consistently. Exact minima are rejected when they are unresolved, negative, outside the representable range, or stronger than the selected target can provide.

## Structured pressure model

A region-tree pressure analysis gives the function body and every loop body an explicit pressure scope. It composes nested loops, sibling scopes, mutually exclusive branch peaks, loop-carried state, and values live across joins without flattening the function into one misleading peak. Each scope records its baseline, required, and current tier, while the function-level envelope ensures a locally acceptable move cannot silently regress another active scope.

Target resource records identify direct and derived pressure, the program point limiting each resource, the next stronger and weaker tier facts, and named reserve contributions. Reserve completeness is retained in the analysis result, so placement decisions and reports distinguish a complete target envelope from a partial estimate instead of presenting both with the same confidence.

## Whole-function placement planning

`place-loop-invariants` selects the target descriptor and lowering policy for each source function, acquires exact value facts, and shares motion, liveness, and pressure analyses across the complete structured region tree. Loops are processed deepest-first with stable source ordering so inner placement decisions become visible to enclosing loops while remaining deterministic.

Candidate discovery uses the common motion legality contract and proceeds in simultaneous fixed-point waves. A wave is collected before any member moves, preventing lexical traversal from making later operations appear invariant only because an earlier candidate was already relocated. Subsequent waves can then discover consumers exposed by a completed dependency wave. Conditional computations remain within their dynamic predicate, inner invariants may move to an enclosing loop boundary, and independent sibling closures are evaluated in the same function pressure context.

Each loop first tries its complete legal dependency closure in one pressure evaluation. This matters when moving a provider alone temporarily lengthens a live range but moving its consumer at the same time shortens the combined lifetime enough to retain the required tier. If the full closure crosses a cliff, the planner restores exact authored positions, builds dependency components from SSA operands, tries the largest dependency-ordered prefix, and then evaluates independent later branches. Refinement is capped at 32 evaluations per loop so compilation remains bounded and deterministic; unselected operations stay at their authored boundary.

Every trial recomputes all affected pressure scopes from the current IR. Automatic and preserve policies protect the accepted cumulative function tier. A numeric minimum establishes the hard floor instead, and a frontier that begins below that floor must make measurable progress toward it. A successful decision is committed as the next baseline for subsequent loops; failure to find any legal placement satisfying an explicit policy is diagnosed at the source contract.

## Durable source-to-target handoff

Successful planning emits compiler-owned `scf.residency.require` and `scf.residency.candidate` markers. The function requirement carries the hard minimum or preserve intent together with the projected authored baseline. Each candidate retains a stable identity, its source value and captures, recomputation cost, baseline-protection intent, fact traits, and an exact source-operation witness covering the operation shape and canonical attributes.

The marker representation is deliberately narrow: it records enough source evidence for exact post-allocation validation and authored-boundary recovery without importing register-allocation mechanics into the structured planner. Candidate results referenced through predicate attributes or dynamic types remain at their authored boundary because those references are outside ordinary SSA operand use lists and therefore cannot yet be redirected and restored exactly.

Authored loop policies are consumed only after planning and marker materialization succeed. Configured unrolling and SCF-to-CFG diagnose any policy that reaches them unconsumed, turning a pipeline-order or unsupported-target mistake into a loud lowering failure instead of silently erasing the residency requirement. Existing unroll factors, schedules, tied results, and result names survive policy consumption.

## Pipeline and observability

The normal source-to-low pipeline runs placement after eager target legalization and before configured unrolling. Target-specific behavior enters through the selected descriptor, source-pressure model, named reserves, and exact value facts; the planner and author contract remain target-neutral.

Structured reports expose each loop's policy, outcome, reason, baseline, selected and required tiers, candidate and rejection counts, limiting resources, reserve contributions, and the selected or retained state of every candidate. Pass statistics summarize functions, loops, candidates, hoists, pressure rejections, refinement work, consumed policies, and functions without a model, making the planning decision inspectable without changing generated code.

### DIFF
--- a/loom/py/loom/dialect/scf/builders/__init__.pyi
+++ b/loom/py/loom/dialect/scf/builders/__init__.pyi
@@ -104,3 +104,24 @@ class ScfBuilder(DialectBuilder):
         result_names: Sequence[str] | None = ...,
         location_id: int | None = ...,
     ) -> list[ValueRef]: ...
+    def require(
+        self,
+        *,
+        minimum: int,
+        preserve: bool,
+        projected_baseline: int,
+        location_id: int | None = ...,
+    ) -> None: ...
+    def candidate(
+        self,
+        *,
+        candidate_id: int,
+        recompute_cost: int,
+        source: ValueRef,
+        results: list[Type | TiedResultSpec],
+        preserves_baseline: bool,
+        name: str | None = ...,
+        names: Sequence[str] | None = ...,
+        result_names: Sequence[str] | None = ...,
+        location_id: int | None = ...,
+    ) -> ValueRef: ...
--- a/loom/py/loom/dialect/scf/builders/__init__.pyi
+++ b/loom/py/loom/dialect/scf/builders/__init__.pyi
@@ -22,6 +22,8 @@ class ScfBuilder(DialectBuilder):
         unroll_factor: ValueRef | None = ...,
         unroll_policy: str | None = ...,
         unroll_schedule: str | None = ...,
+        residency_minimum: ValueRef | None = ...,
+        residency_policy: str | None = ...,
         body: Region | None = ...,
         name: str | None = ...,
         names: Sequence[str] | None = ...,
--- a/loom/py/loom/dialect/scf/defs.py
+++ b/loom/py/loom/dialect/scf/defs.py
@@ -112,6 +112,18 @@ ScfForUnrollSchedule = EnumDef(
     doc="Local scf.for unroll body ordering.",
 )
 
+ScfForResidencyPolicy = EnumDef(
+    "ScfForResidencyPolicy",
+    [
+        EnumCase(
+            "preserve",
+            1,
+            doc="Protect the target source-pressure model's authored-placement tier, falling back to the exact authored tier after a projection miss.",
+        ),
+    ],
+    doc="Target-informed scf.for loop-invariant placement policy.",
+)
+
 # ============================================================================
 # scf.yield — variadic region terminator
 # ============================================================================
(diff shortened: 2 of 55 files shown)


benvanik:
CI triage: the gfx110X job ran on the known non-VMM host and every AMDGPU execution path failed at device creation with `HSA_AMD_SYSTEM_INFO_VIRTUAL_MEM_API_SUPPORTED == false`. PR #193 contains this commit and its fresh gfx110X job passed on a VMM-capable runner.

benvanik:

PR #193, opened 2026-07-20
[Loom] Enforce residency after physical allocation

benvanik:
Validate authored residency contracts after physical allocation and repair exact cliff regressions only through finite materialization alternatives proven legal by source motion. Source placement requirements and producer witnesses lower into target-low markers, survive configured unrolling and structural lowering, and are consumed before final emission into an allocator-independent exact contract.

Final target-low preparation seals every candidate as an explicit SSA-tracked producer dependency DAG after target lowering, canonicalization, operand selection, and dead-code elimination. Captures bound the external inputs; recipe operands preserve membership through canonical replacement; dominance proves availability at the candidate; and fact-identity plus hard same-storage relations keep the marker non-consuming without assigning it an independent allocation. Contract collection rejects unsealed candidates, uncaptured inputs, non-dominating producers, incomplete dependencies, and cyclic or stale recipes.

AMDGPU source-to-low boolean analysis follows those transparent fact identities as part of exact-contract lowering. Cross-block durability propagates backward through candidate results, and bidirectional native-mask walks use an active-path guard, preserving transient scalar condition-code values only while every eventual use remains valid in the defining block.

Each successful allocation is evaluated with the same target residency model used by source planning. When physical alignment, fixed target state, or allocation fragmentation crosses a protected tier, the emission frame clones the recorded producer DAG once at each exact leaf-use block, rebuilds scheduling and allocation, and terminates after the finite candidate set. Reconstruction follows dependency readiness rather than original packet order, so independent chains may be interleaved and producers may span dominated CFG blocks. A preserve contract first validates the source model's protected tier; if exact allocation falls below that projection, finite baseline-protecting alternatives restore and measure the exact authored placement before an independent numeric minimum is applied. An impossible minimum or exhausted baseline repair produces a structured backend failure.

Candidate and requirement markers are replaced or erased as the contract is consumed, so placement metadata never reaches emitted machine IR. Terminal allocation and repair evidence remains available to structured reporting without making the allocator depend on source policy.

### Review surface

- `residency_recipes.*` seals the post-lowering SSA dependency boundary.
- `residency_contract.*` validates and consumes markers into finite exact candidates.
- Allocation rematerialization reconstructs dependency-ordered producer DAGs and rewrites only recorded leaf uses.
- Emission-frame construction owns the finite schedule/allocation repair loop and terminal contract state.
- Source-to-low, configured unrolling, bank SROA, and SCF-to-CFG preserve candidate identity without importing source policy into the allocator.


benvanik:

PR #194, opened 2026-07-20
[Loom] Project source pressure into residency tiers

benvanik:
Loom can now project structured source SSA liveness into the same ordered target-residency model used to reason about physical resource allocation. Source transforms no longer need to approximate a scheduling decision with a target-independent value count, and target-low validation does not need a second definition of the cliffs that matter. Both stages consume one immutable model over direct and derived resource vectors.

## Target residency evaluation

The residency layer now provides a prepared evaluator for repeated whole-model tier queries. Initialization validates the immutable target tables once; each subsequent query evaluates a dense direct-resource vector without allocating diagnostic storage or revalidating the model. This is the query shape needed by a pressure sweep or placement search that may evaluate many program points.

Direct resources contribute through their own ordered cliff tables. Derived resources combine one or more direct resources using target-defined allocation granularities before applying shared cliffs, allowing a target to describe coupled pools and aggregate limits without teaching the compiler what those resources mean. The effective residency tier is the most restrictive tier produced by any direct or derived resource. Saturating arithmetic keeps malformed or extreme inputs from wrapping into a falsely favorable result.

The existing diagnostic query remains the richer interface for explaining a selected vector: it identifies limiting resources, the next worse cliff, and the reductions needed to recover a better tier. The prepared evaluator and diagnostic query share the same validated model and cliff semantics, so a fast planning decision and its later explanation cannot drift.

Model absence is explicit. An unavailable model accepts no resource vector and produces tier zero; it is not interpreted as unlimited capacity. Callers can therefore distinguish a target that has no residency contract from one whose resource usage genuinely occupies the best tier.

## Structured source-pressure projection

The new source-pressure analysis maps source SSA values through the selected lowering policy into target-authored direct resource units. Resource identities are the descriptor register-class IDs already used by lowering, so the source projection and physical allocation speak the same target contract without generic code testing source types for AMDGPU-specific classes.

The analysis acquires or reuses the function's region-tree value domain, view-region facts, target query state, and structured liveness. Each mapped liveness segment becomes a pair of deterministic pressure events. Segment ends are applied before starts at the same program point, preventing adjacent non-overlapping lifetimes from being counted together. Leaf-operation results that have no ordinary live interval still receive a one-point issue segment, preserving the transient destination pressure required to execute the operation rather than treating an unused result as free.

Sweeping those events produces two deliberately different forms of evidence:

- Per-resource peaks report the maximum observed units for each direct resource independently.
- The vector captured at the earliest worst-tier program point reports resources that were simultaneously live when residency was actually constrained.

That distinction matters for derived resources. Combining independent SGPR-like and VGPR-like peaks from different branches or program points could invent a shared-pool cliff that no feasible execution reaches. Evaluating the simultaneous vector at every structured program point preserves branch exclusivity while still accounting for join-live values, loop-carried values, values live through nested regions, and transient results. The projection therefore composes alternative paths by their feasible maximum rather than summing mutually exclusive pressure.

Named reserve contributors account for target state that is not represented by source SSA values, such as fixed ABI state or bounded lower-level temporaries. Each reserve carries a stable name and a complete direct-resource vector, is copied into analysis-owned storage, and is applied at every program point. A separate completeness bit records whether those contributors bound all non-source pressure. Consumers can use an incomplete envelope as evidence without mistaking it for a proof that a residency floor will survive physical allocation.

The resulting arena-owned envelope records model availability, projection completeness, the worst tier and its earliest program point, independent resource peaks, the simultaneous worst-tier vector, the aggregate and per-contributor reserves, and counts for mapped values, non-register values, liveness segments, and transient issue segments. Invalid resource IDs, zero-sized register mappings, descriptor/model mismatches, malformed reserve vectors, liveness underflow, and arithmetic overflow fail loudly instead of silently weakening the estimate.

## Target integration and impact

Lowering policies gain one optional callback that selects the immutable residency model for a target contract. AMDGPU implements it by returning the existing occupancy model associated with the selected descriptor set. The generic evaluator and source analysis contain no SGPR, VGPR, AGPR, wave, or occupancy branches; another target can define different resource classes, coupled pools, or throughput tiers behind the same interface.

This gives placement and scheduling transforms a target-aware but target-neutral question: for the values simultaneously live at this source placement, which ordered residency tier does the selected target model predict? The answer can be compared across legal alternatives before unrolling and structural lowering, while final physical allocation can be checked against the same cliff tables afterward. A projection miss is therefore measurable disagreement between conservative source evidence and exact allocation, not disagreement between two independently implemented cost models.

Coverage exercises exact tier boundaries, direct and derived resource interaction, explicit model absence, mutually exclusive branch pressure, simultaneous cross-class pressure, loop-carried and live-through values, transient issue pressure, and named reserves that constrain even an otherwise empty function. Both Bazel and CMake build and test the analysis as a compiler library with visibility restricted to the low pipeline and loop transforms.

### DIFF
--- a/loom/src/loom/codegen/low/lower/BUILD.bazel
+++ b/loom/src/loom/codegen/low/lower/BUILD.bazel
@@ -74,6 +74,7 @@ loom_cc_library(
         "//loom/src/loom/target:low_descriptor_registry",
         "//loom/src/loom/target:low_legality",
         "//loom/src/loom/target:registers",
+        "//loom/src/loom/target:residency",
         "//loom/src/loom/target:types",
         "//loom/src/loom/util:cfg_graph",
         "//loom/src/loom/util:fact_table",
@@ -104,6 +105,42 @@ loom_cc_library(
         "//loom/src/loom/target:types",
         "//runtime/src/iree/base",
         "//runtime/src/iree/base/internal:arena",
+    ],
+)
+
+loom_cc_library(
+    name = "source_pressure",
+    srcs = ["source_pressure.c"],
+    hdrs = ["source_pressure.h"],
+    visibility = [
+        "//loom/src/loom/codegen/low/transforms/pipeline:__pkg__",
+        "//loom/src/loom/transforms/loop:__pkg__",
+    ],
+    deps = [
+        ":lower",
+        "//loom/src/loom/analysis:liveness",
+        "//loom/src/loom/analysis:view_regions",
+        "//loom/src/loom/ir",
+        "//loom/src/loom/util:adaptive_sort",
+        "//runtime/src/iree/base",
+        "//runtime/src/iree/base/internal:arena",
+    ],
+)
+
+loom_cc_test(
+    name = "source_pressure_test",
+    srcs = ["source_pressure_test.cc"],
+    deps = [
+        ":source_pressure",
+        "//loom/src/loom/format/text:parser",
+        "//loom/src/loom/ir",
+        "//loom/src/loom/ops/func",
+        "//loom/src/loom/ops/scalar",
+        "//loom/src/loom/ops/scf",
+        "//loom/src/loom/testing:module_ptr",
+        "//runtime/src/iree/base/internal:arena",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
     ],
 )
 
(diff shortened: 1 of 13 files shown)


benvanik:
CI triage: the gfx942 UBSAN target exhausted its aggregate 120-second deadline after selected PM4/native-kernarg tests became pathologically slow (27.9s and 60.7s); the log reached `Pm4DynamicDispatchUsesBindingTableSlots`, rather than establishing a deadlock there. The exact July 20 nightly HSA runtime completes the whole 22-test binary on the NOD gfx942 in 3.3s, and downstream PR #195 contains this commit and passed its fresh gfx942 sanitizer lane.

PR #195, opened 2026-07-20
[Loom/AMDGPU] Report and qualify residency planning

benvanik:
Expose residency planning and exact physical outcomes through a versioned compile-report schema, then qualify the complete production pipeline against distinct AMDGPU target models and a configured gfx1250 fragment-bank kernel. Source decisions, direct and derived cliff witnesses, named reserves, stable candidate identities, exact resources, projection misses, repairs, and terminal shortfalls are now observable without parsing diagnostics.

One targetless preserve-policy kernel compiles through the normal pipeline for gfx942, gfx1100, and gfx1250. The resulting artifacts demonstrate target-specific resource domains from identical source: gfx942 reports AGPR and derived VGPR+AGPR pressure with a 64-lane subgroup and 62% modeled occupancy, while gfx1100 and gfx1250 resolve through different resource sets at tier 16 and gfx1250 additionally reports its mode resource.

The configured gfx1250 8x4x2 fragment-bank kernel exercises the optimistic-projection path with real WMMA lowering. Source planning projects tier 16; provisional physical allocation reaches tier 3 at 288 VGPRs and 18% modeled occupancy; one finite restoration recovers the exact authored baseline; and the terminal report records zero shortfall plus the 32-VGPR reduction needed to reach tier 4. Both qualification paths emit non-empty HSACO artifacts, proving that reporting observes shipping compilation rather than a test-only analyzer.

### Review surface

- Summary reports retain stable aggregate counts while detail reports emit typed source, candidate, repair, and exact-resource rows.
- Planning history and terminal exact state use separate fields so a successful repair cannot hide the provisional allocation that triggered it.
- The compile-tool manifest checks production pipeline selection, cross-target model differences, finite repair, terminal satisfaction, occupancy, and emitted HSACO artifacts.

### DIFF
--- a/loom/src/loom/target/reporting/format.h
+++ b/loom/src/loom/target/reporting/format.h
@@ -17,6 +17,9 @@
 extern "C" {
 #endif
 
+// Version of the structured compile-report JSON schema.
+#define LOOM_TARGET_COMPILE_REPORT_JSON_SCHEMA_VERSION 1u
+
 typedef enum loom_target_compile_report_format_mode_e {
   // Does not format a compile report.
   LOOM_TARGET_COMPILE_REPORT_FORMAT_MODE_NONE = 0,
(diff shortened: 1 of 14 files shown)


benvanik:

PR #210, opened 2026-07-28
[HIP] Time event records on the device timeline

zjgarvey:
hipEventElapsedTime now reports the device work between two records rather than the cost of issuing them. A timing-enabled record on a device advertising a timestamp domain enqueues a device timestamp instead of a bare barrier, and elapsed time differences the two ticks with the device's own frequency and counter width. A submitted record also ends the event's association with any capture, so query, synchronize and a stream wait act on the record rather than on the capture it displaced.

## Motivation

Elapsed time differenced two host timestamps sampled when the records were issued, so it measured the two hipEventRecord calls rather than the device work between them: a pair bracketing hundreds of milliseconds of queued work reported a fraction of a millisecond. An event also named its capture forever, so a later record left it readable as both recorded and captured.

## Two commits at the base repair main, not this feature

`iree_hal_streaming_context_destroy` detached every stream while holding the context's stream list mutex, taking each stream's mutex underneath it. Ending a capture takes those two in the other order. Two threads reaching those regions at once deadlock, and ThreadSanitizer reports the cycle the first time one process walks both orders — which the host-side graph executable tests added here are the first thing in the tree to do, so this lands red without the fix. Emptying the list under its own mutex is what takes the streams out of reach; the detach itself needs only the stream mutex, so releasing the list mutex first removes the inversion and weakens nothing.

`cuStreamWaitEvent` was left calling `iree_hal_streaming_stream_wait_event` at its previous arity when that function grew a parameter, so `libhrx/src/binding/cuda/driver.c` does not compile. No Bazel target names the file and its CMake directory is gated on `LIBHRX_BUILD_CUDA_BINDING`, which defaults off, so no build reports it. The binding now forwards the flag CUDA already defines for that question, and validates it the way hipStreamWaitEvent does.

Both are separable and sit at the base of the stack so every commit above them compiles.

## What it costs

Calls that previously succeeded are now refused. cuEventRecord returns CUDA_ERROR_INVALID_HANDLE on a stream of a context other than the event's, since the tick slot comes from that context; hipEventRecord already refuses that pair, and stating the rule at the record itself is what holds every other way of seating one to it. A record made while the stream is capturing names a dependency frontier and no queue point, so it draws no slot and the rule does not reach it. hipGraphLaunch settles the same question once for a whole executable holding any event record, its own or a child graph's, returning hipErrorInvalidValue before submitting any of the graph. hipGraphExecEventRecordNodeSetEvent and hipGraphExecEventWaitNodeSetEvent refuse an event of another context, and hipGraphExecChildGraphNodeSetParams now runs the template builder's child-graph validation, refusing a destroyed handle, a child of another context, and containment reaching back to the parent. hipEventElapsedTime returns hipErrorNotSupported where the device advertises no domain, with no host-time fallback, and hipErrorCapturedEvent for a captured pair; cuEventElapsedTime also rejects null handles and a cross-context pair.

A timed record is real queue work on the one hardware queue serving every stream, so a record issued behind a blocked host callback stays outstanding until that callback returns. A record that cannot obtain a slot from its context's pool fails rather than going silently untimed.

## Verifying

Catching a reversion means pinning the ratio, not the sign: bracket a long device-side occupancy between two records, time the host window around enqueue and around completion separately, then assert the reported milliseconds are a substantial fraction of that window and fit inside it. Host timestamps satisfy any check that asks only for a positive, ordered value.

The new binding/common tests run on the CPU device and carry no resource label, so the Bazel libhrx lane and the CMake CPU lanes run them; the HIP conformance tests are tagged runtime-resource=amd-gpu and run only on GPU lanes.

Three invariants are pinned by tests demonstrated to fail without them: the propagation of an executable's event-record flag out of a child graph, which is what lets a launch refuse a cross-context replay before submitting any of it; the memory type and usage the tick slab requests, asserted against the request rather than the resulting buffer because the heap allocator rewrites both; and the teardown lock order, which ThreadSanitizer reports as an inversion when reverted.


AWoloszyn (approved):

zjgarvey:
Rebase notes: Landing 380 required a bit of work in the rebase, but I think the profile is minimal enough to land. 

1. adjusted iree_hal_streaming_context_destroy vs. iree_hal_streaming_end_capture mutex ordering due to a newly introduced conflict (potential deadlock).
2. Some rewriting of event api around the handle registry work from 380.

zjgarvey:
Landing as the 942 stall seems unrelated.

PR #211, opened 2026-07-28
[HIP] Anchor event records to the timeline that carries them

zjgarvey:
An event no longer owns a timeline. A record names a point on the timeline of whichever submission carries it: the recording stream's own timeline for a stream record, and for a graph record node either the launching stream's timeline or one of the executable's internal partition semaphores. `iree_hal_streaming_recorded_point_t` in `libhrx/src/binding/common/internal.h` carries that point as one value: the semaphore, the value it reaches, and the stream timeline point reaching it is known to follow. The event retains the semaphore and swaps the whole struct under its own mutex. Readers copy the point and retain the semaphore before waiting, so no reader pairs one record's semaphore with another record's value and no re-record frees a timeline a waiter still names.

That last part is the ordering claim `hipStreamWaitEvent` files in the waiting stream's memory reuse ledger: exact for a stream record and for a record ending a launch, and a lower bound (the stream tail the launch waited behind) for a record internal to a launch. Every remaining gap errs one way: a missing or understated claim withholds an allocation from reuse and can never grant it.

## Motivation

The event's private semaphore was driven by values read off other timelines, so re-recording an event on a stream whose timeline sat behind it targeted a value already passed. `hipEventQuery` reported the new record complete before its work had run, `hipEventSynchronize` returned immediately, and a `hipStreamWaitEvent` on it was satisfied at submission, a cross-stream dependency dropped with no error anywhere. The reuse ledger filed that same counter paired with whichever stream last recorded the event, so a graph record could hand a waiting stream an allocation the producer was still using.

Repairs in the same paths: one helper reserves stream timeline values and owns the overflow check; a stream event wait also waits on the stream's own tail; a graph executable's resource set solely owns its internal semaphores, fixing a leak and keeping a record queryable after the executable is gone; semaphore bases advance past a block that failed to submit, which used to re-signal values and fail the executable permanently; stream query and synchronize propagate a failed timeline rather than reading it as incomplete.

## Tests

`libhrx/cts/tests/hip/event_test.cpp` drives the built `amdhip64` shim through public HIP entry points only. A host callback parks a stream until the test releases it, so a test can observe an event while its record is provably outstanding: the cross-stream re-record case requires `hipErrorNotReady` where the old counter read complete, and six reuse cases pin the grant and deny sides in pairs differing by one call, so neither blanket policy survives. A record that chains streams back together shows up as a hang, not a failed assertion. The concurrency case is a manual race-detector probe: unaided it pins only that every call succeeds and every stream drains.

Worth a close read: the claim in `graph_exec.c` that a record block's first signal value is the point the record names, and the lock order, with the event mutex innermost.

### DIFF
--- a/libhrx/cts/BUILD.bazel
+++ b/libhrx/cts/BUILD.bazel
@@ -143,6 +143,24 @@ hrx_cts_test(
 )
 
 hrx_cc_test(
+    name = "hip_event_test",
+    srcs = ["tests/hip/event_test.cpp"],
+    data = ["//libhrx/src/binding/hip:amdhip64"],
+    env = {
+        "HRX_TEST_LIBAMDHIP64": "$(location //libhrx/src/binding/hip:amdhip64)",
+    },
+    sanitizer_suppressions = hsa_suppressions,
+    tags = AMD_GPU_CTS_TAGS,
+    target_compatible_with = _LINUX_COMPATIBILITY,
+    deps = [
+        "//build_tools:dl",
+        "//libhrx/src/binding/hip:launch_params",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
     name = "hip_module_execution_test",
     srcs = [
         "core/amdgpu_executable_test_data.hpp",
--- a/libhrx/cts/CMakeLists.txt
+++ b/libhrx/cts/CMakeLists.txt
@@ -260,6 +260,30 @@ hrx_cc_test(
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 iree_cc_test(
   NAME
+    hip_event_test
+  SRCS
+    "tests/hip/event_test.cpp"
+  DATA
+    libhrx::src::binding::hip::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx::src::binding::hip::launch_params
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+  LABELS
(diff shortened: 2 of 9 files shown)


AWoloszyn (approved):

zjgarvey:
@AWoloszyn This is rebased and a few semantic fixes needed to be applied, so I don't want to land without a re-review. 

If you think the event tests overlap too much with the external hip tests, let me know and I can cut them out.

AWoloszyn:
I am happy with the tests, but  my agent caught a couple of issues.

  1. High: rejected command-buffer submission leaves an unsignaled stream tail.
     /tmp/hrx-pr211-review/libhrx/src/binding/common/stream.c:481 publishes pending_value before /tmp/hrx-pr211-review/libhrx/src/binding/common/stream.c:500 accepts the submission. If execution is rejected, such as during resource exhaustion, later operations wait on a timeline value nothing will signal and stream synchronization can hang. This directly violates the reservation contract documented in /tmp/hrx-pr211-review/libhrx/src/binding/common/internal.h:550. pending_value should advance after queue_execute accepts the submission. A subsequent queue_flush failure must be handled separately because the submission may already own that timeline value.

  2. Medium: elapsed-time checks timestamps and completion from different records.
     /tmp/hrx-pr211-review/libhrx/src/binding/common/event.c:298 snapshots each timestamp and then independently calls /tmp/hrx-pr211-review/libhrx/src/binding/common/event.c:311, which takes another snapshot of the recorded point. A concurrent re-record can therefore pair one record’s timestamp with another record’s semaphore/value. This is precisely the state the mutex comment says must remain atomic in /tmp/hrx-pr211-review/libhrx/src/binding/common/internal.h:762.

     The event needs one acquisition helper returning both the retained recorded point and its corresponding timestamp. Elapsed-time should query those exact retained points. The redundant API-level timestamp precheck in /tmp/hrx-pr211-review/libhrx/src/binding/hip/api.c:11258 can then be removed.

AWoloszyn (approved):

PR #231, opened 2026-07-30
[hip] Implement virtual memory APIs.

AWoloszyn:
## Summary

  This change implements stream-ordered HIP memory allocation on top of virtual-memory-backed TLSF pools. It separates the system into three layers:

  - The AMDGPU HAL implements the existing generic virtual-memory contract.
  - HRX owns VMM slab construction, pool policy, accounting, and lifetimes.
  - The HIP binding implements stream ordering, allocation reuse, pool attributes,
    and API behavior.

  Asynchronous allocations now receive stable device addresses backed by
  page-aligned virtual reservations. Physical backing is allocated, mapped, and
  protected through the HAL, then subdivided by TLSF in HRX. Freed ranges remain
  ordered against stream work and can be reused without attempting to discover
  device pointers embedded inside kernel arguments or device memory.

  ## AMDGPU Virtual Memory

  The AMDGPU allocator now implements the HAL virtual-memory operations:

  - Query allocation granularity.
  - Reserve and release virtual address ranges.
  - Allocate and free physical memory handles.
  - Map and unmap physical memory.
  - Apply queue-specific access protection.
  - Process supported virtual-memory advice.

  The implementation tracks reservations, mappings, and physical allocations in
  allocator-local state. It validates ownership, granularity, alignment, range
  bounds, mapping state, queue affinity, and release ordering.

  ROCr requires each mapping operation to cover a complete physical allocation.
  The implementation therefore rejects partial physical mappings instead of
  claiming support that the underlying interface cannot provide. VMM is also
  rejected when address sanitization or required allocation flags cannot be
  represented correctly.

  ## HRX Pool Architecture

  HRX provides a VMM slab provider that:

  1. Reserves a virtual address range.
  2. Allocates matching physical backing.
  3. Maps and protects the complete range.
  4. Exposes the resulting device address as a TLSF slab.
  5. Tears resources down in dependency order: unmap, free physical memory, then
     release the virtual reservation.

  Suballocations are represented as HAL buffer subspans. A generic callback-capable
  subspan constructor ties pool bookkeeping to the final buffer reference, so a
  TLSF range cannot be reused while any materialized view still retains it.

  Pool policy remains in HRX rather than AMDGPU. This includes:

  - Per-device default pools and explicitly created pools.
  - Logical allocation accounting and maximum-size enforcement.
  - Physical slab retention and trimming.
  - Oversized allocation routing.
  - Release thresholds expressed as byte counts.
  - Reuse policy attributes.
  - Pool access and location validation.

  The generic HAL additions are limited to the uncached-memory requirement,
  callback-capable subspans, and trimming a TLSF pool toward a byte threshold.

  ## Stream-Ordered Allocation

  The HIP binding now implements the primary stream-ordered allocation APIs,
  including:

  - hipMallocAsync
  - hipMallocFromPoolAsync
  - hipFreeAsync
  - Default and current device memory-pool selection
  - Memory-pool creation, destruction, attributes, access, and trimming

  An asynchronous free removes the allocation from the public buffer table before
  enqueueing its retirement operation. Capacity for rollback is reserved first,
  so an enqueue failure can restore ownership without performing another
  allocation or losing the buffer.

  Completed frees follow the configured reuse policy:

  - Opportunistic reuse returns completed allocations to TLSF immediately.
  - Conservative reuse retains completed allocations until an explicit stream
    synchronization, pool trim, or valid dependency-based reuse point.

  - Same-stream ordering and recorded cross-stream event dependencies allow reuse
    without global synchronization.

  Stream synchronization snapshots its target timeline value under the stream
  lock and only waits for work included in that snapshot. Completed asynchronous
  frees are retired after the corresponding timeline point is reached.

  ## Transfers and Graphs

  Host-visible and pageable transfers preserve stream ordering through registered
  buffers, owned staging allocations, and stream host calls. The direct HAL
  transfer path remains a serialized fallback for cases that cannot be represented
  as queued buffer operations. It validates the entire source or destination range
  before issuing any transfer chunks, preventing partial updates from an invalid
  request.

  Graph-owned staging allocations use the same queue-visible host-memory path.
  Graph dependency barriers now include both dispatch and transfer read/write
  scopes, preserving memory visibility when graph nodes alternate between kernels
  and copies.

  ## Error Handling and Cleanup

  The change also tightens failure behavior throughout the memory stack:

  - Buffer-table pointer and range checks avoid integer overflow.
  - Buffer-table growth checks capacity and allocation-size overflow.
  - Failed asynchronous-free enqueue operations restore ownership without an
    allocation.

  - VMM teardown preserves and joins cleanup failures.
  - Mapped physical resources remain alive when unmapping fails.
  - Pool trimming propagates errors and honors the configured byte threshold.
  - Synchronous frees verify that the allocation was removed successfully.
  - Unsupported VMM placements fail explicitly rather than silently changing
    memory properties.

  ## Verification

  The following verification passes on the final stack:

  - All 32 buffer-table tests.
  - All 20 TLSF pool tests.
  - Branch-specific regression coverage for allocation-free async-free rollback.
  - Branch-specific regression coverage for byte-valued TLSF release thresholds.
  - Repository pre-commit runtime and libhrx tests.
  - Formatting, build metadata validation, text hygiene, and Semgrep checks.

### DIFF
--- a/libhrx/include/hrx_runtime.h
+++ b/libhrx/include/hrx_runtime.h
@@ -1025,9 +1025,9 @@ HRX_API hrx_status_t hrx_mem_pool_set_attribute(hrx_mem_pool_t pool,
 HRX_API hrx_status_t hrx_mem_pool_trim(hrx_mem_pool_t pool,
                                        size_t min_bytes_to_keep);
 
-// Releases a fully idle pool's backing storage when its release threshold is
-// zero. This is intended for stream-ordered free completion paths.
-HRX_API void hrx_mem_pool_release_unused(hrx_mem_pool_t pool);
+// Trims unused backing storage toward the pool's configured release threshold.
+// This is intended for stream-ordered free completion paths.
+HRX_API hrx_status_t hrx_mem_pool_release_unused(hrx_mem_pool_t pool);
 
 // Records a logical allocation backed by |pool|. Logical usage is distinct
 // from physical TLSF reservations, which can remain retained for later
--- a/libhrx/include/hrx_runtime.h
+++ b/libhrx/include/hrx_runtime.h
@@ -1001,9 +1001,14 @@ typedef enum hrx_mem_pool_attr_t {
 } hrx_mem_pool_attr_t;
 
 typedef struct hrx_mem_pool_props_t {
+  // Platform allocation-handle types requested when the pool was created.
   uint32_t alloc_handle_type;
+  // Platform location type for the pool's backing storage.
   uint32_t location_type;
+  // Platform device ordinal for the pool's backing storage.
   int location_id;
+  // Maximum live reservation bytes, or zero when the pool is unbounded.
+  size_t max_size;
 } hrx_mem_pool_props_t;
 
 HRX_API hrx_status_t hrx_mem_pool_create(hrx_device_t device,
@@ -1019,6 +1024,21 @@ HRX_API hrx_status_t hrx_mem_pool_set_attribute(hrx_mem_pool_t pool,
                                                 uint64_t value);
 HRX_API hrx_status_t hrx_mem_pool_trim(hrx_mem_pool_t pool,
                                        size_t min_bytes_to_keep);
+
+// Releases a fully idle pool's backing storage when its release threshold is
+// zero. This is intended for stream-ordered free completion paths.
+HRX_API void hrx_mem_pool_release_unused(hrx_mem_pool_t pool);
+
+// Records a logical allocation backed by |pool|. Logical usage is distinct
+// from physical TLSF reservations, which can remain retained for later
+// stream-ordered reuse.
+HRX_API void hrx_mem_pool_record_logical_allocation(hrx_mem_pool_t pool,
+                                                    size_t size);
+
+// Records retirement of a logical allocation backed by |pool|. The physical
+// reservation remains owned by its buffer until that buffer is released or
+// reused.
+HRX_API void hrx_mem_pool_record_logical_free(hrx_mem_pool_t pool, size_t size);
 
 // Allocates a buffer from |pool| using the same parameter contract as
 // hrx_allocator_allocate_buffer. The returned buffer owns the allocation and
--- a/libhrx/src/binding/common/BUILD.bazel
+++ b/libhrx/src/binding/common/BUILD.bazel
@@ -22,6 +22,7 @@ hrx_cc_library(
     srcs = [
         "context.c",
         "device.c",
+        "direct_transfer.c",
         "event.c",
         "fat_binary.c",
         "graph.c",
@@ -39,12 +40,15 @@ hrx_cc_library(
         "tls.c",
     ],
     hdrs = [
+        "direct_transfer.h",
         "fat_binary.h",
         "graph.h",
         "hrx_bridge.h",
         "internal.h",
         "kernel_arguments.h",
         "kpack.h",
+        "memory.h",
+        "stream.h",
         "tls.h",
     ],
     includes = [
--- a/libhrx/src/binding/common/CMakeLists.txt
+++ b/libhrx/src/binding/common/CMakeLists.txt
@@ -8,16 +8,20 @@ iree_cc_library(
   NAME
     common
   HDRS
+    "direct_transfer.h"
     "fat_binary.h"
     "graph.h"
     "hrx_bridge.h"
     "internal.h"
     "kernel_arguments.h"
     "kpack.h"
+    "memory.h"
+    "stream.h"
     "tls.h"
   SRCS
     "context.c"
     "device.c"
+    "direct_transfer.c"
     "event.c"
     "fat_binary.c"
     "graph.c"
--- a/libhrx/src/binding/common/buffer_table_test.cc
+++ b/libhrx/src/binding/common/buffer_table_test.cc
@@ -71,6 +71,14 @@ static iree_status_t iree_hal_streaming_buffer_table_insert(
                               buffer->size, (hrx_buffer_t)buffer, nullptr));
 }
 
+static iree_status_t iree_hal_streaming_buffer_table_insert_reserved(
+    iree_hal_streaming_buffer_table_t* table,
+    iree_hal_streaming_buffer_t* buffer) {
+  return BufferTableStatus(hrx_buffer_table_insert_reserved(
+      table, buffer->device_ptr, buffer->host_ptr, buffer->size,
+      (hrx_buffer_t)buffer, nullptr));
+}
+
 static iree_status_t iree_hal_streaming_buffer_table_remove(
     iree_hal_streaming_buffer_table_t* table, uint64_t any_ptr) {
   return BufferTableStatus(hrx_buffer_table_remove(table, any_ptr));
@@ -641,6 +649,47 @@ TEST(BufferTableTest, InsertRemoveInsert) {
   iree_hal_streaming_buffer_table_free(table);
   FreeDummyBuffer(buffer1, allocator);
   FreeDummyBuffer(buffer2, allocator);
+}
+
+TEST(BufferTableTest, ReservedInsertSurvivesCapacityPressure) {
+  iree_allocator_t allocator = iree_allocator_system();
+  iree_hal_streaming_buffer_table_t* table = nullptr;
+  IREE_ASSERT_OK(iree_hal_streaming_buffer_table_allocate(allocator, &table));
+
+  auto* removed_buffer = CreateDummyBuffer(0x100000000ULL, 4096, allocator);
+  IREE_ASSERT_OK(iree_hal_streaming_buffer_table_insert(table, removed_buffer));
+  const size_t reserved_capacity = table->capacity;
+  IREE_ASSERT_OK(BufferTableStatus(hrx_buffer_table_reserve_insert(table)));
+  IREE_ASSERT_OK(iree_hal_streaming_buffer_table_remove(
+      table, removed_buffer->device_ptr));
+
+  std::vector<iree_hal_streaming_buffer_t*> competing_buffers;
+  competing_buffers.reserve(reserved_capacity - 1);
+  for (size_t i = 0; i < reserved_capacity - 1; ++i) {
+    auto* buffer =
+        CreateDummyBuffer(0x200000000ULL + i * 0x10000, 4096, allocator);
+    competing_buffers.push_back(buffer);
+    IREE_ASSERT_OK(iree_hal_streaming_buffer_table_insert(table, buffer));
+  }
+  EXPECT_EQ(table->capacity, reserved_capacity);
+  EXPECT_EQ(table->reserved_insert_count, 1u);
+
+  IREE_EXPECT_OK(
+      iree_hal_streaming_buffer_table_insert_reserved(table, removed_buffer));
+  EXPECT_EQ(table->capacity, reserved_capacity);
+  EXPECT_EQ(table->count, reserved_capacity);
+  EXPECT_EQ(table->reserved_insert_count, 0u);
+
+  iree_hal_streaming_buffer_t* found = nullptr;
+  IREE_EXPECT_OK(iree_hal_streaming_buffer_table_lookup(
+      table, removed_buffer->device_ptr, &found));
+  EXPECT_EQ(found, removed_buffer);
+
+  iree_hal_streaming_buffer_table_free(table);
+  FreeDummyBuffer(removed_buffer, allocator);
+  for (auto* buffer : competing_buffers) {
+    FreeDummyBuffer(buffer, allocator);
+  }
 }
 
 TEST(BufferTableTest, MixedOperations) {
--- a/libhrx/src/binding/common/context.c
+++ b/libhrx/src/binding/common/context.c
@@ -78,6 +78,7 @@ iree_status_t iree_hal_streaming_context_create(
   context->peer_capacity = 0;
   memset(&context->symbol_map, 0, sizeof(context->symbol_map));
   memset(&context->buffer_table, 0, sizeof(context->buffer_table));
+  context->pending_free_head = NULL;
   context->pageable_h2d_staging_buffer = NULL;
   context->pageable_h2d_staging_size = 0;
   iree_atomic_store(&context->capture_stream_count, 0,
@@ -85,6 +86,7 @@ iree_status_t iree_hal_streaming_context_create(
   context->host_allocator = host_allocator;
   iree_slim_mutex_initialize(&context->mutex);
   iree_slim_mutex_initialize(&context->direct_transfer_mutex);
+  iree_slim_mutex_initialize(&context->pending_free_mutex);
 
   // Initialize global list pointers.
   context->context_list_entry.next = NULL;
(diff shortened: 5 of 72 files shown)


benvanik:
(review in-flight - we've locally proven the HAL design to be wrong, but won't block this on the larger reworking - may suggest some simplifications/adjustments to limit surface area that is expected to be rewritten, though - nothing much should change in hrx/hip: "The implementation confirms the worst architectural suspicion in literal form: one allocator-wide mutex, a linear reservation list, a linear mapping list, a mapped-byte counter on every physical handle, and native ROCr calls made while holding the global lock.")

benvanik (changes requested):
Thank you for pushing this through as an end-to-end change. This PR gives HAL virtual memory its first demanding HRX/HIP consumer, and that concrete use has exposed ownership and asynchronous-lifetime contracts that were previously much harder to test. We reviewed the runtime and AMDGPU portions deeply at `7e235473f037bdd1052ecd382babdc752f1a7662`, using the HRX/HIP code as the production caller rather than trying to review the entire compatibility surface at once.

The feature direction is viable and worth landing. We did find several correctness issues that cannot land as written, most of them at asynchronous ownership boundaries. The encouraging part is that the runtime corrections simplify the implementation, remove redundant state, and add focused coverage rather than growing a second VMM subsystem.

## Review outcome

This needs changes before merge. The current code can physically release a TLSF slab while queued device work still owns it, lose VMM handles after teardown failure, leak accepted host operations when a prerequisite fails, and advertise an uncached-memory property from allocators that cannot provide it. These are concrete ownership faults rather than style preferences.

There is a bounded path to completion. We published a [correction branch](https://github.com/ROCm/hrx-system/commits/users/benvanik/pr231-runtime-review/) on top of the exact reviewed head so the proposed mechanisms are executable rather than prose-only. It is intended as an implementation handoff: cherry-pick it, adapt it, or reproduce the same invariants in the PR branch.

## Corrections already implemented

1. **Gate TLSF slab release on its death frontier.** [`6296b494`](https://github.com/ROCm/hrx-system/commit/6296b494bde1002f3717fae4cc16caf3debaeb8b) fixes the P0 in [`iree_hal_tlsf_pool_trim_unused_slabs_locked`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/runtime/src/iree/hal/memory/tlsf_pool.c#L1153-L1215). A logically free slab may still carry the causal frontier for queued GPU accesses; `allocation_count == 0` is therefore insufficient proof that its physical storage is dead. The correction queries the coalesced full-range block and reuses the pool's existing nonblocking epoch predicate. It adds no persistent state or host wait: incomplete, tainted, or unqueryable frontiers retain the slab, and a completed frontier permits immediate release. Focused coverage exercises empty, incomplete, completed, and tainted frontiers.

2. **Make ROCr authoritative for AMDGPU VMM ownership.** [`672c1a2f`](https://github.com/ROCm/hrx-system/commit/672c1a2f095f1dbba92c2a00d1170f80afc4b637) removes the allocator-global reservation and mapping registries introduced in [`virtual_memory.c`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/runtime/src/iree/hal/drivers/amdgpu/virtual_memory.c#L24-L203). The submitted implementation duplicates ROCr's reservation, mapping, overlap, and handle-reference state under a second global mutex, with linear scans and a host allocation for every mapping. ROCr already validates those transitions under its own memory lock, and the submitted HRX slab caller passes the exact reservation and physical handles to every operation. The correction keeps ownership on those objects and calls ROCr directly. It also moves native address release into the public status-bearing operation, so the [void buffer callback](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/runtime/src/iree/hal/drivers/amdgpu/virtual_memory.c#L224-L238) can no longer report success after native release failed and destroyed the retry state. The production implementation becomes smaller while preserving VMM support.

   The same commit keeps native minimum and recommended allocation granularities distinct, records physical-handle allocator statistics, and adds hardware-backed lifecycle coverage. That coverage includes mapped-reservation release failure followed by retry, invalid transition rejection by ROCr, GPU access through a mapping, physical allocation statistics, and a legal minimum-granule host allocation mapped into a recommendation-aligned device reservation.

3. **Give queued host-operation state a terminal owner.** [`e7bb644e`](https://github.com/ROCm/hrx-system/commit/e7bb644e806c418a6e0323788c24960af9db3c29) adds an optional HAL resource retain to a host call, and [`1587cdbb`](https://github.com/ROCm/hrx-system/commit/1587cdbbd1037caf3d12ea8ac66186e869710e09) applies it to the PR's staged D2H operation. Every reviewed backend correctly skips the user callback when a wait prerequisite fails, but the submitted D2H path gives its staging buffer and descriptor exclusively to that callback. Once the queue accepts the operation, a prior stream failure therefore leaks both allocations. The correction separates the success effect from cleanup: the callback scatters bytes only when it runs, while the resource destructor reclaims the state after either callback return or terminal cancellation. CTS proves delayed invocation and failed-prerequisite cleanup without a timeout or queue drain.

4. **Close the HIP pool lifetime and zero-size gaps.** [`ccb43062`](https://github.com/ROCm/hrx-system/commit/ccb4306262c01e29f433035b24353fe14f84b2ff) releases the retained current pool after [`hipGraphAddMemAllocNode`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L18191-L18207) materializes its allocation, and makes [`hipMallocFromPoolAsync(size=0)`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L22755-L22820) return success with `NULL`, matching the canonical HIP behavior and the neighboring default-pool entry point. Dynamic-ABI coverage catches the pool leak without suppressing leaks originating in the shim under test.

## Remaining changes required

1. **Replace the process-global deferred-free state machine with terminally owned, context-scoped pending operations.** The current [`iree_hal_streaming_deferred_device_free_t` registry](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/common/memory.c#L616-L775) puts every pending free in one process-global mutex/list and scans that list on each pool allocation. Its cost is O(process-wide pending frees), even though reuse is meaningful only within the owning context, pool, size class, and dependency relation.

   There are two correctness holes in addition to that scaling pressure. With opportunistic reuse disabled, the host callback can mark an operation ready and return while leaving it registered; a later allocation removes and reuses the buffer but has no remaining actor that destroys the free record or releases its retained context. Conversely, if an earlier stream operation fails, the callback is skipped and the record remains permanently not-ready while retaining the detached buffer and context. The [enqueue and publication path](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/common/memory.c#L1049-L1233) needs exactly one terminal owner covering callback execution, cancellation, reuse, and final pool return.

   The host-call resource contract above supplies cancellation ownership, but the reuse policy still needs a coherent state model. Keep pending reuse state within the owning context/pool rather than a process-global scan. Once completion occurs, either return the wrapper to the HAL pool or place it in an explicitly defined completed quarantine with a bounded drainage path. `ReuseAllowOpportunistic=false` must have a precise observable contract; if the intended HIP semantics cannot be implemented without retaining completed wrappers indefinitely, report that policy unsupported rather than leaving operations in ownership limbo.

2. **Preserve VMM slab ownership through teardown failure.** [`hrx_vmm_slab_provider_release_slab_state`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/libhrx/vmm_slab_provider.c#L66-L103) correctly avoids freeing physical memory or releasing the reservation after unmap fails, but then unconditionally frees the only object holding those handles. The provider's [void release callback](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/libhrx/vmm_slab_provider.c#L181-L193) prints and discards the resulting status after retry ownership is already gone. Acquisition rollback has the same handle-loss mechanism even though it joins the cleanup status into the primary failure.

   The teardown contract needs to be internally consistent. A fallible path must keep an owner until every handle is consumed and permit the caller to retry or escalate the failure. A terminal void callback must use an infallible/raw cleanup contract whose success is an invariant; it cannot construct a status, discard it, and free the retry state. In either form, slab metadata cannot be freed while it still contains a live mapping, physical handle, or virtual reservation.

3. **Report the actual committed VMM slab length.** The provider rounds `min_length` to `allocation_size`, then reserves, allocates, and maps `allocation_size`, but [returns `min_length` as the slab length](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/libhrx/vmm_slab_provider.c#L117-L175). Pool accounting charges the returned length, so non-page-aligned requests underreport physically committed memory. Return `allocation_size`.

4. **Make `DEVICE_UNCACHED` truthful across the HAL.** The new public bit explicitly says [unsupported allocators must reject it](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/runtime/src/iree/hal/buffer.h#L98-L101), but only AMDGPU recognizes the requirement. The [heap compatibility path](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/runtime/src/iree/hal/allocator_heap.c#L134-L174), along with the CUDA, HIP, and WebGPU compatibility allocators, currently preserves the unknown required bit while still reporting the allocation as supported. A direct heap-allocator probe received ordinary cached heap storage whose reported type still included `DEVICE_UNCACHED`. Update every allocator to reject the unsupported requirement, or keep this property out of the generic HAL memory-type surface until that contract can be enforced.

5. **Resolve the four new HIP lifecycle clang-tidy failures rather than suppressing them.** The affected declarations are [`iree_hip_mem_pool_retain`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L328-L353), [`handle_ref_count`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L14405-L14420), [`iree_hip_user_object_handle_release`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L14447-L14461), and [`iree_hip_user_object_registry_retain`](https://github.com/ROCm/hrx-system/blob/7e235473f037bdd1052ecd382babdc752f1a7662/libhrx/src/binding/hip/api.c#L14463-L14479). The registry helpers perform fallible lookup/validation and should be named as such rather than as conventional retain operations; the handle counter does not own the enclosing object's lifetime and should use an ordinary atomic integer; and release should follow the null-safe lifecycle shape expected by the runtime checks.

## Landing shape

The strongest path is to transplant the correction stack first, then repair the deferred-free and slab-provider ownership models in the PR branch, make the public memory-type requirement truthful, and clear the lifecycle checks. At that point the runtime portion has a compact ownership story: HAL owns accepted host-operation state through every terminal path, ROCr owns native VMM transition validation, per-object wrappers own the handles needed by their callers, and TLSF releases physical storage only after both logical release and device completion.

That is a good foundation for the product feature this PR is trying to unlock. The review is large because asynchronous memory ownership is unforgiving, not because the contribution is headed in the wrong direction. This PR supplied the concrete workload that let us find and test these contracts, and finishing the corrections will leave both HRX and the shared runtime substantially stronger.

One separate queue semaphore conversion defect surfaced during the review, but history shows it was already present at the merge base. We have excluded it from this PR's requested changes and will handle it independently on main.

AWoloszyn:
@benvanik PTAL again. That should cover the issues presented.

benvanik (changes requested):
Thank you for the substantial refresh. I re-reviewed exact head [`5ced8a4e`](https://github.com/ROCm/hrx-system/commit/5ced8a4e96fe6977ea11600dd9d67aa86e5464a7), focusing on generic HAL and AMDGPU while tracing HRX as the production caller.

The hard structural work from the previous review is now addressed. AMDGPU delegates mapping authority to ROCr without a duplicate registry, TLSF trimming respects death frontiers, accepted host calls have terminal ownership, deferred frees are context-local, VMM slab teardown preserves retry state, and the accounting and lifecycle issues are corrected.

## Review outcome

Two bounded changes remain before this can land. Both are implemented on the [two-commit correction branch](https://github.com/ROCm/hrx-system/commits/users/benvanik/pr231-refresh-review/).

1. Successful [AMDGPU virtual-memory release](https://github.com/ROCm/hrx-system/blob/5ced8a4e96fe6977ea11600dd9d67aa86e5464a7/runtime/src/iree/hal/drivers/amdgpu/virtual_memory.c#L305-L321) directly destroys the reservation buffer. Queue notification reclaim may still hold an internal reference after the user-visible semaphore advances, so its later release races the freed wrapper. The focused lifecycle test reproduces this under TSAN. After native address release and callback disarm, the implementation must call `iree_hal_buffer_release` to consume the caller's reference. [`48bfc830`](https://github.com/ROCm/hrx-system/commit/48bfc8303e2f719015e27be2b8fc652649c70543) supplies the correction and passes the identical focused TSAN witness.

2. [`DEVICE_UNCACHED`](https://github.com/ROCm/hrx-system/blob/5ced8a4e96fe6977ea11600dd9d67aa86e5464a7/runtime/src/iree/hal/buffer.h#L98-L101) is documented as a required property that unsupported allocators must reject. The refreshed sweep fixes heap, CUDA, HIP, and WebGPU, while Vulkan rejects it through normal memory-type matching. [Metal still reports the request allocatable](https://github.com/ROCm/hrx-system/blob/5ced8a4e96fe6977ea11600dd9d67aa86e5464a7/runtime/src/iree/hal/drivers/metal/direct_allocator.m#L144-L180) despite having no device-cache bypass mechanism. [`088d3b78`](https://github.com/ROCm/hrx-system/commit/088d3b7894ec194300a3badea366d7e9e7f2b8b6) adds the required rejection.

## Landing shape

Please also rebase onto current main. The Linux presubmit failure comes from Loom files added after this branch diverged; I do not consider the gfx1150 runner collapse, empty Windows failure, or currently queued Actions jobs author obligations.

This is now a much smaller and more coherent runtime change. Once the two corrections are folded in and the branch is rebased, I consider the HAL/AMDGPU surface ready to approve.

benvanik (approved):
Thank you for working through the review so carefully. Both remaining runtime requests are addressed exactly as intended.

The AMDGPU VMM release path now consumes the caller's reference with `iree_hal_buffer_release`, allowing any notification-ring reference to retire safely instead of racing direct destruction. The focused VMM lifecycle and full AMDGPU allocator suite pass under TSAN. Metal now also rejects required `DEVICE_UNCACHED` allocations it cannot provide. The rest of the previously reviewed stack is patch-identical after the rebase.

One small current-main integration correction remains before merge: `iree_hip_mem_pool_registry_lookup_retain` needs to become `iree_hip_mem_pool_registry_acquire`, including its eight callsites. Current main's lifecycle checker reserves a trailing `_retain` for void retain functions. The helper already has correct acquire semantics; this is only a naming correction. I tested that exact change locally, including repository formatting, and the complete paranoid precommit then passes with no further failures. I do not need another review round for it.

The remaining red GPU checks do not appear attributable to this change. The gfx110X job used an older gfx1103 runner whose ROCr reports no VMM API support; current main's updated runner-selection flow passes that lane. The gfx1150 failures are broad ROCr `HSA_STATUS_ERROR_OUT_OF_RESOURCES` failures across unrelated executables, and the gfx942 jobs were canceled before producing relevant test evidence.

Approved from the runtime/HAL side. This was a substantial correction cycle, and the resulting ownership and lifecycle behavior is much stronger for it.

PR #235, opened 2026-07-31
[Loom] Keep unroll display names off the hot path

benvanik:
SCF unrolling now treats SSA display names as presentation data instead of compiler identity. It preserves an authored name on the first clone, leaves synthesized induction values and later clones anonymous, and lets serialization assign deterministic numeric spellings when text is requested.

Previously every unrolled iteration formatted suffixes, interned a distinct string, hashed and probed the module string table, and retained that storage for the module lifetime. A 1,024-operation source-to-native witness performed 2,047 such name intern operations. This removes all of them without changing SSA identity, dataflow, scheduling, or emitted code.

The change applies to ordinary and interleaved unrolling, partial unrolling, and downstream transforms that consume the resulting IR. Text remains unambiguous; only cosmetic names change from derived suffixes to value-ID spellings.


benvanik:
Closing this because removing derived unroll names is a regression. Those names preserve transformation provenance for authoring, diagnostics, reports, and debugging; numeric SSA identity is not a substitute. The measured string-table cost needs an implementation fix that retains naming semantics.

PR #243, opened 2026-08-01
[Loom] Honor invocation target facts in math legalization

benvanik:
Math legalization now consumes the same effective function target facts as the rest of the target pipeline. This repairs targetless source kernels selected through an invocation target, whose semantic math operations were previously skipped by legalization and then rejected during target-low lowering.

The pass removes its private authored-symbol lookup and resolves facts through the shared target pass environment. Truly targetless IR remains untouched, while a selected target without a linked math policy continues to fail loud. A host-only `loom-compile --target=gfx1100` witness covers both SiLU decomposition and exact rounding so invocation-selected targets cannot silently diverge again.


benvanik:
I hate the foundation this is modifying. Future changes will rework it.

PR #251, opened 2026-08-03
[Runtime] Add handle-based graph buffer nodes

stellaraccident:


libhrx/include/hrx_runtime.h
stellaraccident: Add comments to these two that they operate on device pointers and have extra accounting overhead compared to the ones that operate on buffer handles.

libhrx/include/hrx_runtime.h
```diff
@@ -1088,6 +1105,14 @@ HRX_API hrx_status_t hrx_graph_add_memcpy_node(
 HRX_API hrx_status_t hrx_graph_add_memset_node(
     hrx_graph_t graph, const hrx_graph_node_t* deps, size_t dep_count,
     const hrx_graph_memset_node_attrs_t* attrs, hrx_graph_node_t* out_node);
+HRX_API hrx_status_t hrx_graph_add_copy_buffer_node(
```
stellaraccident, line 1116: Note that these operate on buffer handles and have no extra accounting overhead compared to those running on raw device pointers.

libhrx/include/hrx_runtime.h
stellaraccident: Let's rename these to match the naming convention below (and rename them in the hip impl). e.g. hrx_graph_add_copy_ptr_node and hrx_graph_add_fill_ptr_node

benvanik (approved):

PR #287, opened 2026-08-06
[Loom/HRX] Tighten launch-path ownership

benvanik:
Workload launch configuration evaluation now treats dynamic workload values as temporary analysis facts over an immutable retained module. Only config materialization clones and rewrites IR, removing whole-module link work from ordinary dynamic launch evaluation and making the analysis boundary const-correct.

HRX graph launch now delegates pending command-buffer submission to the stream flush owner and serializes executable semaphore queries with timeline advancement. This closes a command-buffer lifetime leak and makes concurrent launches observe one coherent executable state transaction.

## Reviewer Notes

The two commits are independent. The Loom commit centers on the distinction between analysis inputs and IR specialization. The HRX commit centers on stream command-buffer ownership and the existing executable mutex contract.

### DIFF
--- a/libhrx/src/libhrx/graph_buffer_test.cc
+++ b/libhrx/src/libhrx/graph_buffer_test.cc
@@ -114,6 +114,36 @@ TEST_F(GraphBufferTest, DependentFillThenCopyProducesExpectedContents) {
   hrx_graph_release(graph);
 }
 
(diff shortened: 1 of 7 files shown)


benvanik:
Superseded by the separated draft reviews: HRX #288 and Loom #289. Closing this mixed-layer review so each change can be reviewed and landed independently.

PR #293, opened 2026-08-06
[Loom/AMDGPU] Expand native RDNA target capabilities

benvanik:
Loom's exact AMDGPU targets now expose native instruction and scheduling families that were previously hidden behind the conservative gfx11 intersection or incorrectly inherited across RDNA generations. Target-independent kernels can retain one source ABI while descriptor capabilities select materially different scalar, cross-lane, matrix, narrow-float, and packet forms for gfx115x, gfx12, and gfx117x processors.

**Design**

The AMDGPU target model now treats generic families as portability intersections and exact processors as capability refinements. gfx1170 through gfx1172 receive a distinct RDNA 4m descriptor set and processor profile instead of inheriting RDNA 3.5 matrix and scheduling facts. Exact-target additions are published through generated descriptor capabilities and indexed target facts; lowering does not branch on processor names, scan registries, or alter kernel ABIs.

Capability additions cover complete instruction families rather than the individual operations encountered by motivating workloads. This keeps type, predicate, operand-form, and encoding support even across the architecture surface and prevents nearby source shapes from falling unexpectedly onto generic expansion paths.

**Scalar And Cross-Lane Operations**

RDNA 3.5 and later exact targets expose complete f16 and f32 scalar arithmetic, conversion, rounding, and comparison families. AMDGPU value analysis retains eligible subgroup-uniform floating-point values in SGPRs, including SCC-producing comparisons and width-preserving select chains, while lane-varying values and generic targets retain their existing VGPR behavior. Register-part operands and fixed assembly modifiers are represented in descriptor-native forms so half-register contracts survive selection and print canonically without string reconstruction.

The complete VINTERP arithmetic family is represented with its mixed f16/f32 register parts, fixed interpolation waits, and implicit M0 dependency. DPP8 is modeled as eight packed lane selectors, while DPP8 and DPP16 variants on RDNA 3.5 and RDNA 4m accept uniform or inline right-hand operands wherever the exact VOP3 encoding permits them. Arithmetic, integer, bitwise, carry, and comparison families share the same generated policy instead of growing one-off target descriptors.

Wave64 integer reductions now use the same payload-neutral scalar publication strategy as floating-point reductions. Supported 32-bit add, multiply, min/max, and bitwise reductions combine native wave32 halves through DPP or permlanex16 and publish through legal SGPR forms without introducing LDS or barriers.

**Matrix And Narrow-Float Operations**

RDNA 4 and RDNA 4m descriptor sets expose the complete wave32 and wave64 WMMA and SWMMAC families with their distinct physical payload widths, fragment layouts, and low descriptors. Narrow IU4 forms represent ignored upper lanes through replicated fragment coordinates rather than changing source semantics.

RDNA 4m gains the OCP FP8 and BF8 conversion and packed-dot families proven by LLVM MC, including every byte selector, packed low/high encode forms, and all mixed FP8/BF8 dot products. Scaled finite FP8 and BF8 fragment loads preserve their scale operands through planning and storage demand, allowing exact targets to select native widening and BF16 packing while older targets retain the existing software decode route. The selected conversion route remains visible in compile reports and does not change the kernel binding contract.

RDNA 4m min/max descriptors now match the numeric and IEEE instruction identities accepted by gfx117x, including VOP2, VOP3, VOP3P, compact clamp, and VOPD forms. Native operations are selected when the complete family is available; indexed compare/select expansions remain the fallback for targets without those capabilities.

**Packet And Scheduling Fidelity**

The gfx117x packet inventory now models dual register moves whose second source uses the target's SRC2 bank rules, compact equality comparisons with inline right operands, and conditional selects whose predicate remains an implicit VCC dependency. Planning preserves semantic dependencies while encoding only physical packet operands, and target-specific source caches prevent gfx11 bank restrictions from leaking into RDNA 4m.

gfx1170 through gfx1172 publish the S_DELAY_ALU scheduling capability used by LLVM's GFX11 delay insertion. Transition depctr and VMEM-result ordering remain absent until independently proven. This makes generated artifacts and target comparisons truthful without claiming scheduling features for which there is no oracle evidence.

**Impact**

Exact targets can now remove avoidable SGPR-to-VGPR materialization, select native scalar floating-point chains, preserve uniform DPP operands, use target-correct matrix fragment ABIs, keep FP8/BF8 storage narrow through fragment loading, and form compact compare/select and dual-move packets. These choices happen during JIT specialization from the same source declarations and launch ABI; applications do not need architecture-specific kernel entry points or routing tables.

The gfx11-generic contract remains the intersection shared by its members. A kernel compiled for a generic family therefore keeps portable operand and instruction assumptions, while compilation for an exact processor can use capabilities established by that processor's descriptor set. Unsupported or unproven features fail to appear rather than being inferred from architecture naming.

The target inventory is also more useful as compiler evidence. Compile reports can distinguish native operations from software expansion, attribute scaled fragment conversion routes, and compare exact-target resource and instruction choices without conflating inherited facts with verified capabilities.

**Evidence Boundaries**

RDNA 3.5 coverage is grounded in the vendor ISA descriptions and retains the generic gfx11 intersection wherever exact encodings differ. The available RDNA 3.5 ISA does not publish native FP8/BF8 matrix families, so this change does not manufacture them; gfx115x continues to use optimized software conversion where required.

The published ISA archive does not include gfx117x XML. RDNA 4m additions are therefore limited to instruction encodings, target membership, and schedule classifications independently accepted by LLVM MC and its target model. The generic gfx11 code-object target remains artifact-decodable but exact-only for compilation while the pinned device compiler rejects the generic processor spelling.

**Reviewer Notes**

The highest-value review surfaces are the exact-versus-generic descriptor-set boundaries, scalar-value placement rules, generated DPP and compare inventories, fragment payload and scale propagation, and gfx117x packet operand contracts. Assembly coverage intentionally pins canonical native syntax and packet bytes, while source-to-low coverage verifies that natural target-independent IR reaches each family rather than requiring authored low-level assembly.


benvanik:
ROCm CI is .... doing its thing. Local presubmit passes.

PR #299, opened 2026-08-06
[Loom] Bound select placement without analysis state

benvanik:
Conservative module-local AMDGPU placement queries have no analysis cache, but native scalar select placement routed select payloads through the full mutually recursive analysis. A dynamic workgroup clamp could then cycle between its condition and result until the compiler stack overflowed.

Restore the bounded direct producer-and-fact predicate only when no analysis state exists. Context-backed lowering and legality continue to use the full cached analysis, preserving native scalar-float placement precision. Source-low corpus coverage exercises a dynamic workgroup clamp feeding a global memory address.


benvanik:
I hate this, but lower.c and co need major reworking before it can be done better.

PR #312, opened 2026-08-08
[Loom] Expose glibc spawn closefrom declaration

dan-garvey:
## Summary
- define `_GNU_SOURCE` before system headers in the Linux process runner
- expose `posix_spawn_file_actions_addclosefrom_np` on glibc 2.34+ strict builds while preserving the existing fallback on other libc versions

## Test plan
- [x] `python dev.py bazel build //loom/src/loom/target/tool:process`
- [x] `python dev.py bazel test --test_output=errors //loom/src/loom/target/tool:process_test`
- [x] build all ten targets from `loom/README.md` without workaround flags
- [ ] `python dev.py bazel precommit` (937 tests passed and 19 skipped; the SUT's unrelated LLVM 14 tool test rejects opaque-pointer IR)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/loom/src/loom/target/tool/process_linux.c
+++ b/loom/src/loom/target/tool/process_linux.c
@@ -4,6 +4,12 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+// _GNU_SOURCE must be defined before any system header so that <spawn.h>
+// declares posix_spawn_file_actions_addclosefrom_np on glibc.
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif  // _GNU_SOURCE
+
 #include "loom/target/tool/process_platform.h"
 
 #if defined(IREE_PLATFORM_LINUX) || defined(IREE_PLATFORM_ANDROID)


dan-garvey:
Was trying out the readme with an agent and hit this. Land if you want it

loom/src/loom/target/tool/process_linux.c
```diff
@@ -4,6 +4,12 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+// _GNU_SOURCE must be defined before any system header so that <spawn.h>
+// declares posix_spawn_file_actions_addclosefrom_np on glibc.
```
benvanik, line 8:
```suggestion
```

benvanik (changes requested):

benvanik:
Thanks for catching this. I reproduced the failure by compiling only process_linux.c under strict C17. The declaration is gated by glibc's __USE_MISC feature set, so #319 landed the fix using a translation-unit-local _DEFAULT_SOURCE rather than the broader _GNU_SOURCE. That keeps the feature namespace scoped to the one consumer while preserving the native glibc path. Closing this as superseded, with your report credited in the replacement.

PR #333, opened 2026-08-11
[Loom] Produce and evaluate launch configuration artifacts

benvanik:
Loom now separates launch-configuration evaluation from device executable compilation with a small compiler-produced artifact that can be loaded once and invoked repeatedly as workload values change. A single artifact contains the launch functions for every prepared kernel in the source module, so applications no longer retain the source module or re-enter compilation to calculate workgroup counts.

The compiler materializes each kernel launch region as an ordinary public pure `func.def` with the kernel's symbol name. Its typed scalar workload operands become function arguments and its xyz workgroup count becomes the result:

```loom
func.def public pure @prefill(%token_count: index) -> (index, index, index) {
  %rounding = index.constant 127 : index
  %rounded = index.add %token_count, %rounding : index
  %items_per_workgroup = index.constant 4 : index
  %workgroups = index.div %rounded, %items_per_workgroup : index
  %subgroup_size = index.constant 32 : index
  %unit = index.constant 1 : index
  func.return %workgroups, %subgroup_size, %unit : index, index, index
}
```

Configuration values, template and provider selection, and target specialization are resolved before this boundary. Extraction uses Loom's normal value-fact engine to reify exact symbolic results into the artifact. Any symbolic dependency that remains unresolved fails extraction instead of copying source configuration or target state into the runtime module. The original source module remains unchanged and continues through device lowering independently.

The loomc API loads and verifies the resulting artifact into an immutable shareable `loomc_launch_config_module_t`. Callers enumerate functions or resolve a stable public name once, create an exclusive reusable `loomc_launch_config_context_t`, and evaluate module-local function tokens directly into caller-owned packed xyz storage. Multiple contexts can evaluate one module concurrently, while steady-state evaluation performs no parsing, source configuration, specialization, string lookup, output allocation, or global locking.

Workload values use positional raw 64-bit slots interpreted according to each function's authored scalar signature. The documented encoding covers i1 through i64, index and offset, FP8, f16, bf16, f32, and f64. The output can point directly into mapped host-local/device-visible storage used by indirect dispatch. The loader and evaluator support functions returning multiple deduplicated xyz tuples, allowing the same API to serve aggregate command-program launch calculations as that producer lands.

Artifact storage may be borrowed, copied, or transferred with an explicit release callback. Loombc is the initial private backing representation; callers observe only module loading, function lookup, context creation, invocation, and packed output. A future VM implementation can therefore replace Loombc evaluation without changing the public API or application ownership flow.

The former `loomc_module_evaluate_launch_config` API and its source-module replay implementation are removed. `loomc_launch_config_context_evaluate_one` retains the conventional single-kernel projection without preserving the old ownership or compilation model.

## Reviewer Notes

The public ownership, concurrency, scalar ABI, and invocation contract is concentrated in `loom/binding/c/include/loomc/launch_config_module.h`. The current Loombc loader and value-fact adapter are private to `loom/binding/c/src/launch_config_module.c`. Compiler-side extraction is isolated in `loom/src/loom/transforms/kernel/launch_config_module.c`, where configuration, template, provider, and target decisions disappear into ordinary function IR. The small `float_facts` change factors existing raw-bit conversion so runtime arguments and compiler bitcasts share one fact representation.


benvanik:
Closing this draft rather than revising it. The proposed API incorrectly conflates per-kernel launch configuration with command-program indirect-dispatch planning. The replacement will restart from main and preserve launch configuration as the complete companion contract for standalone compiled kernels; command programs will own their scheduling calculations independently.

PR #349, opened 2026-08-12
[HAL/AMDGPU] Report peer transport through topology edges

AWoloszyn:
PR description

  ## Summary

  Extend generic HAL topology edges with normalized physical transport
  information and populate it from the AMDGPU backend.

  This adds:

  - A backend-neutral physical link type.
  - A physical path hop count.
  - Accessors and builders for both values.
  - AMDGPU topology refinement using ROCr peer-link information.
  - Formatting and unit-test coverage for the new topology fields.

  ## Motivation

  Higher runtime layers need peer topology information to implement APIs such as
  `hipDeviceGetP2PAttribute` and `hipExtGetLinkTypeAndHopCount`. The existing HAL
  topology represented portable scheduling properties, peer access, and atomic
  capabilities, but discarded the physical transport type and path length.

  Exposing a separate AMDGPU query API would couple those layers directly to the
  backend and duplicate topology discovery. Instead, this change carries the
  information through the existing generic topology edge.

  Backends that cannot report physical transport details retain
  `IREE_HAL_TOPOLOGY_LINK_TYPE_UNKNOWN` and a zero hop count.

  ## Design

  The topology edge's previously reserved cold-word bits now contain:

  - An 8-bit normalized first-hop transport type.
  - An 8-bit physical path hop count.

  These values remain outside the scheduling word. Existing placement and
  resource-origin hot paths continue to read and cache only the scheduling word,
  so this does not add work or storage to dispatch and compatibility paths.

  AMDGPU obtains the information while performing its existing topology
  refinement:

  - Memory-pool access determines direct peer accessibility.
  - Native atomic support is derived from the reported peer path.
  - ROCr link types are normalized into generic HAL link types.
  - Hop count is recovered from ROCr's aggregate link record and documented NUMA
    distance conventions.
  - Composite logical devices use the longest path across their physical device
    pairs.
  - A composite edge reports a link type only when every physical pair has the
    same first-hop transport; otherwise it reports `UNKNOWN`.

  Topology discovery remains a control-plane operation performed during device
  registry construction. No ROCr query or additional branch is added to kernel
  submission.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/api.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/api.h
@@ -495,6 +495,11 @@ IREE_API_EXPORT iree_status_t iree_hal_amdgpu_logical_device_create(
     const iree_hal_device_create_params_t* create_params,
     iree_allocator_t host_allocator, iree_hal_device_t** out_device);
 
+// Queries immutable physical-link details between two AMDGPU logical devices.
+IREE_API_EXPORT iree_status_t iree_hal_amdgpu_device_query_peer_link(
+    iree_hal_device_t* source_device, iree_hal_device_t* destination_device,
+    iree_hal_amdgpu_link_type_t* out_link_type, uint32_t* out_hop_count);
+
 //===----------------------------------------------------------------------===//
 // iree_hal_amdgpu_driver_t
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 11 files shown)


benvanik (approved):
Approved. I traced the AMDGPU path through ROCr and production CLR. ROCr exposes one aggregate LINK_INFO record and clamps num_hop because it cannot describe each physical hop; reconstructing the physical path length from summed NUMA distance with 13 for xGMI and 20 otherwise matches the existing HIP implementation. The composite-device reduction is conservative and order-independent, and placing these facts in previously reserved cold-edge bits leaves the hot scheduling word and object sizes unchanged.

One semantic caveat worth preserving in future documentation is that this is the transport of ROCr's aggregate path record plus a HIP-compatible reconstructed path length, not a full physical route graph. The UNKNOWN result for mixed composite paths is the right behavior.

Two follow-up consumer details are important but do not belong in this PR: the generic HAL enum values intentionally differ from HSA/HIP and therefore require an explicit mapping in libhrx, and remoting should carry named validated topology fields rather than serializing the packed in-process edge. This gives both consumers the information they need without coupling either one to AMDGPU.

benvanik (approved):

PR #350, opened 2026-08-12
[AMDGPU] Enqueue queue-ordered buffer value waits

AWoloszyn:
PR description

  ## Summary

  Add an AMDGPU-private queue operation that waits until a 32-bit or 64-bit
  value in device-visible memory satisfies a comparison.

  The operation supports:

  - Greater-than-or-equal and equality comparisons.
  - Bitwise AND and bitwise NOR conditions.
  - 32-bit and 64-bit comparison masks.
  - Existing HAL wait and signal semaphore ordering.
  - Deferred submission under prerequisite waits or queue pressure.

  The implementation uses a small device builtin submitted through the normal
  AMDGPU host queue. It does not poll memory from the host.

  ## Motivation

  HIP exposes stream-ordered waits on arbitrary device-visible memory through
  `hipStreamWaitValue32`, `hipStreamWaitValue64`, and
  `hipStreamBatchMemOp`.

  These operations cannot be represented by HAL timeline semaphore waits. The
  watched value belongs to the application, and the operation has explicit
  comparison and mask semantics:

  1. Earlier stream operations must complete before polling starts.
  2. Later stream operations must remain blocked until the condition succeeds.
  3. The target allocation must remain alive while the queue is waiting.
  4. A producer on another stream, device, or permitted host mapping must be
     observable with the appropriate device memory scope.

  Host polling cannot provide this contract without occupying a host thread,
  adding wake-up latency, and separating the observation from device queue
  ordering.

  ## Design

  The private AMDGPU API accepts a retained HAL buffer and byte offset rather
  than a raw pointer. It validates:

  - Non-null devices, buffers, and semaphore-list storage.
  - A value width of exactly four or eight bytes.
  - Natural alignment of both the subspan offset and backing device address.
  - Buffer usage, access rights, and target range.
  - Checked subspan-offset arithmetic.
  - Supported conditions and flags.
  - Queue affinity and device ownership through normal queue selection.

  A one-work-item builtin repeatedly performs a system-scope atomic load, applies
  the requested mask and comparison, and yields while the condition is false.
  The public operation and builtin ABI have compile-time checks preventing their
  condition ordinals from diverging.

  The generated AMDGPU target binaries contain both 32-bit and 64-bit support for
  all currently supported target families.

  ## Queue Ordering And Lifetime

  The wait uses the existing host-queue dispatch, AQL ring, kernarg ring, and
  pending-operation machinery.

  When prerequisites and queue capacity are ready, the builtin is submitted
  directly. Otherwise, the complete operation is captured for later issue,
  including its scalar parameters, semaphore lists, and retained target buffer.

  The target remains retained through prerequisite waits, queue-capacity
  deferral, execution, cancellation, and failure handling. Signal semaphores
  advance only after the polling builtin exits, preserving ordering for following
  stream work.

  An operation is never partially submitted. Cancellation and asynchronous
  device failure use the existing pending-operation cleanup paths.

  ## API Boundary

  This remains an AMDGPU-private operation rather than a generic HAL API.

  HAL semaphores represent runtime-owned timeline values. A wait on arbitrary
  application memory has a different ownership and memory-model contract, and
  other backends may implement it using native wait packets, shaders, kernels, or
  not at all.

  AMDGPU owns the target-specific atomic scope, builtin binary, AQL packet,
  kernarg allocation, and queue resource lifetime needed for this implementation.
  The HIP binding remains responsible for translating HIP flags and resolving
  HIP pointers into retained HAL buffers.

  ## Performance

  Ordinary dispatch and transfer submission paths are unchanged.

  Only an explicit value-wait request pays the cost of preparing and submitting
  the one-work-item builtin. Steady-state submission uses the existing AQL and
  kernarg rings without allocating, and no host polling thread is created.


benvanik:
Investigating now to see how this fits. There's a few options, but the easiest landing path (that can unblock but still upgrade later) is to just use queue_dispatch and build the kernel yourself under libhrx (you can reuse the same build rules that the HAL device library uses) - that is the best way to support the HIP behavior exactly and is compatible with remoting/replay. We may still want a HAL-level spinloop primitive for where we can implement it efficiently (PM4), but the AMD punch-through into the HAL isn't viable.

AWoloszyn:
This has been superceded

PR #351, opened 2026-08-12
[AMDGPU] Add lazy queue provisioning for execution contexts

AWoloszyn:
## Summary

  Add AMDGPU queue provisioning needed to support execution contexts with
  restricted execution-unit masks.

  The device now distinguishes between:

  - queue capacity reserved for stable affinity ordinals; and
  - the initial queue count created during device initialization.

  Ordinary queues are created eagerly. Additional specialized queues are created
  lazily when first selected, avoiding the resource cost of creating every
  possible queue up front.

  ## Motivation

  HIP execution contexts can restrict streams to an exact set of execution units.
  The restriction is a persistent property of the underlying HSA queue, not a
  per-dispatch parameter. Each concurrently usable restriction therefore requires
  an independently owned queue.

  This enables follow-on support for APIs including:

  - `hipExtStreamCreateWithCUMask`
  - `hipExtStreamGetCUMask`
  - `hipDevSmResourceSplit`
  - `hipDevSmResourceSplitByCount`
  - execution-resource and green-context creation
  - streams created from restricted execution resources

  ## Design

  Queue-affinity ordinals remain stable for the lifetime of the device. The
  logical device reserves storage for the full configured capacity while only
  materializing the initial queue set during device creation.

  Selecting an inactive ordinal activates the required queue under a dedicated
  activation mutex. The queue is published through the active-queue mask only
  after its initialization completes. Subsequent use of an active queue avoids
  the activation lock.

  The ordinary dispatch path is unchanged: initial queue ordinals return before
  the lazy-activation atomic check or mutex.

  A private AMDGPU operation configures an exact execution-unit mask on an
  exclusively owned queue. It validates the complete mask width and serializes the
  native mask operation with queue submission. The owner remains responsible for
  ensuring that the queue is idle before changing or restoring its mask.

  Profiling requires a fixed queue topology, so all advertised queues are
  materialized before profiling freezes that topology.

  ## Validation

  The implementation rejects:

  - duplicate queue configuration extensions;
  - invalid capacity or initial-count combinations;
  - out-of-range and non-singular queue affinities;
  - null, partial, or incorrectly sized execution-unit masks; and
  - queue activation after profiling has frozen the topology.

  The focused AMDGPU target builds successfully:

  `python dev.py --cmake-build-dir build/pair-amdgpu-check cmake build
  iree_hal_drivers_amdgpu_amdgpu --parallel 16`

  ## API Scope

  The new configuration and mask operations are private to the AMDGPU driver.
  No generic HAL interface changes are introduced.

  This boundary is intentional: persistent HSA queue masks and native queue
  lifecycle management are AMDGPU-specific, while execution-resource policy and
  queue ownership remain in the HIP integration layer.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/api.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/api.h
@@ -28,7 +28,26 @@ typedef uint32_t iree_hal_amdgpu_device_create_params_extension_type_t;
 enum iree_hal_amdgpu_device_create_params_extension_type_e {
   // Provides one opaque hostcall service for each physical device.
   IREE_HAL_AMDGPU_DEVICE_CREATE_PARAMS_EXTENSION_TYPE_HOSTCALL_PROVIDER = 1u,
+  // Configures eager and lazily activated host queue counts.
+  IREE_HAL_AMDGPU_DEVICE_CREATE_PARAMS_EXTENSION_TYPE_HOST_QUEUES = 2u,
 };
+
+// AMDGPU device creation extension controlling host queue provisioning.
+//
+// The device allocates |capacity| queue slots and advertises them through its
+// queue affinities. The first |initial_count| queues are initialized when the
+// device is assigned to its topology. Remaining queues are initialized on
+// first use, avoiding HSA queue and ring allocation costs until needed.
+typedef struct iree_hal_amdgpu_host_queue_extension_t {
+  // Common device creation extension prefix.
+  iree_hal_device_create_params_extension_t base;
+
+  // Maximum number of host queues per physical device.
+  iree_host_size_t capacity;
+
+  // Number of host queues initialized during device setup.
+  iree_host_size_t initial_count;
+} iree_hal_amdgpu_host_queue_extension_t;
 
 // Scalar physical-device facts available to opaque hostcall providers.
 typedef struct iree_hal_amdgpu_hostcall_provider_device_info_t {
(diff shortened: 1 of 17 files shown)


AWoloszyn:
I pushed a follow-up addressing the queue lifecycle and concurrency issues.

  The key change is that specialized queues now have an explicit lease lifecycle:

  - Queue initialization is published atomically, so concurrent executable loading cannot observe partially initialized
    queue storage or size its queue-scope array from a changing count.

  - Eagerly initialized queues, such as those required by instrumentation, remain distinct from ordinary queues exposed
    for general submission.

  - Specialized affinities are unavailable until native CU-mask configuration succeeds and the lease is published.
  - Final release first closes the queue to new submissions, waits for its last committed GPU epoch, and only then
    restores the default CU mask.

  - A failed mask update is reset before the slot can be reused. If native state cannot be restored, the slot is
    quarantined and the device reports the failure instead of silently reusing an unknown configuration.

  - Ordinary queue submission remains on the existing fast path. Only specialized affinities pay the additional lease-
    publication check.

  This preserves the intended model: each distinct CU mask owns a dedicated hardware queue for the duration of its
  lease, without conflating virtual streams with queue configuration or allowing a queue to be reconfigured while GPU
  work is still executing.

  I added coverage proving that a specialized affinity is rejected before acquisition, accepted while leased, and
  rejected after release. The release test submits actual GPU work before dropping the lease, exercising the required
  wait-before-reset behavior. All seven relevant AMDGPU GPU-backed tests pass, along with formatting, hygiene, Semgrep,
  and clang-tidy checks.

benvanik:
Closing this older queue-provisioning carrier in favor of #444, which is the current review thread for the reduced implementation. The fixed-physical-queue successor is `users/zjgarvey/handoff/pr444-native-2026-08-31` at `44fb26f2850f720ecea67ba89bcd491dc05bfb78`; it removes HAL-owned queue selection, equal-mask sharing, leasing, and recycling. The successor review direction is recorded at https://github.com/ROCm/hrx-system/pull/444#issuecomment-5501176675.

PR #354, opened 2026-08-12
[AMDGPU] Expose loaded-kernel resource metadata

AWoloszyn:
## Summary

  Parse, validate, and expose immutable kernel resource metadata from the HSACO image selected by the AMDGPU executable
  loader.

  The change records the following properties for each loaded kernel:

  - maximum flat workgroup size;
  - fixed group-segment memory;
  - fixed private-segment memory;
  - vector register count;
  - maximum dynamic group-segment memory; and
  - whether the kernel requires uniform workgroups.

  A private AMDGPU query returns these properties for a loaded executable function. This does not extend the generic HAL
  interface.

  ## Motivation

  HIP exposes loaded-kernel properties and uses them for launch validation and occupancy calculations. This supports
  downstream implementations of:

  - hipFuncGetAttribute
  - hipFuncGetAttributes
  - occupancy query APIs
  - dynamic shared-memory validation
  - ordinary and extended kernel launch validation
  - partial-workgroup validation
  - cooperative-launch resource checks

  These values must describe the exact architecture-specific HSACO image selected by the executable loader. Reparsing a
  module in the HIP layer could inspect the wrong image and would duplicate the code-object parser.

  ## Design

  The existing AMDGPU HSACO metadata parser reads the additional resource fields while loading the selected image.
  Values are normalized into immutable per-export metadata owned by the executable.

  Executable loading rejects:

  - missing required resource fields;
  - duplicate fields;
  - invalid metadata types;
  - invalid uniform-workgroup values;
  - zero maximum flat workgroup sizes; and
  - workgroup limits exceeding the physical-device limit.

  The private query validates the function ordinal and copies the stored values to the caller. It does not expose
  MessagePack nodes, code-object storage, or parser internals.

  The AMDGPU driver reports immutable executable facts only. HIP-specific attribute mapping, occupancy policy, and
  launch policy remain in the binding layer.

  ## Performance

  Metadata parsing and validation occur once during executable loading.

  The query reads immutable fields directly and requires no allocation, parsing, search, or synchronization. Ordinary
  kernel dispatch does not call the query and gains no additional hot-path work.

  ## API Boundary

  This is an AMDGPU-private API because the values come from HSACO metadata and have backend-specific semantics. Other
  executable formats are not required to invent equivalents, and no generic HAL vtable or interface is changed.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/api.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/api.h
@@ -495,6 +495,16 @@ IREE_API_EXPORT iree_status_t iree_hal_amdgpu_logical_device_create(
     const iree_hal_device_create_params_t* create_params,
     iree_allocator_t host_allocator, iree_hal_device_t** out_device);
 
+// Queries immutable metadata for one executable function.
+IREE_API_EXPORT iree_status_t
+iree_hal_amdgpu_executable_query_function_attributes(
+    iree_hal_executable_t* executable, iree_hal_executable_function_t function,
+    uint32_t* out_maximum_workgroup_invocations,
+    uint32_t* out_workgroup_local_memory_size,
+    uint32_t* out_private_memory_size, uint32_t* out_register_count,
+    uint32_t* out_maximum_dynamic_workgroup_local_memory_size,
+    bool* out_requires_uniform_workgroups);
+
 //===----------------------------------------------------------------------===//
 // iree_hal_amdgpu_driver_t
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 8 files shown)


benvanik:
Thanks for putting this together and for isolating the loaded HSACO metadata requirements. After tracing the callers, we’re going to take ownership of the runtime/HAL portion and supersede this PR with a generic executable-function metadata path that works through replay and remoting.

Please pause revisions here, along with any dependent work that relies on the AMDGPU-private query. We’ll preserve the useful parser and test work with attribution, then follow up with the replacement API and PR for libhrx to consume.

benvanik:
Superseded by #360. It carries forward the HSACO parser and required-metadata validation from this PR with your authorship preserved, exposes the loaded facts through generic `iree_hal_executable_function_info`, enforces selected-device limits in AMDGPU, and maps them into libhrx without a concrete-driver dependency.

Thank you for isolating the resource metadata requirements and for pausing while we moved the ownership boundary. The resulting path supports the compatibility APIs you need, continues through replay, and gives remoting one backend-neutral record to publish and cache.

PR #360, opened 2026-08-12
[HAL] Expose loaded executable function resource facts

benvanik:
Expose immutable resource facts for each loaded executable function so runtime clients can inspect the selected artifact through generic HAL instead of linking a concrete driver. This supersedes #354 while carrying forward its useful HSACO metadata parsing and validation with the original authorship preserved.

**Design**

`iree_hal_executable_function_info_t` now reports an optional function-specific workgroup invocation limit and a provided-fields resource record for fixed workgroup-local memory, fixed private memory, and per-invocation register use. Provided bits distinguish an unavailable observation from a measured zero, and all facts describe the exact image selected by executable loading.

The contract stays backend-neutral. Drivers publish only facts they know, replay continues delegating through the generic executable, and clients never downcast or issue target-specific property queries. Dynamic shared-memory policy and mutable HIP/CUDA attributes remain outside HAL.

**AMDGPU**

The HSACO path requires and parses the resource metadata used by the loaded kernel, then cross-checks parser values against the HSA-loaded symbol. Executable loading validates function workgroup limits, fixed workgroup dimensions, and fixed LDS use against the selected logical device.

Native dispatch descriptors receive the owning physical device's LDS capacity and precompute the legal dynamic LDS bound as capacity minus fixed use. AQL and host-queue dispatch validation therefore reject one byte beyond the real device limit instead of relying on 32-bit arithmetic headroom.

**libhrx**

Module loading copies the generic facts once into a libhrx-owned attribute record. HIP and CUDA queries, setters, launch validation, and compatibility occupancy calculations consume that record without including AMDGPU headers or calling driver-private symbols.

The initial dynamic shared-memory limit is derived from the device's default per-workgroup capacity, while the setter ceiling uses its opt-in capacity. The mutable setting is atomic for concurrent host API calls, and launch validation preserves the distinction between a known zero limit and unavailable metadata.

This change deliberately does not claim generic occupancy or partial-workgroup execution. Those require additional target facts and an exact-global-size dispatch contract, respectively; neither belongs in a metadata-only API.

**Reviewer Notes**

The highest-value review surfaces are the generic ownership and absence semantics in `runtime/src/iree/hal/executable.h`, AMDGPU load/descriptor limit normalization, and the libhrx default-versus-opt-in shared-memory mapping.


benvanik:
CI exposed a CMake regression already present on `main`: platform-conditional Loom descriptor and encoding tables were excluded unconditionally, so gfx942 configuration referenced the correctly absent CDNA4 targets. That repair is isolated in #361.

Once #361 lands I'll rebase this branch onto the corrected `main` and rerun CI. The other failed jobs in this run terminated while downloading Bazel, `rules_python`, GoogleTest, or Google Benchmark; the completed CPU, Vulkan, Windows, sanitizer, gfx110X/gfx1151, and gfx942 TSAN/UBSAN build and test surfaces passed.

AWoloszyn (approved):

PR #361, opened 2026-08-12
[Loom] Select AMDGPU descriptor tables by target

benvanik:
AMDGPU descriptor and encoding tables now follow the same target-selection model as the contracts that consume them. A narrow profile such as `gfx942` declares only its CDNA3 storage tables; generic profiles select the exact storage backing their view; `loom_defaults` continues to make every supported table available.

The storage closure is derived from the existing target-info view-to-storage relationships instead of architecture lists duplicated across BUILD and CMake. This keeps Bazel and generated CMake aligned as new processors, generic profiles, and stepping overlays are added.

This removes the descriptor-specific `EXCLUDE_FROM_ALL` flag and post-declaration mutation path. Registries and encoders already express real dependencies on selected tables, so ordinary target compatibility now defines both availability and build membership. It also fixes narrow CMake configurations that previously tried to exclude a descriptor library which target selection had correctly omitted.


benvanik:
I dunno, probably fine. Stupid ROCM CI.

PR #366, opened 2026-08-13
[AMDGPU] Propagate asynchronous GPU faults through queue failures

AWoloszyn:
## Summary

  Propagate asynchronous AMDGPU memory faults and fatal system events through the existing IREE device, queue,
  semaphore, and status failure paths.

  GPU faults can occur after a kernel submission has already returned successfully. Without native fault delivery,
  pending synchronization may block indefinitely and later operations may incorrectly appear to succeed. This change
  converts those native events into durable failures that callers can observe through normal HAL operations.

  ## Implementation

  - Registers one process-wide HSA system-event callback.
  - Associates native GPU agents with their live IREE logical devices.
  - Latches the first fatal error on the affected logical device.
  - Fails host queues associated with the affected physical device.
  - Wakes pending semaphore, frontier, and notification waiters with the failure.
  - Rejects later submissions after the device has entered the failed state.
  - Coordinates callback delivery with device teardown to prevent access to destroyed queues.
  - Preserves the first failure when duplicate native notifications arrive.

  ## HIP Impact

  This provides the asynchronous error delivery required for faulting kernel launches to be observed by operations such
  as:

  - hipStreamSynchronize
  - hipStreamQuery
  - hipDeviceSynchronize
  - event query and synchronization
  - stream dependencies waiting on faulted work
  The HIP binding remains responsible for translating the resulting IREE status into the appropriate HIP error.

  ## Architecture

  Native event ingestion remains private to the AMDGPU driver. The portable HAL already represents the externally
  visible behavior through failed operations, semaphores, queues, and devices, so this change does not add a generic HAL
  API or expose HSA agents outside the driver.

  This also preserves the remoting boundary: a remote implementation only needs to transport ordinary HAL failures
  rather than reproduce the HSA event model on the client.


zjgarvey:
942 error seems unrelated, so I'm going to land this.

zjgarvey:
I'm so sorry, I had two pages open at once and meant to merge a different PR.

I'll reopen this one and land the other.

PR #367, opened 2026-08-13
[HAL/AMDGPU] Add execution-scoped virtual memory access

AWoloszyn:
## Summary

  Add portable host and device execution scopes to the HAL virtual-memory protection API and implement them in the
  AMDGPU backend.

  This enables callers to distinguish between memory accessible by device execution, host execution, or both. Queue
  affinity continues to select the participating device topology, while the new access scope independently selects the
  execution domains receiving access.

  AMDGPU resolves these portable scopes to native agents internally. Native agent handles and access-policy details
  remain private to the driver.

  The change also improves host-memory imports by preserving both the original host address and the agent-visible
  address when native registration produces different values.

  ## Motivation

  HIP virtual-memory APIs such as hipMemSetAccess require independent host and device access control. Queue affinity
  alone cannot represent this distinction because host execution is not a device queue.

  Handling this through an AMDGPU-private API would require HRX to identify the concrete backend and directly call
  driver internals. That would break HAL encapsulation and would not work through allocator wrappers, proxies, or remote
  execution.

  The generic execution-scope contract preserves the requested policy through the HAL boundary while leaving each
  backend responsible for translating or rejecting it.

  ## Implementation

  - Add host, device, and combined virtual-memory access scopes to the HAL allocator API.
  - Forward the scope through caching, debugging, replay, and other allocator wrappers.
  - Have backends return IREE_STATUS_UNIMPLEMENTED when they cannot independently enforce a valid scope.
  - Resolve AMDGPU scopes to bounded, deduplicated native agent lists.
  - Keep native topology and agent handles internal to AMDGPU.
  - Preserve separate host and agent-visible addresses for imported host memory.
  - Remove the AMDGPU-private scoped-protection entry point.
  - Keep all topology resolution and native access updates off the dispatch hot path.

  ## HIP functionality enabled

  This supports the backend requirements for:

  - hipMemSetAccess and hipMemGetAccess
  - Host-only and device-only VMM permissions
  - Managed and registered host memory
  - Multi-device host-visible allocations
  - VMM-backed asynchronous allocation and memory pools

### DIFF
--- a/libhrx/src/libhrx/allocator.c
+++ b/libhrx/src/libhrx/allocator.c
@@ -284,5 +284,6 @@ hrx_status_t hrx_allocator_virtual_memory_protect(
       allocator->hal_allocator, virtual_buffer->hal_buffer,
       (iree_device_size_t)virtual_offset, (iree_device_size_t)size,
       /*queue_affinity=*/IREE_HAL_QUEUE_AFFINITY_ANY,
+      IREE_HAL_VIRTUAL_MEMORY_ACCESS_SCOPE_ALL,
       (iree_hal_memory_protection_t)protection));
 }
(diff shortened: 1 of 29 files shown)


benvanik:
Closing this older access-scope carrier in favor of #451. That PR contains the newer portable VMM execution-access-scope extraction, preserves the same-device ownership boundary, and now includes the complete record/replay lifecycle. Any remaining native pointer-import or foreign-ownership requirement from this branch should be extracted separately rather than keeping the mixed carrier open.

PR #372, opened 2026-08-14
[Loom] Preserve target-low facts across AMDGPU ABI lowering

benvanik:
AMDGPU HAL ABI materialization now preserves verified constraints on direct kernel arguments when it replaces signature values with kernarg loads. Downstream target-Low passes retain exact values, ranges, and relations for operand-form selection and other fact queries instead of silently losing them at the ABI boundary.

The reported Qwen RMSNorm compilation exposed the underlying contract violation: direct arguments were removed from the target-Low function while function predicates could still refer to their deleted SSA values. A stronger code-generation witness makes the consequence visible. Given `eq(step, 3)`, operand-form selection can replace `s_add_u32(position, step)` with `s_addk_i32(position, 3)` before ABI materialization, but the real ABI-first prepared-Low pipeline previously lost that fact and retained the generic register-register add.

## Carry facts as target-Low values

This introduces `low.assume`, a target-neutral register identity carrying predicate constraints into target-Low IR. Its results have the same types and exact physical storage as their corresponding operands. Fact inference applies unary and relational predicates to those result values, while parsing, bytecode, cloning, remapping, Python construction, legalization, scheduling, and allocation treat the operation as ordinary IR with explicit semantics.

`low.assume` is compile-time-only rather than a compiler hint. Allocation hard-aliases each result to its operand, packet-based targets omit it, and SSA-ID emitters map the result to the existing backend value. It therefore introduces no copy, instruction, or additional register storage. LLVM IR, SPIR-V, WebAssembly, IREE VM, and native AMDGPU emission all honor that contract.

Contradictory assumptions use the same target-legalization diagnostics as scalar and index assumptions. Register facts are normalized through their scalar element contract before conflict and topology checks, so impossible target-Low input fails at the established diagnostic boundary rather than reaching planning.

## Preserve facts at the AMDGPU ABI boundary

AMDGPU HAL materialization snapshots function predicates while the authoritative direct-argument-to-kernarg-value mapping still exists. It materializes only arguments with real body or type uses, translates every fully representable direct-argument predicate onto the loaded values, and builds one `low.assume` over the unique participating values before replacing uses and removing the signature.

Batching the transferred predicates is semantically important: the shared fact engine sees the complete predicate set at once, so source ordering cannot prevent a relation from consuming bounds established by a later predicate. Predicates over values that otherwise disappear do not make those arguments artificially live or introduce dead kernarg loads.

The exact `step == 3` witness now selects `s_addk_i32` through the production ABI-first pipeline. The original Qwen case has a predicate-only argument at this boundary, so both the unusable predicate and unnecessary load disappear while its generated artifact remains unchanged.

## Enforce the IR ownership contract

Block argument removal now rejects values referenced from predicate attributes in addition to ordinary operands and type uses. Signature-erasing transforms must transfer, rewrite, or remove those contracts explicitly; they can no longer leave dangling predicate metadata behind.

SPIR-V does not require an equivalent ABI transfer today. Its shader-entry and raw-BDA paths retain the original Low arguments through optimization and bind those same value IDs to emitted descriptor or push-constant loads only during final binary emission. Explicit `low.assume` values are still supported there as zero-cost SSA aliases.

## Reviewer Notes

The highest-value review surfaces are the storage-identity and compile-time-only semantics of `low.assume`, the batched predicate transfer before AMDGPU signature erasure, and the generic predicate-use guard in block argument removal.


benvanik:
gfx942 timeout is a flake - could be us, but 1000 locked runs on a real non-rocm-CI machine could not reproduce.

PR #380, opened 2026-08-17
[HIP] Define stream and event runtime contracts

AWoloszyn:
Defines the HIP stream and event runtime contracts around validated opaque handles, explicit ownership, concurrent lifetime safety, and stream-ordered synchronization.

 ## Changes
  - Validates stream and event handles before dereferencing them and retains live objects while API operations are in progress.
  - Resolves explicit streams through their owning contexts instead of implicitly creating or selecting an unrelated current context.
  - Retires public handles safely during stream destruction and context reset while preserving in-flight operations.
  - Implements stream identifiers, attributes, ordered value reads and writes, event flag validation, and stable callback names.
  - Orders legacy default-stream dependencies through device timeline barriers instead of host synchronization.
  - Detects timeline exhaustion and updates stream submission state only after successful queue submission.
  - Rejects compute-unit-masked stream creation with hipErrorNotSupported because the current queue submission path cannot enforce a per-stream mask.
  - Reports the active device compute-unit mask for supported mask queries instead of retaining unenforced per-stream state.

  ## Correctness
  The implementation makes stream and event lookup concurrency-safe, preserves context provenance, avoids exposing partially submitted timeline state, and returns explicit errors for behavior the runtime cannot currently enforce.

  ## Validation

  Focused stream and event coverage passes with 31 passing cases and no failures, crashes, or hangs. Masked stream creation returns a clean unsupported result rather than silently accepting an ineffective mask. The focused HRX build, unit tests, formatting, generated CMake validation, text hygiene, and libhrx hygiene checks pass.


zjgarvey:
Reviewed `d9c5c6a1b` against its base `8c274e544`, tracing the changed call paths and checking behavior against upstream ROCm. Line references are to the PR head. Findings below are defects only; severity is consequence, confidence is whether the mechanism was traced end to end. Anything the diff did not cause is separated into the out-of-scope section at the end.

| # | Sev | Conf | Issue | Location |
|---|---|---|---|---|
| 1 | HIGH | HIGH | Launch on a stream owned by another context now fails | `api.c:12886`, `stream.c:562` |
| 2 | HIGH | HIGH | New `hipStreamWriteValue*` under capture records a pointer to a dead stack slot | `api.c:11216`, `memory.c:2247` |
| 3 | HIGH | HIGH | `hipStreamDestroy` no longer detaches `stream->context` | `api.c:10619`, `stream.c:357` |
| 4 | HIGH | MED | New resolver drops the context pin the base held, behind a check that reads as a guarantee | `api.c:1663`, `api.c:1668` |
| 5 | MED | HIGH | Launch paths collapse five error codes into `hipErrorInvalidValue` | `api.c:12819`, `api.c:12963` |
| 6 | MED | HIGH | `hipStreamSetAttribute` returns `hipSuccess` for a policy nothing enforces | `api.c:10802` |
| 7 | MED | HIGH | `hipStreamWaitEvent` now rejects `hipEventWaitExternal`; upstream accepts it | `api.c:11069` |
| 8 | MED | HIGH | New value-write path has no legacy ordering; two ordering mechanisms now coexist | `api.c:11186`, `api.c:6584` |
| 9 | MED | HIGH | `hip_api_name_test` passes when it tests nothing, and silently tests ROCm when run outside its target | `hip_api_name_test.cc:33` |
| 10 | MED | HIGH | Process-global mutex added to the per-launch path | `handle_registry.c:168` |
| 11 | MED | HIGH | Dead `bool` return; two new/invalidated header comments contradict their code | `internal.h:1549`, `internal.h:1597` |
| 12 | MED | HIGH | Registry tests never exercise concurrency, and no test rejects a stale handle through a public entry point | `handle_registry_test.cc` |
| 13 | MED | MED | `wait_blocking_streams` submits O(streams) barriers where the API takes a list | `context.c:1034` |
| 14 | MED | MED | Async-free reclamation no longer fires on the ordering path | `stream.c:556` |
| 15 | MED | MED | Memory-reuse memo is now load-bearing for ordering, and its key collides | `stream.c:580`, `context.c:632` |

---

### 1. HIGH / HIGH — kernel launch on a stream from another context now fails

`hipLaunchKernel` takes its context from the calling thread at `api.c:12809` and its stream from the caller's handle at `api.c:12816`, then passes that pair to `iree_hip_order_legacy_stream_dependencies` at `api.c:12886`. That helper now calls `iree_hal_streaming_stream_wait_stream(stream, context->default_stream)` (`api.c:1753-1754`), whose new guard at `common/stream.c:562-565` rejects the pair with `"stream dependency crosses contexts"`. `hipModuleLaunchKernel` has the same shape at `api.c:13108`, `:13115`, `:13180`. At the base the same helper called `iree_hal_streaming_stream_synchronize(context->default_stream)`, which host-blocked on the caller's default stream and returned `hipSuccess`; no cross-context predicate existed on that path.

Trigger: `hipSetDevice(0); hipStreamCreate(&s0); hipSetDevice(1); hipLaunchKernel(..., s0);` — every launch now fails with `hipErrorInvalidValue` at the ordering step, before the dispatch, where the base had no predicate to fail. More generally, any launch whose stream belongs to a context other than the calling thread's current context.

The owning context is already in hand a few lines earlier: `api.c:12869` passes `stream_obj->context->device_entry` to launch validation, which tolerates the mismatch. `hipMemcpyAsync` sources the context from the stream; the launch paths source it from the thread. The PR description says the change "[r]esolves explicit streams through their owning contexts instead of implicitly creating or selecting an unrelated current context" — that is exactly what the two launch entry points still do not do.

Fix: source the context from the stream for explicit handles, falling back to the current context only for the sentinels.

### 2. HIGH / HIGH — `hipStreamWriteValue32/64` under graph capture records a pointer to a dead stack slot

`iree_hip_enqueue_stream_value_write` (`api.c:11186`) takes `value` **by value** and passes `&value` to `iree_hal_streaming_memcpy_host_to_device` at `api.c:11216-11218`. Outside capture that is safe because the copy happens at record time. Inside capture the source pointer is stored verbatim at `common/memory.c:2247` (`callback_data->src = src;`) into an arena node that outlives the call, and dereferenced at every graph launch by `common/memory.c:51` (`memcpy(callback_data->dst, callback_data->src, callback_data->count);`).

Trigger: begin capture, `hipStreamWriteValue64` on a device-only buffer with default flags, end capture, instantiate, launch. The callback reads whatever now occupies that stack region. Silent wrong data on every replay.

Newly reachable: both entry points were `hipErrorNotSupported` stubs at the base and this helper did not exist. `common/memory.c` is byte-identical between the two revisions, so the fix belongs in the new helper, not in the capture path — and the sibling branch of the same function already does it correctly, heap-allocating and copying `value` when `host_ptr` is present (`api.c:11219-11242`, the copy at `:11230`).

Fix: copy the value into storage that outlives the node, as the `host_ptr` branch does.

### 3. HIGH / HIGH — `hipStreamDestroy` no longer detaches the stream from its context

The base ended `hipStreamDestroy` with `streaming_stream->context = NULL;` immediately before the final release. The diff deletes that line; the head unregisters at `api.c:10619` but never clears the back-pointer. The destructor branches on that field at `common/stream.c:357-366` and, when it is non-NULL, calls `iree_hal_streaming_context_unregister_stream(context, stream)`, which locks `context->stream_list_mutex` (`common/context.c:674`). Context teardown only detaches streams still in the list (`common/context.c:205-214`), so a stream already unlinked by `hipStreamDestroy` never gets its back-pointer cleared.

Trigger: thread A is inside an API call holding the reference the registry hands out; thread B calls `hipStreamDestroy(s)`, then `hipDeviceReset()`. Teardown skips `s` because it is no longer listed, and frees the context. Thread A drops the last reference; the destructor takes the `if (context)` branch on a dangling pointer and locks a mutex inside freed memory. No racy read required, only that ordering.

What made the base safe is the line this diff deletes: clearing the back-pointer before the release (base `api.c:10389-10390`) meant the destructor skipped the `if (context)` branch regardless of which reference happened to be last — and other references do exist, all pre-existing and untouched here: the context's stream list (`common/context.c:640`), every stream snapshot (`common/context.c:748`), and an event's recording stream (`common/event.c:108`). This diff deletes that clear and adds a fourth source of outstanding references (`iree_hip_stream_lookup_retain`, `api.c:1629`), so the branch now runs on a stream whose context may already be freed.

Fix: restore `streaming_stream->context = NULL;` before the final release at `api.c:10623`.

### 4. HIGH / MED — the new resolver drops the context pin the base held, behind a check that reads as a guarantee

`iree_hip_resolve_registered_stream` (`api.c:1657`) checks `(*out_stream)->context` at `api.c:1668-1671` and returns `hipErrorContextIsDestroyed` when it is NULL. That check is new, and it reads at the call site as a liveness guarantee it cannot provide: `iree_hal_streaming_stream_t::context` is an unowned back-pointer by design (`common/internal.h:486`, "Parent context, unowned (to avoid cycles)"), so retaining the stream retains nothing of the context.

The delta that matters is what the explicit-handle branch stopped doing. Base `iree_hip_resolve_stream` called `iree_hip_ensure_context(&context)` unconditionally on every path including explicit handles, and `iree_hal_streaming_context_set_current` retains, so a thread resolving a stream ended up holding a TLS reference to a context — the stream's own, in the ordinary single-context case. The new branch calls only `iree_hip_ensure_initialized()` (`api.c:1663`). A thread with no current context that passes an explicit handle now holds nothing while the code immediately dereferences the field: `api.c:10884` (`stream_obj->context->device_ordinal`), `api.c:11202` (`stream_object->context` into the range lookup, new code), `api.c:11367` (`stream_object->context->device_entry`).

Trigger: thread A passes the check; thread B completes `hipDeviceReset`, NULLing the field at `common/context.c:208` and freeing the context; thread A proceeds into `context->device` or `context->stream_list_mutex`. Confidence is MED because the base's protection was incidental rather than contractual — it held only when the thread's implicit context was the stream's.

Fix: have the stream retain its context, or return a retained context alongside the retained stream so the check means what the call site reads it as meaning.

---

### 5. MED / HIGH — launch paths collapse five error codes into `hipErrorInvalidValue`

`api.c:12819` replaces the base's `HIP_RETURN_ERROR(init_result)` with a literal, and the rewritten `hipExtLaunchKernel` does the same at `:12963`, `:12972`, and `:12983`. Every code the resolver can produce is erased: `hipErrorNotInitialized` and `hipErrorNoDevice` from `iree_hip_ensure_initialized` (`api.c:1468-1469`, `:1475`), `hipErrorOutOfMemory` from the sentinel branch (`api.c:1513`, `:1565`), and the two the resolver computes itself and the caller then discards — `hipErrorInvalidResourceHandle` (`api.c:1666`) and `hipErrorContextIsDestroyed` (`api.c:1671`). All five reach the application as `hipErrorInvalidValue`. The base propagated whatever the resolver returned (base `api.c:12216-12219`); the head discards it on the hottest entry point, in the change whose purpose is precise handle errors — the two codes this PR adds are the two it then throws away.

### 6. MED / HIGH — `hipStreamSetAttribute` replaces a truthful error with silent success

`hipStreamSetAttribute` (`api.c:10788`) stores `value->syncPolicy` at `api.c:10802-10803` and returns `hipSuccess`, replacing a stub that returned `hipErrorNotSupported`. Nothing in the tree reads the field: `synchronization_policy` appears only in its declaration (`common/internal.h:492`), its initialization to `..._POLICY_AUTO` (`common/stream.c:305`), and the get/set/copy entry points. Upstream enforces it. This is the opposite of the call made in the same commit for CU masks, where `hipExtStreamCreateWithCUMask` (`api.c:11342-11353`) returns `hipErrorNotSupported` precisely so it does not "expose a stream that violates its contract".

### 7. MED / HIGH — `hipStreamWaitEvent` now rejects a flag upstream accepts

The `if (flags != 0)` reject at `api.c:11069-11072` is new; the base had no flags predicate and fell straight through to `iree_hal_streaming_stream_wait_event`. So the base accepted `hipEventWaitExternal` (0x01) by ignoring it and the head returns `hipErrorInvalidValue`. Upstream accepts it (`hipamd/src/hip_stream.cpp:453`, which permits `hipEventWaitDefault` and `hipEventWaitExternal` and rejects everything else) and gives it real behavior at `:473-482`, adding an external-event wait node to the capture graph. The constant is not defined anywhere in this tree, so a caller including ROCm's header and this runtime gets a hard failure on a documented flag.

### 8. MED / HIGH — the new value-write path has no legacy ordering, and two ordering mechanisms now coexist

`iree_hip_order_legacy_stream_dependencies` has exactly three call sites (`api.c:6608`, `:12886`, `:13180`). The newly implemented `hipStreamWriteValue32/64` path (`iree_hip_enqueue_stream_value_write`, `api.c:11186-11248`) is not one of them and enqueues with no ordering at all. Write-value exists for producer/flag handoff, so unordered submission against the legacy stream means wrong results, not just weak ordering.

Separately, converting the shared helper from a host block to device barriers left two mechanisms in the tree: null-stream launches now order by queue barrier (`api.c:12886`), while null-stream `hipMemcpyAsync` still host-blocks first through its own prologue (`api.c:6582-6585`, unchanged from the base). At the base both went through one helper that host-blocked, so the two agreed.

### 9. MED / HIGH — `hip_api_name_test` is a false green

`hip_api_name_test.cc:33` calls `GTEST_SKIP()` when `dlopen` fails, so all three cases skip and the target reports PASSED — the test whose job is "the library loads and exports `hipApiName`" turns green exactly when it does not load. With the env var absent — which happens whenever the built binary runs outside its test target, since both `BUILD.bazel:189-191` and the CMake target (`CMakeLists.txt:221-222`) point `HRX_TEST_LIBAMDHIP64` at the artifact under test — and `HRX_TEST_LIBAMDHIP64_PATH` undefined, as it is under Bazel, the bare-soname fallback at `:24` loads the system `/opt/rocm-*/lib/libamdhip64.so` and two of three cases pass against AMD's runtime instead of this one. Both branches were reproduced.

### 10. MED / HIGH — process-global mutex on the per-launch path

`iree_hip_handle_registry_lookup_retain` (`handle_registry.c:160-184`) takes one process-global `iree_slim_mutex_t` at `:168` per validated call and runs `retain_fn(handle)` inside it at `:175`. Measured on the real table and mutex: 8.7 ns uncontended, 82 ns at 2 threads, 166 ns at 4, 491 ns at 8. The same table per-thread stays at 22 ns; striped 16 ways, 37 ns — so 90-95% of the contended cost is the single lock, not the validation. Kept at MEDIUM because `iree_hip_ensure_initialized` already takes a global mutex unconditionally (`api.c:1450`) and the launch path enters it twice, so this adds a third process-global acquisition to a path that already serialized rather than introducing the class. See the closing note for a cheap offset.

### 11. MED / HIGH — dead return value, and two header comments that contradict their code

`iree_hal_streaming_context_unregister_stream` gained a `bool` return, documented at `common/internal.h:1549-1551` as letting "public handle destruction atomically reject stale or concurrently destroyed handles". All three callers discard it (`common/stream.c:365`, `api.c:1652`, `api.c:10619`); the rejection the doc describes is actually performed by `iree_hip_handle_registry_remove` via `iree_hip_stream_unregister_public_handle` (`api.c:1613-1627`).

Two header comments are wrong as of this diff. `common/internal.h:1592-1593` is new and names two exclusions ("Non-blocking streams and |stream| are excluded"); the loop applies five, also skipping NULL entries, `context->default_stream`, and any capturing stream (`common/context.c:1026-1032`). `common/internal.h:1597` ("Queries whether any stream in the context still has queued work.") was true at the base and is invalidated by this diff, which adds a non-blocking skip to `iree_hal_streaming_context_query` at `common/context.c:1055-1059`.

### 12. MED / HIGH — the registry tests do not defend the invariant the design rests on

`handle_registry_test.cc` has three cases, all single-threaded: no test creates a thread, and no test anywhere drives a stale handle through a public entry point to check that the API rejects it instead of dereferencing it. Twelve defects were injected into `handle_registry.c` and each built and run. Caught: tombstone removal, lookup matching tombstones, rehash dropping entries, duplicate rejection, remove-returns-true, and disabled growth (as a 60 s timeout). **Not caught, all 3/3 passing:** removing every mutex lock/unlock; moving `retain_fn` outside the lock; tombstones never reused; `tombstone_count` never decremented; identity hash. `rehash` forgetting `count = 0` is caught only by `IREE_ASSERT(registry->count == 0)` in `iree_hip_handle_registry_deinitialize` (`handle_registry.c:95`), which the `Default` CMake variant compiles out (`assertions: false`); the ASAN, TSAN and UBSAN variants set assertions on, so it is live there.

Deleting all locking from a structure whose reason for existing is concurrent access does not fail its suite, and neither does moving the retain outside the lock — the single property that makes the design safe. Existing context worth knowing while sizing the suite: this suite does build and run under the ASAN, TSAN and UBSAN CMake variants, but a single-threaded suite gives TSAN nothing to observe, and the end-to-end HIP paths that would exercise handle validation on a device are `runtime-resource=`-labelled and so excluded from those jobs (`.github/workflows/ci_core_linux.yml:95`), so no sanitizer sees a concurrent lookup anywhere. `libhrx/cts/tests/hip/event_test.cpp` is the natural fixture for the end-to-end half; it passes 21/21 at this head but none of its cases touch handle validation.

### 13. MED / MED — O(streams) barriers where the API takes a list

The per-stream loop in `iree_hal_streaming_context_wait_blocking_streams` (`common/context.c:1024-1035`) is pre-existing; converting its body from `iree_hal_streaming_stream_synchronize` to `iree_hal_streaming_stream_wait_stream` (`:1034`) is what changes the cost. Each call flushes both streams, reserves a timeline value, and issues its own `iree_hal_device_queue_barrier` plus `iree_hal_device_queue_flush` (`common/stream.c:596-635`). Sixteen blocking streams produce sixteen submissions and sixteen destination timeline values, and `iree_hal_device_queue_barrier` already accepts a semaphore list, so one submission would do.

### 14. MED / MED — deferred async-free reclamation no longer fires on the ordering path

Reclaiming completed `hipFreeAsync` allocations was a side effect of the `iree_hal_streaming_stream_synchronize` this PR replaced on the legacy-ordering path: that function calls `iree_hal_streaming_memory_release_completed_async_frees` (`common/stream.c:721`), and the new `iree_hal_streaming_stream_wait_stream` (`common/stream.c:556-651`) does not. Reclamation still happens on the explicit synchronize paths, so this is a leak in shape rather than in total: a workload that only ever orders against the legacy stream and never calls `hipStreamSynchronize` now accumulates pending frees indefinitely.

### 15. MED / MED — the ordering memo is now load-bearing, and its key collides across contexts

`wait_stream` consults the memory-reuse table at `common/stream.c:580-584` and returns before submitting anything on a hit, so that table now decides whether an ordering barrier is emitted. The table, its helpers, and its key are pre-existing and unchanged: `stream_id` is built from `device_ordinal` at `common/context.c:632-634`, so two contexts on one device mint identical ids. At the base a collision could only cause premature memory reuse; now a cross-context entry can suppress a barrier that was required. Either widen the key to identify the context, or key the memo by the stream pointer.

---

### Lower severity

- `iree_hal_streaming_stream_wait_stream`'s header (`common/stream.h:18-20`) states neither of the two things a caller needs: that the call flushes both streams on the calling thread (`common/stream.c:571-572`, `:596`), and that the destination must not be capturing — an invariant enforced entirely by its callers (`api.c:1741-1744`; `common/context.c:1026-1032` additionally excludes capturing sources) while the callee checks nothing, so a capturing destination would get a real device barrier submitted behind the graph. Its semaphore list is also built from three parallel ternaries on one predicate (`common/stream.c:612-620`) where the adjacent `wait_event` uses an accumulator that cannot desynchronize (`common/stream.c:890-907`).
- `hipStreamBatchMemOp` (`api.c:11305-11316`) validates the stream handle before its arguments; upstream reverses that order (`hipamd/src/hip_stream_ops.cpp:14-20`), so a call with both a bad stream and bad arguments yields a handle error here and `hipErrorInvalidValue` upstream.
- `hipStreamGetAttribute` still does not serve `hipStreamAttributePriority`. The base returned `hipErrorNotSupported`, the head returns `hipErrorInvalidValue` (`api.c:10771-10773`); upstream's switch handles it (`hipamd/src/hip_stream.cpp:890`). Only the error code changed, and the new one is arguably worse because it no longer says "unimplemented" — worth closing while the entry point is being written.
- The new API-name table records the ABI shape but no ROCm version (`api_metadata.c:11-13`), and its last entry is `[479] = "hipLibraryGetManaged"` (`:480`) while upstream's namespace runs to 483 (`HIP_API_ID_LAST = 483`, ids 480-483 being the discard-batch family); `amdhip64.map` by contrast names its source in its own header (`amdhip64.map:2`), which is the minimum the table is missing.
- Disclosure gap rather than a defect: `hrx_hip_spt_default_stream` (`abi_stubs.c:394-401`) stops creating a private thread-local stream with `hipStreamNonBlocking` and returns the `hipStreamPerThread` sentinel instead. That is correct — it matches upstream, whose per-thread stream is created with `hipStreamDefault`, and it removes a per-thread `hipStream_t` that the base never destroyed. But it changes implicit ordering for the 29 `_spt` entry points that resolve the per-thread stream (23 through `hrx_hip_spt_stream_or_explicit`, 6 direct, out of 35 `_spt` entry points), and it appears in neither the commit message nor the PR description, and in no test.

### Description accuracy

"Detects timeline exhaustion and updates stream submission state only after successful queue submission" describes `iree_hal_streaming_stream_reserve_next_value_locked`, which is byte-identical between the two revisions — the `"stream timeline value overflow"` check is at head `common/internal.h:563` and base `common/internal.h:569`, the six-line net shift being the `cu_mask` fields the diff deletes from the struct above. The submission-state half is likewise pre-existing, at base `common/stream.c:827-828` inside `wait_event`; the diff's only writes to `pending_value` are the copy of that pattern inside the new `wait_stream`. The only `IREE_STATUS_RESOURCE_EXHAUSTED` this PR adds are the two hash-table capacity errors in `handle_registry.c`. The new path obeys an existing contract, which is correct, but does not introduce it.

"Focused stream and event coverage passes with 31 passing cases" is not reproducible from the commit. This PR adds six cases across two new targets (`handle_registry_test.cc` and `hip_api_name_test.cc`, three each), neither of which covers streams or events, and no stream or event test target is added or modified anywhere in the diff. No single committed target reports 31 cases, and the stream-and-event groupings give 29 (`stream` 8 + `hip_event_test` 21), 34 (plus `stream_ops` 5), or 35 (plus `hip_device_event_sink_test` 1); reaching 31 requires pulling in a target that is neither.

---

### Out of scope

The behavior each of these describes predates `8c274e544` and is not changed by this diff. Listed because the migration this PR starts is the natural place to finish them, not as defects introduced here.

- **Stale handles still segfault on the unconverted half of the API.** 23 call sites still use `iree_hip_resolve_stream` (raw cast, borrowed pointer) against 19 using the new validated resolver, and the base had 33 of the former. The split is not principled: `hipLaunchKernel` validates, `hipModuleLaunchKernel` (`api.c:13115`) does not; `hipStreamSynchronize` validates, the whole capture family does not. Measured at this head on an MI300X, one process per case, on a destroyed stream: `hipMemsetAsync` (`api.c:9900`) SIGSEGV; `hipMemcpyAsync` SIGSEGV; `hipStreamBeginCapture` (`api.c:21918`) SIGSEGV; `hipMallocAsync` (`api.c:23358`) SIGSEGV; `hipStreamIsCapturing` (`api.c:22236`) returns `hipSuccess` after a use-after-free read. `hipMemsetAsync` also SIGSEGVs on `0xdeadbeef`. Single-threaded and deterministic: `hipStreamCreate(&s); hipStreamDestroy(s); hipMemsetAsync(p, 0, n, s);` The two resolvers differ by one word and neither carries a doc comment stating its ownership contract, which is worth fixing even before the remaining sites are converted.
- **Streams never retain their parent context.** `iree_hal_streaming_stream_t::context` is unowned at the base too (`common/internal.h:486`), and the base has the same dereference-after-borrow shape on the same route. Events already retain their context. Making the stream retain its context closes finding 4 and this together.
- **Context teardown mutates the stream list without holding `stream_list_mutex`.** `common/context.c:205-214` writes `stream->context = NULL` and swap-removes entries outside the lock every other mutator takes.
- **`hipMemsetAsync` orders against nothing.** It reaches `iree_hal_streaming_memory_memset` (`api.c:9906`) with no legacy-stream ordering at either revision, so it is the fourth entry point that arguably belongs in finding 8's list.

One observation, not a request: the launch path takes the process-global init mutex twice per launch and did so at the base too. `hipLaunchKernel` calls `iree_hip_ensure_context` at `api.c:12809`, and the stream resolver's first act at `api.c:1663` (or `:1677` for the sentinels) is another trip into `iree_hip_ensure_initialized`, which locks `iree_hip_global_init_mutex` unconditionally with no already-initialized fast path (`api.c:1448-1477`). Removing that second acquisition would roughly offset the cost of the new registry lock in finding 10.

zjgarvey (commented):
The following graph APIs still raw-cast public `hipEvent_t` values before proving the handle is live:

- `hipGraphAddEventRecordNode` (`libhrx/src/binding/hip/api.c:18067-18087`): checks only that `event` is non-null, then casts it to `iree_hal_streaming_event_t*`. `iree_hal_streaming_graph_add_event_node` dereferences `event->context` at `libhrx/src/binding/common/graph.c:1769` and retains the pointer at `graph.c:1790`.
- `hipGraphAddEventWaitNode` (`libhrx/src/binding/hip/api.c:18101-18121`): same raw-cast path for wait nodes.
- `hipGraphEventRecordNodeSetEvent` (`libhrx/src/binding/hip/api.c:20174-20179`) and `hipGraphEventWaitNodeSetEvent` (`api.c:20210-20215`): cast `event` and read `streaming_event->context` before registry validation.
- `hipGraphExecEventRecordNodeSetEvent` (`libhrx/src/binding/hip/api.c:20281-20283`) and `hipGraphExecEventWaitNodeSetEvent` (`api.c:20307-20309`): pass the raw event pointer to `iree_hal_streaming_graph_exec_set_event_node_event`, which retains it at `libhrx/src/binding/common/graph_exec.c:991`.

`hipEventDestroy` removes the public handle from the live-event registry at `libhrx/src/binding/hip/api.c:11635-11641`, so these APIs should reject a destroyed `hipEvent_t` before any cast, context check, or retain. Today they can accept a destroyed handle that happens to still be internally retained, or dereference freed event storage.

Suggested fixes: route each public graph-event argument through `iree_hip_event_lookup_retain`, use the retained event for context checks and graph mutation, and release the API operation reference after the graph/node/exec takes its own reference.

zjgarvey (approved):
I wasn't able to find anything major on the last round of reviews.

PR #398, opened 2026-08-22
[amdgpu] Import an allocation this topology can reach but does not own

Geramy:
Device allocation import refuses any pointer whose owning HSA agent is outside the importing device's logical topology. Two GPUs opened as two logical devices are never in each other's topology, so a buffer allocated by one cannot be imported by the other even when peer access has been granted for it — which leaves a copy through host memory as the only way for one logical device to read another's buffer, on hardware where the link supports the read directly.

Ownership is not the question that decides whether a dispatch may bind an allocation; reachability is. This asks ROCr for the pointer's accessible-agent list and admits the import when every GPU agent in the importing topology appears in it, keeping the refusal for memory that genuinely cannot be reached.

An imported allocation the device does not own stays `DEVICE_LOCAL`: only an owned one may be promoted to host visible on the strength of its own SVM configuration, which is a property of the owner rather than of the mapping.

### Testing

Two gfx1201 GPUs in one process, opened as separate logical devices. Before: `hrx_allocator_import_buffer` fails with `PERMISSION_DENIED; device allocation is owned by an HSA GPU agent outside the AMDGPU HAL logical topology`. After: the import succeeds and the imported range reads back the bytes the owning device wrote.

### DIFF
--- a/libhrx/cts/core/hrx_loader.cpp
+++ b/libhrx/cts/core/hrx_loader.cpp
@@ -133,6 +133,8 @@ void HrxLoader::load(const std::string& path) {
   LOAD(semaphore_signal);
 
   LOAD(stream_create);
+  LOAD(stream_create_on_queue);
+  LOAD(stream_get_queue_affinity);
   LOAD(stream_retain);
   LOAD(stream_release);
   LOAD(stream_synchronize);
--- a/libhrx/cts/core/hrx_loader.hpp
+++ b/libhrx/cts/core/hrx_loader.hpp
@@ -91,6 +91,8 @@ class HrxLoader {
 
   // Streams.
   decltype(&hrx_stream_create) stream_create;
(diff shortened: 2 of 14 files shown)


benvanik:
You should not be opening those as separate logical devices. Much of the system requires knowing they are two physical devices of the same type. Is there a reason you are doing that?

(also, beware that you have random other changes such as loom stuff - please treat those separately for review)

Geramy:
> You should not be opening those as separate logical devices. Much of the system requires knowing they are two physical devices of the same type. Is there a reason you are doing that?
> 
> (also, beware that you have random other changes such as loom stuff - please treat those separately for review)

I just now have realized that it seems you create a logical device instead I will actually revisit this PR and review it a second time. This is the first time I've used and seen HRX so it's a bit new to me, perhaps next time I'll wait a day or two before I submit a PR. Thanks for responding so quickly though! @benvanik

PR #401, opened 2026-08-22
[Loom] Materialize unique launch configurations

benvanik:
Model-scale command programs frequently schedule hundreds or thousands of dispatches while reusing only a handful of launch configurations. The static command planner previously cloned the complete configuration body for every dispatch and asked canonicalization and CSE to rediscover that repetition afterward. The final program was compact, but JIT construction time and retained arena storage had already scaled with dispatch count.

This change makes exact launch identity part of schedule consumption. An identity is the concrete kernel definition plus its ordered workload SSA values: equal ranges or similar facts do not imply equality, while repeated uses of the same kernel and SSA tuple do. The planner assigns each scheduled launch an identity ordinal during the traversal it already performs, materializes each unique configuration body once, and projects the resulting direct count or dynamic tuple ordinal back onto every launch row.

The representation is adaptive for the JIT hot path. Up to four unique identities use bounded linear matching and uninitialized stack scratch regardless of dispatch count. A fifth unique identity promotes the set to an open-addressed arena table. Once body construction is complete, tuple compaction reuses those same slots instead of allocating another table, and the scratch result storage is reused when rebuilding the compact return. There are no string identities, IR markers, side dictionaries, or additional module traversals.

On a production-shaped root with 128 dispatches, two configuration bodies, and three exact identities, body construction falls from 1,027 value rows and 643 live operations to 27 value rows and 18 operations. Retained module storage falls from 1,283 value rows to 33. Both paths finish with the same 13 live operations, nine host results, launch order, wave structure, direct counts, and dense tuple ordinals.

Direct unique-body materialization is deliberate for this static product. Building private pure helpers and calls before generic inlining produced the same final body but doubled retained value rows and added planning cost. VM command programs can retain ordinary pure calls because they execute those helpers rather than constructing and immediately erasing an intermediate call graph.

Production-path Loom coverage includes the exact 128-dispatch sharing case and a ten-identity case that crosses both table-growth boundaries. Existing command-product corpus cases continue to pin launch placement and serialized scheduling behavior.

### DIFF
--- a/loom/src/loom/target/arch/cmd/lower/launch_graph.c
+++ b/loom/src/loom/target/arch/cmd/lower/launch_graph.c
@@ -34,37 +34,126 @@ typedef struct loom_cmd_launch_graph_producer_frame_t {
   uint16_t next_operand;
 } loom_cmd_launch_graph_producer_frame_t;
 
-typedef struct loom_cmd_launch_graph_build_t {
+// Non-recursive pure-producer traversal stack.
+typedef struct loom_cmd_launch_graph_producer_stack_t {
+  // Active frames in dependency traversal order.
+  loom_cmd_launch_graph_producer_frame_t* frames;
+  // Number of active frames in |frames|.
+  iree_host_size_t count;
+  // Allocated entry capacity of |frames|.
+  iree_host_size_t capacity;
+} loom_cmd_launch_graph_producer_stack_t;
+
+// Final placement shared by launches with the same exact configuration
+// identity.
+typedef struct loom_cmd_launch_graph_placement_t {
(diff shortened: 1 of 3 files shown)


benvanik:
Closing this draft because the implementation does not meet the design and evidence bar. The structural observation remains valid: launch-configuration construction should not scale with repeated dispatch count. This version, however, couples an unmeasured small-table cutoff to several policies, tests private container boundaries, and does not establish the missing launch-plan ownership boundary or production Qwen qualification. A replacement should start from that ownership and evidence rather than incrementally preserve this implementation.

PR #403, opened 2026-08-22
[HAL/AMDGPU] Enable Windows execution across CMake and Bazel

benvanik:
Enables native AMDGPU HAL and Loom execution on Windows with a dynamically loaded ROCr shared runtime, while making clang-cl the default Windows host compiler and retaining an explicit MSVC lane. This carries forward the Windows HSA execution work explored in #386 and turns the platform constraints found there into explicit runtime and build-system contracts.

The AMDGPU runtime now models platform-owned capabilities independently from GPU ISA support. Windows queue doorbells use HSA notification, clock correlation uses HSA counters paired with QPC, queue authority is separated from host-accessible memory placement, feedback channels retain distinct host and device addresses, and pagefile-backed ring aliases preserve host wraparound when ROCr assigns a different GPU address. Missing SVM no longer rejects a system whose explicit HSA virtual-memory path remains usable.

Sanitizer shadows use sparse GPU-only mappings on Windows: virtual shadow space is reserved up front, while physical spans are committed only for materialized slabs. This preserves the production memory shape without premapping the full shadow or granting incidental CPU access.

The Windows build surface is first-class in both build systems. Bazel uses clang-cl by default, exposes an MSVC configuration, handles MASM and long linker responses, launches generated shell tests through `rules_shell`, and does not invoke or require CMake. CMake owns its nested integration tests in CTest, derives exact build roots for filtered selections, refreshes stale test graphs before enumeration, and propagates the selected generator and host compiler into nested fixtures. Repository clang-tidy follows the same boundary: Bazel and CMake independently link the IREE checks into a native analyzer because the supported Windows LLVM distribution cannot load out-of-tree modules.

Presubmit subprocesses now stream stdout and stderr as produced, synchronizing concurrent batches while retaining only a fixed 64 KiB tail per subprocess for failure recaps. Whole-tree native analysis therefore remains observable without memory use growing with phase output.

The pinned WASI SDK host tools are now available on Windows in both build systems. Bazel copies only the small `llvm-objdump.exe` payload into its output tree on Windows instead of executing it through an external-repository symlink whose NTFS metadata changes invalidated and refetched the 535 MiB SDK archive; Unix and macOS retain the symlink path.

Structured-loop allocation now preserves controls, captures, induction variables, and iteration arguments across backedges in both physical lifetime and sparse semantic liveness. This prevents loop-body temporaries from aliasing values required by the next iteration, and the Wasm assembly expectations reflect the resulting direct branch and result carries.

Full-platform qualification also exposed backend-independent correctness issues fixed here: Vulkan buffers now carry queue-family sharing and byte-granular transfers route away from dword-only native commands; SPIR-V physical-storage offsets and cooperative-matrix strides stay byte-addressed through byte-pointer operands; writable file handles can be resized without reopening paths; and benchmark/test helpers now fail according to their public setup and handle-selection contracts.

The HAL continues to load `hsa-runtime64.dll` rather than vendoring ROCr. The Windows shared-runtime and multi-signal-wait work remains in ROCm/rocm-systems#10379.

### Reviewer Notes

- The AMDGPU representation changes are concentrated in platform capabilities, virtual memory, feedback channels, access policy, and device-clock correlation.
- The Windows build changes deliberately preserve separate host and device compilers: clang-cl or MSVC builds host code, while ROCm clang builds AMDGPU device artifacts.
- Windows clang-tidy is an in-process executable because the supported LLVM prebuilt has no plugin exports; Bazel imports and links its static archives without crossing into CMake.
- The Vulkan and SPIR-V fixes are cross-platform semantic corrections discovered by running the complete Windows execution matrix, not Windows-only fallback paths.


benvanik:
Closing this omnibus draft now that its landed pieces have durable, independently reviewable history:

- #392 established Windows MSVC builds for Loom and the AMDGPU HAL.
- #395 made the native Windows CMake and Bazel workflows first-class.
- #415 added portable open-file resizing.
- #416 fixed Vulkan queue-family buffer sharing.
- #422 qualified the Windows host toolchains and developer/CI workflows.

The remaining Loom compiler correctness, profile-clock alignment, Vulkan command-buffer, and Windows AMDGPU runtime work is being split into separate workstreams and reconstructed from current `main` for landing. This PR remains as historical context for the original integrated experiment and will not be updated or landed as a unit.

PR #405, opened 2026-08-24
[HAL/Vulkan] Route byte transfers across queue families

benvanik:
Makes Vulkan buffer ownership and transfer-lane selection preserve the queue and byte-range contracts exposed by the HAL.

Buffers accessible from more than one queue family now use concurrent sharing with the exact normalized family set, while command buffers retain their normalized affinity so submission selects a compatible queue lane after bindings resolve.

Native Vulkan fill, update, and copy commands are used only for dword-aligned resolved ranges. Byte-granular direct, deferred, and storage-backed file transfers route through the compute-capable path instead of issuing invalid transfer commands.

The file CTS truncates its already-imported file through a new portable writable-handle resize operation. The operation uses the live descriptor or HANDLE and therefore preserves handle identity rather than reopening a path behind an active import.

Clock-correlation changes are intentionally excluded and reviewed separately.

### DIFF
--- a/runtime/src/iree/hal/drivers/vulkan/allocator.c
+++ b/runtime/src/iree/hal/drivers/vulkan/allocator.c
@@ -13,6 +13,7 @@
 #include "iree/base/threading/mutex.h"
 #include "iree/hal/drivers/vulkan/atomic.h"
 #include "iree/hal/drivers/vulkan/buffer.h"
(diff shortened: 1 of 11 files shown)


benvanik:
Superseded by the reviewable stack #415, #416, #417, #418, and #419. The old head branch remains intact because #406 is currently based on it; closing this PR removes the duplicate review surface without changing that dependency.

PR #424, opened 2026-08-25
[AMDGPU] Propagate asynchronous GPU faults through queue failures

zjgarvey:
## Summary

Propagate asynchronous AMDGPU memory faults and fatal system events through the existing IREE device, queue, semaphore, and status failure paths.

GPU faults can occur after a kernel submission has already returned successfully. Without native fault delivery, pending synchronization may block indefinitely and later operations may incorrectly appear to succeed. This change converts those native events into durable failures that callers can observe through normal HAL operations.

## Implementation

- Registers one process-wide HSA system-event callback, and installs it again after a complete HSA runtime shutdown, which destroys the runtime's callback registry.
- Pins the module implementing the callback into the process before handing its pointer to the HSA runtime, and refuses to register where no pin is available: the runtime keeps that pointer until its final shutdown and offers no unregister, so a module that can still be unloaded would leave it able to jump into unmapped code. Device creation therefore fails on platforms with no pin mechanism.
- Associates native GPU agents with their live IREE logical devices, returning an opaque registration stored on the logical device so partial initialization and exact removal are explicit.
- Claims only memory faults and hardware exceptions, and only when they were actually delivered to a live registration; memory error events, and events nothing can observe, are left to the HSA runtime's own abort. A fault arriving for a device whose frontier is already deassigned is therefore no longer claimed-and-dropped, and reaches that abort.
- Latches the first fatal error on the affected logical device.
- Fails every host queue of the affected logical device, whose queues share one epoch signal table spanning all of its GPU agents.
- Routes the HSA queue error callback, the process-wide system event callback, and the driver's own invalid-wait paths through one terminal queue transition, so every asynchronous fatal producer makes the same state change.
- Closes queue admission under the submission mutex before the final failure drain, so no epoch published later can escape it, and rejects later submissions on both the failed queue and the failed device.
- Wakes pending semaphore, frontier, and notification waiters with the failure.
- Binds fault delivery to host queue lifetime, publishing targets as the last step of frontier assignment and retiring them once every queue has passed its idle/error boundary, so a fault can still release the teardown wait it is the only thing able to release, and no event can reach a destroyed queue.
- Preserves the first failure when duplicate native notifications arrive.

## Testing

`system_event_test` covers delivery with no device: the registration path calls exactly one HSA entry point, so a libhsa whose only populated thunk is that registration captures the callback pointer and lets the test dispatch synthetic events into the same code the runtime calls. It covers the claim decision per event type, the agent match and fanout, registration removal, first-error-wins, registrations sharing an agent, handler re-registration after an HSA shutdown event, and every outcome of the module pin.

The only file touched outside the driver is one line of `runtime/src/iree/base/testing/BUILD.bazel`, making an existing `testonly` embedded shared object publicly visible so the pin test can load it; that package's hand-maintained `CMakeLists.txt` already declares it `PUBLIC`, so this brings the Bazel graph into line with it.

`host_queue_failure_test` covers the terminal transition on real hardware with work in flight: deferred and submitted operations both report the failure the queue recorded, an operation parked for submission capacity and resumed after admission closes reports it too, only the first failure is kept, a submission after the transition is rejected, teardown releases the recorded failure, and a queue whose hardware is parked behind a barrier it cannot pass still returns from its teardown wait once a failure is recorded.

## HIP Impact

This provides the asynchronous error delivery required for faulting kernel launches to be observed by operations such as:

- hipStreamSynchronize
- hipStreamQuery
- hipDeviceSynchronize
- event query and synchronization
- stream dependencies waiting on faulted work

The HIP binding remains responsible for translating the resulting IREE status into the appropriate HIP error.

## Architecture

Native event ingestion remains private to the AMDGPU driver. The portable HAL already represents the externally visible behavior through failed operations, semaphores, queues, and devices, so this change does not add a generic HAL API or expose HSA agents outside the driver.

This also preserves the remoting boundary: a remote implementation only needs to transport ordinary HAL failures rather than reproduce the HSA event model on the client.

Note: this reopens https://github.com/ROCm/hrx-system/pull/366, which landed without review and was reverted.


zjgarvey:
@benvanik I accidentally merged and then reverted immediately Andrew's old PR at #366 . I think this PR needs some changes before landing (there is a possible hang when logical device teardown and a device fault race). I'm going to implement a fix and request your review in a bit.

benvanik (changes requested):
This is the right integration point for asynchronous GPU faults. ROCr's system event carries the useful VM-fault detail that its per-queue callback does not, and the agent fanout, first-error-wins statuses, unmatched-event fallback, and registration after full device construction are all good foundations.

I'm requesting changes because fault delivery, queue publication, and teardown need one coherent terminal-state transition. The current patch has two races that can strand waiters, plus a process-lifetime issue with the registered callback.

1. A memory fault after device teardown starts can make destruction wait forever.

   `iree_hal_amdgpu_logical_device_destroy` marks the registry entry as tearing down before any host queue begins teardown. A matching event then sets `handled = true`, skips all queue/device failure delivery, and returns `HSA_STATUS_SUCCESS`. Later, host-queue teardown closes submission admission and waits indefinitely for either the final submitted epoch or its stop signal.

   Those two behaviors are incompatible for VM faults. In ROCr `920e5a8a70669d28b930c33fcef2b18c4b47d152`, [`AqlQueue::ExceptionHandler`](https://github.com/ROCm/rocm-systems/blob/920e5a8a70669d28b930c33fcef2b18c4b47d152/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp#L1453-L1463) deliberately does not invoke the queue error callback for a memory fault; it routes that fault to [`Runtime::VMFaultHandler`](https://github.com/ROCm/rocm-systems/blob/920e5a8a70669d28b930c33fcef2b18c4b47d152/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp#L2385-L2461). If any system-event callback returns success, ROCr considers the fault handled and suppresses its abort fallback. A fault in this window can therefore stop the GPU epoch, wake neither IREE wait, and leave logical device destruction blocked forever.

   Please coordinate this with the existing AMDGPU teardown pattern: close admission for every affected queue, keep the callback path capable of releasing the waits teardown still depends on, quiesce callback access, and only then join/destroy callback-visible resources. Because physical-device deinitialization currently combines those phases queue by queue, this may require separating queue shutdown/quiescence from final resource destruction across the device. Claiming an event while intentionally not delivering it is not a valid teardown state.

2. Queue failure can race a submission past the one final notification drain.

   `iree_hal_amdgpu_host_queue_fail` atomically stores the error and wakes the completion thread, but it does not close the submission gate. The completion thread first drains/fails the notification ring and only afterward calls `iree_hal_amdgpu_host_queue_cancel_pending`, which acquires `submission_mutex` and sets `is_shutting_down`.

   A producer may already hold `submission_mutex` when the fault arrives. The completion thread can snapshot `last_published`, finish its only failure drain, and then block trying to cancel pending work. The producer can publish a later epoch and commit its AQL packet before releasing the mutex. The completion thread then closes admission and exits without another drain, so the later notification, semaphore, frontier state, and retained resources remain live until a later device destruction; a waiter can hang in the meantime. The logical-device failure check does not protect a submission already in flight.

   Please make terminal failure a single queue transition. The completion owner can close admission under `submission_mutex` first, thereby waiting out the current publisher, then fail/drain the now-stable published state, and then settle deferred operations using their existing callback ownership protocol. Keeping the blocking transition on the completion thread also keeps it out of ROCr's callback. The HSA queue callback, the system-event callback, and the invalid wait-result path should all converge on this same transition instead of having subtly different cleanup behavior.

3. The process-wide callback currently outlives the code that implements it.

   ROCr stores each system-event function pointer until the final HSA shutdown; there is no unregister operation. Our `libhsa` contract explicitly allows another process component to retain HSA after an IREE driver releases its reference, and CMake supports `BUILD_SHARED_LIBS`. If an embedding application unloads the IREE/AMDGPU shared objects while HIP or another HSA client keeps ROCr initialized, ROCr retains a pointer into unloaded IREE code. A later event, or final HSA shutdown, can jump through that stale pointer. Handling `HSA_AMD_SYSTEM_SHUTDOWN_EVENT` cannot repair this because the callback must still be callable to receive it.

   This needs an explicit, enforceable owner for the callback's process/HSA lifetime: for example an unregister capability, a portable module pin, or a build/runtime contract that makes AMDGPU callback code non-unloadable. Device registry removal alone only protects the borrowed device pointer, not the callback function pointer. Please add coverage for whichever lifetime contract is selected.

While repairing the lifecycle, I would also make ownership visible in the API. The object named `iree_hal_amdgpu_system_event_device_t` is actually a registry entry that owns copied agent handles and borrows a logical device. Returning an opaque `iree_hal_amdgpu_system_event_registration_t` from registration, storing that nullable handle on the logical device, and using it for quiescence/removal would make partial initialization and exact pairing explicit instead of doing raw-device-pointer searches with silent no-op failure. Similarly, `iree_hal_amdgpu_host_queue_fail` sounds like failing one operation, while this API records a permanent terminal queue failure; `set_failure` or `record_failure` would communicate that state change more clearly. The existing queue callback comment should also stop listing memory faults now that the system-event path owns their delivery for this ROCr behavior.

This new process-global asynchronous lifecycle also needs deterministic tests, not just the existing build coverage. At minimum, please cover a fault racing an in-flight publication, a fault after teardown admission closes but before HSA queues are destroyed, callback traversal racing registration removal, multiple logical devices sharing an agent, and final HSA shutdown followed by reinitialization. Hooks/barriers around the real state transitions will make those tests useful; sleeps will not.

The event decoding, agent matching, fanout, and sticky-status mechanics can remain. The part that needs redesign is the ownership transition around them.

benvanik (approved):
Re-reviewed the current head. The teardown/admission race, terminal failure drain ordering, and callback module-lifetime issues from my earlier review are resolved. HSA event ingestion remains private to AMDGPU and surfaces through the normal terminal HAL failure path. This is ready to land.

PR #444, opened 2026-08-27
[HAL/AMDGPU] Add fixed masked execution queues

zjgarvey:
## Motivation

libhrx HIP execution contexts need native AMDGPU queues whose execution is confined to caller-selected compute-unit sets. The HAL queue surface has no way to bind one specific physical queue to an exact native CU mask without moving libhrx's mapping and capacity policy into the driver.

## Summary

This change adds an in-tree AMDGPU bridge for querying a physical device's reserved queue range and native mask topology, configuring one selected slot, and returning its exact `iree_hal_queue_t`. An opt-in logical-device option reserves up to 64 additional slots per physical device. The default remains zero, and these slots remain outside ordinary HAL queue enumeration and lookup.

## Design

Each successful slot is bound once to one immutable mask until device teardown; AMDGPU does not choose, share, coalesce, or recycle slots. Before creating a queue, the driver validates the native device and ordinal, complete mask width and tail, nonempty selection, architecture-defined alignment groups, and coverage of every interleaved hardware partition. It then creates the HSA queue, applies the mask before publishing submission resources, and destroys the unpublished queue on failure so the slot can be retried. Exact masking requires native GPU-consumed AQL and IREE-owned ROCr initialization with `HSA_CU_MASK` unset. Permanently reserved queue axes and synchronized epoch, fault-delivery, profiling, and teardown bookkeeping support safe sparse publication.

## Testing

`clang-format` 22.1.3 dry-run and `git diff --check` were clean; AMDGPU Bazel-to-CMake regeneration produced no changes. Seven scoped AMDGPU targets passed on ROCm SDK 7.14.0a20260622 with gfx942. The AMDGPU `//libhrx/...` build passed, along with all 12 common binding tests and all 19 serialized libhrx GPU CTS. Mutation checks showed that removing alignment validation lets a misaligned mask reach queue creation, while inverting the predicate rejects the aligned control; the restored implementation passed both.

### DIFF
--- a/runtime/src/iree/hal/drivers/amdgpu/BUILD.bazel
+++ b/runtime/src/iree/hal/drivers/amdgpu/BUILD.bazel
@@ -147,11 +147,49 @@ iree_runtime_cc_library(
     ],
 )
 
+# CMake has no target visibility equivalent. Define an in-tree interface target
+# directly because iree_cc_library installs every non-test header target.
+iree_cmake_extra_content(
+    content = """
+if(IREE_HAL_DRIVER_AMDGPU)
+add_library(iree_hal_drivers_amdgpu_api_experimental INTERFACE)
+target_sources(iree_hal_drivers_amdgpu_api_experimental
+  INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/api_experimental.h>"
+)
+target_link_libraries(iree_hal_drivers_amdgpu_api_experimental
+  INTERFACE
+    iree::base
+    iree::hal
+)
+iree_add_alias_library(
+  iree::hal::drivers::amdgpu::api_experimental
+  iree_hal_drivers_amdgpu_api_experimental
+)
+endif()
+""",
+    inline = True,
+)
+
+# Restricted experimental declarations are exposed only to in-tree libhrx
+# consumers and remain outside the installed AMDGPU API surface.
+iree_runtime_cc_library(
+    name = "api_experimental",
+    hdrs = ["api_experimental.h"],
+    tags = ["skip-bazel_to_cmake"],
+    visibility = ["//libhrx:__subpackages__"],
+    deps = [
+        "//runtime/src/iree/base",
+        "//runtime/src/iree/hal",
+    ],
+)
+
 iree_runtime_cc_library(
     name = "amdgpu",
     srcs = [
         "allocator.c",
         "allocator.h",
+        "api_experimental.h",
         "aql_atomic.c",
         "aql_atomic.h",
         "aql_block_processor.c",
@@ -601,6 +639,23 @@ iree_runtime_cc_test(
 )
 
 iree_runtime_cc_test(
+    name = "api_experimental_test",
+    srcs = ["api_experimental_test.cc"],
+    deps = [
+        ":amdgpu",
+        ":api_experimental",
+        ":headers",
+        "//runtime/src/iree/base",
+        "//runtime/src/iree/hal",
+        "//runtime/src/iree/hal/cts/util:profile_test_util",
+        "//runtime/src/iree/hal/cts/util:test_base",
+        "//runtime/src/iree/hal/testing:mock_device",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+iree_runtime_cc_test(
     name = "driver_options_test",
     srcs = ["driver_options_test.cc"],
     deps = [
@@ -686,6 +741,7 @@ iree_runtime_cc_test(
     ],
     deps = [
         ":amdgpu",
+        ":api_experimental",
         ":headers",
         "//runtime/src/iree/base",
         "//runtime/src/iree/hal",
--- a/runtime/src/iree/hal/drivers/amdgpu/BUILD.bazel
+++ b/runtime/src/iree/hal/drivers/amdgpu/BUILD.bazel
@@ -649,6 +649,7 @@ iree_runtime_cc_test(
         "//runtime/src/iree/hal",
         "//runtime/src/iree/hal/cts/util:profile_test_util",
         "//runtime/src/iree/hal/cts/util:test_base",
+        "//runtime/src/iree/hal/drivers/amdgpu/util:queue_primitives",
         "//runtime/src/iree/hal/testing:mock_device",
         "//runtime/src/iree/testing:gtest",
         "//runtime/src/iree/testing:gtest_main",
--- a/runtime/src/iree/hal/drivers/amdgpu/CMakeLists.txt
+++ b/runtime/src/iree/hal/drivers/amdgpu/CMakeLists.txt
@@ -186,6 +186,23 @@ iree_cc_library(
 endif()
 
 if(IREE_HAL_DRIVER_AMDGPU)
+add_library(iree_hal_drivers_amdgpu_api_experimental INTERFACE)
+target_sources(iree_hal_drivers_amdgpu_api_experimental
+  INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/api_experimental.h>"
+)
+target_link_libraries(iree_hal_drivers_amdgpu_api_experimental
+  INTERFACE
+    iree::base
+    iree::hal
+)
+iree_add_alias_library(
+  iree::hal::drivers::amdgpu::api_experimental
+  iree_hal_drivers_amdgpu_api_experimental
+)
+endif()
+
+if(IREE_HAL_DRIVER_AMDGPU)
 iree_cc_library(
   NAME
     amdgpu
@@ -194,6 +211,7 @@ iree_cc_library(
   SRCS
     "allocator.c"
     "allocator.h"
+    "api_experimental.h"
     "aql_atomic.c"
     "aql_atomic.h"
     "aql_block_processor.c"
@@ -738,6 +756,32 @@ endif()
 if(IREE_HAL_DRIVER_AMDGPU)
 iree_cc_test(
   NAME
+    api_experimental_test
+  SRCS
+    "api_experimental_test.cc"
+  DEPS
+    ::amdgpu
+    ::api_experimental
+    ::headers
+    iree::base
+    iree::hal
+    iree::hal::cts::util::profile_test_util
+    iree::hal::cts::util::test_base
+    iree::hal::testing::mock_device
+    iree::testing::gtest
+    iree::testing::gtest_main
+  LABELS
+    "iree-build-requirement=runtime.hal.amdgpu"
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+  RESOURCE_GROUP
+    iree-hal-drivers-amdgpu-tests
+)
+endif()
+
+if(IREE_HAL_DRIVER_AMDGPU)
+iree_cc_test(
+  NAME
     driver_options_test
   SRCS
     "driver_options_test.cc"
@@ -862,6 +906,7 @@ iree_cc_test(
     "host_queue_command_buffer_test_util.h"
   DEPS
     ::amdgpu
+    ::api_experimental
     ::headers
     iree::base
     iree::hal
--- a/runtime/src/iree/hal/drivers/amdgpu/CMakeLists.txt
+++ b/runtime/src/iree/hal/drivers/amdgpu/CMakeLists.txt
@@ -767,6 +767,7 @@ iree_cc_test(
     iree::hal
     iree::hal::cts::util::profile_test_util
     iree::hal::cts::util::test_base
+    iree::hal::drivers::amdgpu::util::queue_primitives
     iree::hal::testing::mock_device
     iree::testing::gtest
     iree::testing::gtest_main
--- a/runtime/src/iree/hal/drivers/amdgpu/api.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/api.h
@@ -256,6 +256,11 @@ typedef struct iree_hal_amdgpu_logical_device_options_t {
     // metadata such as device-side command-buffer fixup inputs without using
     // the file staging pool.
     uint32_t upload_capacity;
+    // Number of additional fixed physical queues reserved for the restricted
+    // execution-queue compatibility API, up to 64. These queues are absent
+    // from the ordinary provisioned queue inventory and are created only after
+    // a caller binds one immutable execution-unit mask.
+    uint32_t experimental_execution_queue_count;
   } host_queues;
 
   // Per-physical-device queue_read/queue_write file staging policy.
--- a/runtime/src/iree/hal/drivers/amdgpu/api_experimental.h
+++ b/runtime/src/iree/hal/drivers/amdgpu/api_experimental.h
@@ -29,7 +29,8 @@ typedef struct iree_hal_amdgpu_experimental_execution_queue_topology_t {
   iree_host_size_t private_physical_queue_count;
   // Number of native compute units addressable by an HSA queue mask.
   uint32_t native_compute_unit_count;
-  // Required alignment of enabled native CU runs in mask bits.
+  // Native mask group size and alignment in bits. Groups begin at bit zero and
+  // must each be wholly enabled or wholly disabled.
   uint32_t native_compute_unit_mask_alignment;
   // Number of hardware partitions interleaved across native mask bits. Native
   // bit N belongs to partition N modulo this count, and a confining mask must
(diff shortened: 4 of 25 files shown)


benvanik:
it'd be useful to show why you need this (if you have a scratch branch, gist, etc) and how you plan to use it. in general, api.h extensions are not allowed: they are incompatible with replay, remoting, and profiling as well as heterogeneous/multi-device (as frontiers break when you randomly mutate queues). the design is incompatible with the HAL, and it's hard to pick apart the feature requirements from the implementation when the feature requirements are absent :)

we can hack in experimental features, but they will have massive caveats. the goal is not to harden this implementation, then, but to remove as much as possible of it entirely so as to lessen the impact going forward. detailing the requirements would help us do that. for example, the original PR conflated dynamic queues with CU masking, which are independent: dynamic queues are not a HAL concept (HAL models hardware queues) and any virtualization/queue multiplexing needs to happen in libhrx. CU masking is independent and likely needs to be carried as additional data on queue operations, as HAL queues are not FIFO so setting any queue-level state requires flushing incompatible work that the HAL itself does not track, etc.

zjgarvey:
> it'd be useful to show why you need this (if you have a scratch branch, gist, etc) and how you plan to use it. in general, api.h extensions are not allowed: they are incompatible with replay, remoting, and profiling as well as heterogeneous/multi-device (as frontiers break when you randomly mutate queues). the design is incompatible with the HAL, and it's hard to pick apart the feature requirements from the implementation when the feature requirements are absent :)
> 
> we can hack in experimental features, but they will have massive caveats. the goal is not to harden this implementation, then, but to remove as much as possible of it entirely so as to lessen the impact going forward. detailing the requirements would help us do that. for example, the original PR conflated dynamic queues with CU masking, which are independent: dynamic queues are not a HAL concept (HAL models hardware queues) and any virtualization/queue multiplexing needs to happen in libhrx. CU masking is independent and likely needs to be carried as additional data on queue operations, as HAL queues are not FIFO so setting any queue-level state requires flushing incompatible work that the HAL itself does not track, etc.

I'm trying to help Andrew land some of his work to enable cooperative dispatch while he is out. The branches are the ones numbered and labelled under `pr231` in https://github.com/AWoloszyn/hrx-system/branches/active 

The branch dependencies aren't completely linear, but I have a DAG in my notes. I'll upload a few of my personal docs to try and triage these branches and share a link in a moment when I'm done doing a portability pass on the notes.

zjgarvey:
Here is a more portable version of one of my local docs: https://gist.github.com/zjgarvey/a568dd96439af1e71e3c374f2eb97009

benvanik:
thanks! I'll look through the doc - it's a useful dag, but not quite a design spec - more a trail of implementation archeology tidbits. I'll see if I can parse anything out of it. RE the H items: I don't generally care about the HRX stuff and you can land whatever you want there - but I do want to guard hacks from landing in the HAL as HRX then depending on it makes it harder to recover from. If there's urgency (and direct customer requirements) we can add big warning signs (renaming things to experimental/etc) to at least make it clearer they are not supported/ratified parts of the API and subject to breakage/incompatibility - but I'd prefer not to do that unless there's actual need instead of checkboxing.

benvanik (changes requested):
I re-reviewed this with a different landing goal: unblock the HIP execution
context work while treating this implementation as a compatibility island in
the AMDGPU driver we expect to replace. I am not asking this PR to invent the
portable execution-partition abstraction first. The native lease table, exact
mask programming, same-mask sharing, final-release drain/reset, and poisoned
slot behavior are coherent enough for that temporary role.

I would approve this after making that temporary boundary mechanically real:

1. Move every new consumer-facing declaration out of the normal AMDGPU
   `api.h` and into `iree/hal/drivers/amdgpu/api_experimental.h`. This includes
   the opaque lease handle and acquire/release/affinity functions. Give the
   types and functions an `iree_hal_amdgpu_experimental_...` prefix so the
   eventual deletion surface is searchable and they cannot be mistaken for
   supported AMDGPU driver API.

   Put the header in a separate Bazel target with
   `visibility = ["//libhrx:__subpackages__"]`. The existing AMDGPU package has
   public default visibility, so moving the declarations to another header
   without a separate restricted target would not enforce anything. The normal
   `:amdgpu` target can depend on the header target so `logical_device.c`
   compiles against its declaration; libhrx should have the only external
   dependency edge. For CMake, keep the header as a private source of the
   existing AMDGPU target rather than a public/installable `HDRS` entry.

2. Reject non-native devices before the AMDGPU cast. Today acquisition calls
   `iree_hal_amdgpu_logical_device_cast` immediately. Replay recorders and
   remote clients have different vtables, so that is an assertion failure in
   checked builds and an invalid object reinterpretation otherwise. Check the
   resource vtable first and return a loud `FAILED_PRECONDITION` (or similarly
   explicit unsupported status) saying that the experimental API requires a
   native in-process AMDGPU logical device. Add a test that passes a real
   non-AMDGPU or replay-wrapped device and verifies the status instead of a
   crash.

3. Remove the privately allocated device-extension type `2u`. Moving that
   declaration to an experimental header does not contain the collision: the
   value is still interpreted in the globally assigned HAL extension chain.
   Provision the extra slots through an AMDGPU-specific device parameter
   instead. A single parameter such as
   `experimental_execution_queue_count=N` can mean "reserve N additional
   mask-programmable queue slots per physical device"; the driver keeps its
   ordinary queue count and derives/validates total capacity internally.
   Libhrx already calls parameterized `iree_hal_driver_create_device_by_ordinal`
   and can pass this AMDGPU-specific parameter without changing generic HAL
   device creation semantics.

4. Keep one libhrx adapter as the only consumer. Ideally it is its own small
   Bazel target under `libhrx/src/binding/hip/`, and that target alone depends
   on `:api_experimental`. It translates HIP CU masks, owns the lease for the
   stream/execution scope, and returns the private affinity to the existing
   generic streaming machinery. Other libhrx code should not include the
   experimental header. Device initialization only needs to pass the named
   AMDGPU parameter and does not need the header.

5. State the actual contract at the top of the experimental header: native
   in-process AMDGPU devices only; not replayable or remotable; no fallback;
   the returned affinity is valid only while the lease is retained; callers
   stop issuing new work before final release; final release drains outstanding
   work before restoring the default mask. These are constraints of the
   temporary bridge, not generic HAL semantics.

6. Keep follow-on escape hatches in this same quarantine. If cooperative
   dispatch or dispatch-property queries are proven necessary for the HIP path,
   they belong in the same restricted target and behind the same one libhrx
   adapter. They should not add a second hole to normal `api.h`.

This leaves the useful native queue machinery intact and lets the HIP work
move. It also gives the AMDGPU replacement one header, one target, one symbol
prefix, and one libhrx adapter to remove or replace. Once these boundaries are
in place I do not think the portable child-device/execution-partition design
needs to block this PR.

## Concrete build shape

The Bazel shape should be equivalent to:

```python
iree_runtime_cc_library(
    name = "api_experimental",
    hdrs = ["api_experimental.h"],
    deps = [
        "//runtime/src/iree/base",
        "//runtime/src/iree/hal",
    ],
    visibility = ["//libhrx:__subpackages__"],
)

iree_runtime_cc_library(
    name = "amdgpu",
    # Existing sources and public api.h remain here.
    deps = [
        ":api_experimental",
        # Existing dependencies...
    ],
)
```

Same-package access allows `:amdgpu` to depend on the restricted target. The
libhrx adapter explicitly depends on both the implementation and experimental
header target. No umbrella header re-exports `api_experimental.h`.

In CMake, add `api_experimental.h` under the existing AMDGPU target's `SRCS`,
not `HDRS`. The repository CMake rule installs `HDRS`, while a header listed as
a private source remains available to the in-tree libhrx build without being
published as installed AMDGPU API. CMake has no equivalent of Bazel visibility,
so non-installation plus the one dependency/include site is the available
boundary there.

## Suggested experimental header contract

The exact prose can change, but the header should be this explicit:

```c
// Temporary native AMDGPU bridge for libhrx HIP execution contexts.
//
// This API is restricted to in-tree libhrx consumers. It is not a HAL
// extension and is not supported by wrapped, replay, or remote devices. The
// AMDGPU driver replacement is expected to remove this API or replace it with
// a portable execution-partition contract.

typedef struct iree_hal_amdgpu_experimental_execution_queue_t
    iree_hal_amdgpu_experimental_execution_queue_t;

// Acquires an immutable native execution-unit partition. Equal masks may share
// one backend queue. The returned affinity is valid until final release.
IREE_API_EXPORT iree_status_t
iree_hal_amdgpu_experimental_execution_queue_acquire(
    iree_hal_device_t* device,
    iree_hal_queue_affinity_t device_affinity,
    iree_host_size_t mask_bit_count,
    const uint32_t* mask,
    iree_hal_amdgpu_experimental_execution_queue_t** out_execution_queue);

// Drains work already issued to the lease, restores the native queue's default
// mask, and releases the lease. The caller must have stopped issuing new work.
IREE_API_EXPORT void
iree_hal_amdgpu_experimental_execution_queue_release(
    iree_hal_amdgpu_experimental_execution_queue_t* execution_queue);

// Returns the private native affinity associated with |execution_queue|.
IREE_API_EXPORT iree_hal_queue_affinity_t
iree_hal_amdgpu_experimental_execution_queue_affinity(
    const iree_hal_amdgpu_experimental_execution_queue_t* execution_queue);
```

## Suggested safe native-device gate

The exact helper can stay private to `logical_device.c`:

```c
static bool iree_hal_amdgpu_logical_device_isa(
    const iree_hal_device_t* base_device) {
  return iree_hal_resource_is(base_device,
                              &iree_hal_amdgpu_logical_device_vtable);
}

if (IREE_UNLIKELY(!iree_hal_amdgpu_logical_device_isa(base_device))) {
  return iree_make_status(
      IREE_STATUS_FAILED_PRECONDITION,
      "experimental AMDGPU execution queues require a native in-process "
      "AMDGPU logical device; wrapped, replay, and remote devices are "
      "unsupported");
}
```

Only after this check should acquisition use the existing checked cast.

## Verification requested after the reshaping

- Existing native `host_queue_submission_test` cases still pass under ASAN.
- The repeated hardware run remains stable.
- A non-AMDGPU or replay-wrapped device returns the documented unsupported
  status from acquire.
- `git grep api_experimental.h -- ':!runtime/src/iree/hal/drivers/amdgpu/**'`
  finds only the one libhrx adapter include.
- A Bazel reverse-dependency query finds only the adapter target beneath
  `//libhrx/...` consuming `:api_experimental`.
- Neither the normal AMDGPU `api.h` nor an umbrella header exposes the symbols.

benvanik (changes requested):
# PR 444 follow-up draft after `e8f6b11ca`

The experimental consumer boundary is now in the right place, but
`amdgpu_testonly` is an unacceptable runtime/HAL testing pattern and a blocking
design issue independent of the lease implementation's line-level correctness.

I understand the reasoning that led here: fail production closed when the ROCr
contract is uncertain, retain a way to exercise the implementation, and prove
that way cannot escape into shipped artifacts. The problem is that compiling a
different constructor into a separately configured driver changes the subject
under test. The production driver rejects this feature before construction;
the tests compile another driver in which that rejection is absent. Passing
those tests is not evidence that the shipped path works. The artifact scans,
negative link probes, install-manifest checks, and symbol checks only prove that
the tested implementation is unreachable by users.

That inversion cannot become precedent in `runtime/`. Tests exercise the
production implementation and its production contracts. A missing runtime or
hardware capability belongs in the test platform requirements, with a focused
availability failure on unsupported deployments. It must not be solved by
compiling a parallel implementation whose only consumer is the test suite.
The second build also doubles the configuration surface of a driver whose
lifetime and concurrency behavior are already difficult to reason about; every
future source change would have to remain correct under a production mode and a
test-only semantic mode.

Please remove `amdgpu_testonly` and all machinery that exists to construct,
link, inspect, or prove isolation of that alternate driver. This is a required
redesign, not optional cleanup. I do not want more effort spent hardening or
documenting the isolation boundary.

We also do not need ROCm 7.2 compatibility for this feature. Production support
should be defined against the modern ROCr version the product will require,
and the production path should be exercised directly on that runtime.

Please rebase onto current `main` and reduce this before another review:

1. Require the modern ROCr contract needed by execution queues. Audit or repair
   ROCr's exact-mask set/readback behavior there; do not emulate support for an
   old runtime in HAL.
2. Delete the separately compiled `amdgpu_testonly` driver, conditional
   constructor semantics, test-only logical-device constructor and harness,
   production/test artifact scanners, negative link probes, install-manifest
   checks, and the build graph needed only to prove those test-only symbols do
   not escape.
3. Exercise the actual production constructor and experimental API on the
   required ROCr. If the runtime contract is unavailable, the feature may fail
   availability cleanly, but we should not merge an otherwise unreachable
   implementation and call the HIP requirement unblocked.
4. Keep the useful containment changes: restricted `api_experimental.h`,
   experimental symbol names, exact-native-device gate, AMDGPU device
   parameter, non-installation, and one downstream libhrx adapter.

There is also a larger scope question in the new accepted-work generation
tracker. The original product requirement is masked execution contexts, not a
general reusable native-lane scheduler. A much smaller temporary contract can
avoid reprogramming a lane after final release:

- configure a specialized lane once for one exact mask;
- share it across simultaneous acquisitions of the same mask;
- remove its affinity from admission after the last release;
- retain the dormant lane for later reacquisition of the same mask;
- reclaim it only during ordinary device teardown;
- return resource exhaustion when the configured distinct-mask capacity is
  consumed.

That contract trades arbitrary distinct-mask churn for a dramatically smaller
and safer compatibility island. It does not need to reset a CU mask while the
device is live, so it does not need lease-generation ownership threaded through
pending operations, replay continuations, file work, staging work, and every
other asynchronous host-queue path merely to prove reset safety.

If arbitrary mask churn and cross-mask slot reuse are actual customer
requirements, state them explicitly with the expected context lifetime and
concurrency. Then we can judge the scheduler/lifetime machinery as its own
feature. They should not be silently presumed into A2 while this branch chain
is already under review pressure.

The intended landing shape is a deliberately limited HIP bridge that can be
deleted during the AMDGPU refactor, not a second queue-lifetime subsystem and a
parallel test build of the driver.

benvanik:
I need to correct one term from my earlier review: "native-lane scheduler" was
too loose. This patch does not add an operation-level scheduler that
load-balances or time-slices submitted work. It does, however, add a HAL-owned
queue virtualization and assignment policy, and that is also outside the
boundary we need here.

The proposed acquire API accepts a physical-device affinity plus a CU mask, not
an exact physical queue. The implementation then:

- searches the hidden specialized queue slots and selects the first available
  slot;
- coalesces equal-mask acquisitions onto one physical HSA queue;
- dynamically publishes that queue's private affinity for the lease lifetime;
- closes and drains or cancels accepted work on final release; and
- resets the CU mask and returns the physical slot so a later acquisition can
  repurpose it for an unrelated mask.

That is queue placement, sharing, and recycling policy. libhrx may own the
virtual streams, but this change makes the AMDGPU HAL decide how those logical
execution scopes map onto physical queues. The generation-tagged accepted-work
tracker threaded through every asynchronous work form is machinery required by
that reusable lease policy; it is not required merely to expose native CU-mask
programming.

The hard boundary is that HAL represents hardware queues. It does not multiplex
logical streams onto them and does not choose, share, or recycle hardware
queues on behalf of a higher-level runtime. The current AMDGPU queue model is
being replaced, so this compatibility change must be the minimum removable
native hook. It must not add interim queue algorithms or a second lifetime
model that the replacement then has to preserve or unwind.

My earlier suggestion to retain a dormant specialized queue keyed by mask was
therefore still too much HAL policy; please treat that part of the review as
withdrawn. The reduction I am asking for is:

- keep only the AMDGPU-native mechanics needed to provision a fixed physical
  queue and program an immutable mask on an exact, caller-selected queue
  identity;
- remove the mask-keyed registry, equal-mask coalescing, first-free selection,
  cross-mask slot recycling, and the generation/accepted-work lease subsystem;
- keep fixed physical queues as fixed physical queues with stable identities;
  if a temporary containment rule is needed, bind the mask for the physical
  queue's lifetime rather than inventing a reusable lease lifecycle; and
- make libhrx own virtual execution contexts and all mapping, sharing,
  exhaustion, and reuse policy. It may explicitly choose among provisioned
  physical queues, but HAL must not choose for it.

This is separate from the `amdgpu_testonly` blocker. Both need to be removed in
the reduction/rebase. Please also avoid using this PR to anticipate or ratify
the replacement queue API; the point of the experimental compatibility island
is that it can be deleted, not that it becomes the seed of another HAL queue
abstraction.

benvanik:
Zach, I found and reviewed
`users/zjgarvey/handoff/pr444-native-2026-08-31` at
`44fb26f2850f720ecea67ba89bcd491dc05bfb78`. This is the right reduction and
resolves the architectural objection in the current review. Thank you for
carrying the redesign through to a substantially smaller ownership boundary.

The important shape is now explicit:

- private slots are fixed physical queue ordinals reserved at device creation;
- the caller selects the exact physical queue, while HAL does not search,
  coalesce equal masks, lease, share, or recycle slots;
- the native mask is applied before the queue is published and remains
  immutable for that slot;
- ordinary `ANY` routing excludes the private range;
- libhrx owns mapping, sharing, exhaustion, and reuse policy; and
- the restricted exact-native API fails closed for wrappers/remoting/replay and
  defaults to no additional queues.

That is the removable compatibility island we were asking for. I would review
this as the replacement implementation rather than asking for more repair work
on the lease-based head currently attached to #444.

I see one concrete correctness item to fix before final approval:
`native_compute_unit_mask_alignment` reports a required two-CU grouping on
gfx10/gfx11/gfx12, but `execution_queue_configure` does not currently validate
that the supplied mask honors that grouping. Exact confinement cannot depend on
ROCr reporting a reduced/rounded mask, since this branch correctly treats that
status as unreliable. Please reject misaligned groups before creating the
queue.

Please either update #444 to this implementation or open a replacement PR and
link it here so it gets the normal checks and line review. The design direction
itself is no longer blocking.

zjgarvey:
Thanks, Ben. 

I'm implementing a fix for the correctness issue you mentioned and will push the handoff branch here. 

@AWoloszyn let me know if you are doing anything with the amdgpu-02 work right now which might collide with this.

AWoloszyn:
No concerns on my end,

PR #450, opened 2026-08-28
[HIP] Pack native kernel arguments from executable metadata

zjgarvey:
## Motivation

HIP pointer-array launches provide pointers to caller-owned argument values,
while AMDGPU dispatch consumes one contiguous native kernarg image whose layout
comes from executable metadata. Reconstructing that layout independently in
immediate and graph paths risks ABI drift and lifetime bugs.

## Summary

- Extract and validate immutable per-symbol argument layouts when a module is
  loaded.
- Pack pointer-array launches into native ABI layout while preserving
  pre-packed argument buffers and their declared lengths.
- Use the same packing and validation helpers for immediate and graph launches,
  with graph templates and executable instances owning captured argument bytes.

## Design

Module loading enumerates executable exports and compiles their parameter
records into compact copy and pointer-resolution operations. Validation rejects
duplicate or out-of-range binding ordinals, binding widths that do not match a
device pointer, overlapping or out-of-order native argument ranges, and sizes,
counts, or ordinals that exceed their stored widths before the module is
published. The resulting symbol table and operations occupy module-owned
storage that remains stable for the module lifetime.

Pointer-array launches merge the copy and pointer-resolution operations by
source ordinal into a zero-initialized native kernarg image. Device-pointer
values are loaded with byte copies, so caller-owned argument storage does not
need host pointer alignment or a typed pointer lvalue. Immediate launches use
bounded inline argument storage for small layouts; larger layouts use the
stream's host allocator and leave through one cleanup path.

Pre-packed launches are validated against the reflected direct-argument extent
without rewriting their bytes or replacing their declared length. Graph
capture copies either the pre-packed span or the newly packed pointer-array
image into graph-owned storage, and graph instantiation copies dispatch bytes
into executable-owned storage. Immediate dispatch and graph replay therefore
do not depend on the lifetime of caller-owned argument containers.

## Testing

- Required changed-target presubmit and clang-tidy: 15/15 targets passed.
- `//libhrx/src/binding/common/...`: 12/12 test targets passed.
- Focused graph, argument-packing, and module tests: 3/3 passed under ASan,
  Clang UBSan, and Clang TSan.
- `//libhrx/cts:hip_module_execution_test` with the production `amdhip64`
  library and a gfx942 code object: 2/2 tests passed in the ordinary
  command-buffer path and 2/2 passed with
  `HRX_DIRECT_QUEUE_DISPATCH=1`; no tests skipped.


zjgarvey:
I'm working out some bad behavior from my new AI setup (e.g., terminology like H6 obviously doesn't belong in a public-facing PR).

AWoloszyn (approved):

PR #451, opened 2026-08-28
[HAL/AMDGPU] Add virtual-memory access scopes

zjgarvey:
## Motivation

HAL queue family affinity selects a device topology, but it cannot express whether virtual-memory protection applies to the selected device agents, their associated host agents, or both. AMDGPU needs that distinction, while backends that cannot enforce a requested scope need an explicit portable failure mode.

The replay allocator advertises the virtual-memory capability of its wrapped allocator, so successful virtual-memory operations also need complete capture and replay support.

## Summary

- Add `DEVICE`, `HOST`, and `ALL` virtual-memory access scopes and reject empty or unknown scopes before backend dispatch.
- Resolve AMDGPU scopes to the selected GPU agents, their associated CPU agents, or the deduplicated union. Keep Vulkan device-only and return `UNIMPLEMENTED` for valid host-access requests it cannot enforce.
- Advance replay format 7.0 to 7.1 and record, dump, and replay reservation, release, physical allocation/free, map, unmap, protect, and advise operations with stable object identities.

## Design

Queue family affinity continues to select topology; access scope selects the execution domains within it. The recorder validates reservation and physical-memory provenance before forwarding operations and consumes their identities only after successful release or free.

The executor validates payloads, object types, allocator ownership, and ranges, then tracks live mappings and tears resources down leaf-first. Queue completion signals and transfer staging remain alive until accepted work completes. Preflight rejects unresolved forward semaphore dependencies and opaque device-memory waits before a protect, unmap, or end-of-file completion drain, preventing replay from waiting forever before a required producer is submitted.

## Testing

- Replay: 6/6 targets and 101/101 cases passed normally; 101/101 cases passed under default-layering ASAN, with no ASAN/LSan diagnostics.
- AMDGPU access policy: 12/12 cases passed. Vulkan VMM/BDA: 30/30 cases passed.
- The full libhrx build passed 229/229 targets; host VMM CTS passed 2/2 cases.
- Clang-format passed for 39/39 changed C/C++ files; BUILD-to-CMake passed 1/1; cumulative diff hygiene and the pinned TheRock SDK `7.14.0a20260622` self-test (27/27) passed. No system ROCm userspace was used.

### DIFF
--- a/libhrx/cts/tests/virtual_memory/virtual_memory_test.cpp
+++ b/libhrx/cts/tests/virtual_memory/virtual_memory_test.cpp
@@ -5,6 +5,46 @@
 
 #include "hrx_test_fixture.hpp"
 
+namespace {
+
+static void IgnoreCleanupStatus(hrx_status_t status) {
+  if (!hrx_status_is_ok(status)) hrx().status_ignore(status);
+}
+
+class VirtualMemoryResources {
+ public:
+  explicit VirtualMemoryResources(hrx_allocator_t allocator)
+      : allocator(allocator) {}
+
+  ~VirtualMemoryResources() {
+    if (is_mapped) {
+      IgnoreCleanupStatus(hrx().allocator_virtual_memory_unmap(
+          allocator, virtual_buffer, /*virtual_offset=*/0, size));
+    }
+    if (physical_memory) {
+      IgnoreCleanupStatus(
+          hrx().allocator_physical_memory_free(allocator, physical_memory));
+    }
+    if (virtual_buffer) {
+      IgnoreCleanupStatus(
+          hrx().allocator_virtual_memory_release(allocator, virtual_buffer));
+    }
+  }
+
+  // Unowned allocator that outlives all resources in this helper.
+  hrx_allocator_t allocator = nullptr;
+  // Owned virtual address reservation released by the destructor.
+  hrx_buffer_t virtual_buffer = nullptr;
+  // Owned physical allocation freed by the destructor.
+  hrx_physical_memory_t physical_memory = nullptr;
+  // Size in bytes of the reservation, allocation, and mapping.
+  size_t size = 0;
+  // Whether the physical allocation is currently mapped.
+  bool is_mapped = false;
+};
+
+}  // namespace
+
 TEST_CASE_METHOD(HrxTestFixture, "query virtual memory support",
                  "[virtual_memory][query]") {
   hrx_allocator_t alloc = hrx().device_allocator(device_);
@@ -14,8 +54,7 @@ TEST_CASE_METHOD(HrxTestFixture, "query virtual memory support",
   REQUIRE_OK(hrx().allocator_query_virtual_memory(
       alloc, HRX_MEMORY_TYPE_DEVICE_LOCAL, &supported, &min_page, &rec_page));
 
-  // The task driver does not support VM — just verify the query doesn't crash
-  // and returns consistent values.
+  // Supported and unsupported allocators report consistent granularity values.
   if (supported) {
     REQUIRE(min_page > 0);
     REQUIRE(rec_page >= min_page);
@@ -25,8 +64,8 @@ TEST_CASE_METHOD(HrxTestFixture, "query virtual memory support",
   }
 }
 
-TEST_CASE_METHOD(HrxTestFixture, "reserve fails gracefully without VM support",
-                 "[virtual_memory][reserve]") {
+TEST_CASE_METHOD(HrxTestFixture, "virtual memory lifecycle",
+                 "[virtual_memory][lifecycle]") {
   hrx_allocator_t alloc = hrx().device_allocator(device_);
 
   bool supported = false;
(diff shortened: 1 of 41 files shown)


benvanik (changes requested):
The access-scope split is the right generic contract. It keeps topology
selection in queue affinity, represents host/device scope as a compact portable
value, fails unsupported Vulkan scopes loudly, and preserves the AMDGPU
same-logical-device ownership checks.

The replay path is not implemented, however. At this head:

- the replay allocator advertises the base allocator's VMM capability;
- successful VMM operations are recorded with
  `IREE_HAL_REPLAY_PAYLOAD_TYPE_NONE`;
- `virtual_memory_protect` forwards `access_scope` to the live base allocator
  but does not write it into the capture; and
- the replay executor has no implementation for any allocator VMM operation
  code, so the resulting successful records later fail with
  `IREE_STATUS_UNIMPLEMENTED`.

That is live wrapper forwarding, not replay support. A program can complete a
VMM lifecycle while recording and leave behind an artifact that contains
neither the resource identities nor the parameters needed to reproduce it.
This is especially dangerous because capture succeeds; the failure is deferred
until someone relies on the artifact.

Complete VMM replay is required for this API change. Please do not contain the
problem by making recorder-wrapped VMM return unavailable or unimplemented.
The landing needs to include:

1. Stable captured identities and replay ownership for virtual reservations and
   physical-memory handles, including release/free.
2. Serialized payloads for the state-changing VMM lifecycle: reserve,
   physical allocation, map, unmap, protect, and release/free. The protect
   payload must preserve the reservation identity, range, queue affinity,
   `access_scope`, and protection exactly. Advise and query operations must be
   deliberately replayed or classified as semantic no-ops rather than falling
   through as unsupported records.
3. Executor reconstruction and application of those operations without raw
   in-process pointers or backend-private identities in the capture format.
4. An end-to-end test that records and then replays a complete successful VMM
   lifecycle and proves the access scope survived the artifact boundary. A
   recorder-only test or live forwarding test is not sufficient.

This is a blocking part of the portable HAL contract, not follow-up cleanup.
If the VMM work is worth exposing through generic HAL, its capture and replay
semantics need to land with it.

zjgarvey:
The latest update completes the replay contract requested in the review:

- reservations and physical-memory handles now use stable captured identities,
  retain their owning allocators during replay, validate allocator ownership,
  and have explicit release and free operations;
- reserve, allocate, map, unmap, protect, advise, release, and free parameters
  are serialized and reconstructed without process-local pointers;
- granularity queries are explicitly treated as semantic no-ops because later
  records contain their resolved parameters; and
- an end-to-end test uses the production recorder wrapper to record and replay
  the complete lifecycle, including the access scope. A truncated-replay test
  also proves that teardown unmaps live ranges before freeing or releasing
  their backing objects.

The replay suite and changed-target clang-tidy pass on the updated branch
rebased onto current `main`.

benvanik (approved):
Thank you for carrying the replay contract through as part of this change. The updated head closes the artifact-boundary gap from the previous review:

- virtual reservations and physical allocations have stable captured identities, retain their owning allocator during replay, and have explicit release/free operations;
- map, unmap, protect, and advise preserve the resource identities and operation parameters, including queue affinity and the new access scope, while the granularity query is deliberately classified as a semantic no-op;
- the executor validates object types and allocator ownership, tracks live mapped ranges, rejects premature release/free, and unwinds mappings before resources when replay ends early; and
- the production recorder wrapper is exercised end to end through a different live allocator, including the complete lifecycle and truncated-capture cleanup.

The portable scope boundary remains sound: queue affinity selects topology, the scope selects host/device execution domains, AMDGPU resolves that to topology agents, and Vulkan rejects scopes it cannot enforce. This now preserves that contract across replay as well as live execution.

Approved at `9b0809b2ee27b02d41309d8e4c0ad9938167553c`.

zjgarvey:
Doing a rebase to check for semantic conflicts and will squash and merge if clean.

zjgarvey:
Rebase review update (`main` `109d935b`, PR head `9b0809b2`): the change applies cleanly, but the combined tree is **not merge-ready** because of four high-severity replay/VMM issues:

- A capture-write failure can override a successful VMM operation, leave a consumed proxy looking retryable, and block the remaining teardown.
- VMM handles are not bound to their originating recorder allocator, so raw or foreign handles can be dereferenced incorrectly or captured with unrelated object IDs.
- Signal-less asynchronous queue work is flushed but not completed before later VMM mutation or end-of-file teardown, allowing in-flight access to unmapped storage.
- If teardown unmap fails, replay discards the mapping ledger and still attempts dependent physical-memory free and reservation release.

I am applying some fixes before either landing or re-requesting a review.

zjgarvey:
Status for base `ecaaf737` / head `c507bb15`: 35/36 exact-head checks succeeded, and GitHub reports `MERGEABLE`, `APPROVED`, and `UNSTABLE`. The merge hold remains on the cancelled IREE Bazel gfx120X job.

The four commits remain patch-equivalent to the preceding range (`=` 4/4 with identical stable patch IDs; aggregate binary/full-index SHA-256 `193695b11ed7a21639ee8678b92396a145c2e8917e4468d6adc31184c28d3380`). The latest `main` delta is one Loom-only commit with no path or semantic overlap. Technical review is **HIGH 0 / MEDIUM 3**: conservative rejection of statically opaque signal-less work; loss of the sole durable cleanup-retry owner on a fallible executor path; and possible physical-memory ownership loss if recorder token allocation and compensating free both fail. The completion-drain ordering and opaque-dispatch dependency-preflight hangs are fixed; Windows enum narrowing is explicit.

Exact-head local checks passed replay (101/101 normally and under default-layering ASAN, with no ASAN/LSan diagnostics), TheRock SDK self-test 27/27, formatting 39/39, BUILD-to-CMake 1/1, AMDGPU policy 12/12, Vulkan 30/30, libhrx 229/229, and host VMM CTS 2/2. No system ROCm userspace was used.

[Attempt 3's gfx120X job](https://github.com/ROCm/hrx-system/actions/runs/34314865172/job/102358979084) built 5,021 targets successfully, then seven live-GPU suites timed out on runner `CS-RORDMZ-DT149` before the 15-minute cancellation. The [same runner reproduced the pattern on pre-PR main history](https://github.com/ROCm/hrx-system/actions/runs/34071047042/job/101588330055), while [exact current `main` passed 103/104 tests with one skip on `DT90`](https://github.com/ROCm/hrx-system/actions/runs/34313292157/job/102344196010). This classifies the cancellation as runner/GPU infrastructure, not PR source, but the PR is not fully passing.

[Issue #527](https://github.com/ROCm/hrx-system/issues/527) remains a nonblocking main-only captured-signal FIFO limitation.

zjgarvey:
I feel like the fixes are minor enough to land with the original approval, so I will merge this as long as main hasn't drifted too much since the rebase last night.

PR #455, opened 2026-08-28
[Loom] Compose recursive command product frontiers

benvanik:
Loom can now finish a portable command program without embedding, compiling, or even reading the implementation bodies of the kernels it launches. The parent product contains serialized command programs plus explicit executable-entry requirements. An optional callback publishes each surviving source-backed kernel specialization as an independently owned ordinary Loom module, ready for a worker pool, content-addressed cache, remote compiler, or target-specific pipeline.

This turns command compilation into a recursive product frontier instead of a monolithic compile-and-package operation. Requester exports determine what the parent publishes, library definitions satisfy dependencies without being implicitly re-exported, and structural kernel facets let command lowering consume the contract and pure launch configuration while stopping before the device implementation.

## What this enables

An embedding no longer has to choose between linking every kernel body into one compiler job and treating every kernel as an opaque external artifact. One indexed source universe supports local kernels, source or bytecode libraries, externally supplied executable entries, and several fact-selected specializations of one kernel definition.

```text
frozen link index shared by workers
  -> command-product job (one workspace)
       -> portable command program bytes
       -> executable-entry requirements
       -> optional owned kernel request modules
            -> cache lookup or independent target compile
```

The default path is body-blind. Without a request callback, command construction returns the portable programs and logical entry requirements while leaving implementation facets unopened. Supplying a callback opens only the implementations required by surviving launch sites, groups those sites by semantic specialization class, and transfers one child module per class.

| Authored/provider shape | Parent result | Child result |
| --- | --- | --- |
| Local `kernel.def` | Command program references its configured entry | Ordinary rooted Loom kernel module |
| `kernel.decl` satisfied by a text or bytecode library | Same as a local definition | Ordinary rooted Loom kernel module |
| One definition selected differently by launch-site facts | One entry requirement per live semantic class | One specialized module per class, not per launch site |
| Bodyless `kernel.entry.decl` | Explicit executable binding requirement | No fake or failed source request |

This is the composition boundary needed by a low-latency host: command bytes can become available without waiting for device compilation, while independent kernel jobs can run concurrently and reuse their normal module serialization, target compilation, and cache keys.

## Use the frontier from the CLI

The checked heterogeneous example links a requester and command library from text, a kernel provider from bytecode, a local kernel, two fact-selected schedule classes, and an externally bound entry:

```loom
command.program.def public @heterogeneous() launch(%storage: buffer) {
  %small = index.constant 64 : index
  %large = index.constant 256 : index
  %unit = index.constant 1 : index
  kernel.launch @local(%storage) : (buffer)
  command.program.launch @linked_stage[%small](%storage) : [index](buffer)
  command.program.launch @linked_stage[%large](%storage) : [index](buffer)
  kernel.dispatch @external[%unit](%storage) : [index](buffer)
  command.return
}
```

Seal the requester against its libraries, allowing the deliberately external entry to remain unresolved:

```shell
loom-link requester.loom \
  --library=stage.loom \
  --library=providers.loombc \
  --mode=link \
  --root=@heterogeneous \
  --allow-unresolved \
  --to=bc \
  --output=heterogeneous.loombc
```

The body-blind form emits only the parent command product:

```shell
loom-compile heterogeneous.loombc \
  --backend=command \
  --root=@heterogeneous \
  --output=commands.json \
  --emit-command-artifacts=commands/
```

Add `--emit-kernel-requests` when the caller also owns source compilation:

```shell
loom-compile heterogeneous.loombc \
  --backend=command \
  --root=@heterogeneous \
  --output=commands.json \
  --emit-command-artifacts=commands/ \
  --emit-kernel-requests=kernel-requests/
```

This example produces one command artifact, three ordinary `.loombc` child modules (`@local`, the small `@classified` class, and the large `@classified` class), and one unresolved `@external` executable requirement. The output directories and JSON manifest are CLI policy; the target library and LoomC API deal only in owned artifacts, modules, and requirement records.

## Use the frontier from LoomC

Applications build and freeze a link index once, share it across compiler workers, and resolve human-readable root names to index ordinals at ingress. Each active job supplies its own workspace and can choose whether it wants only the body-blind parent or both the parent and child source requests.

The core source-publication call has this shape:

```c
static loomc_status_t publish_kernel_request(
    void* user_data, loomc_cmd_kernel_request_t request) {
  request_queue_t* queue = (request_queue_t*)user_data;

  // Ownership of request.module transferred at callback entry. The queue can
  // compile it locally, serialize it for a remote worker, or probe a cache.
  request_queue_push(queue, request.entry_requirement_ordinal,
                     request.root_symbol, request.member_count,
                     request.module);
  return loomc_ok_status();
}

loomc_cmd_program_product_options_t options = {
    .type = LOOMC_STRUCTURE_TYPE_CMD_PROGRAM_PRODUCT_OPTIONS,
    .structure_size = sizeof(options),
    .link_index = frozen_link_index,
    .root_symbol_ordinals = &root_symbol_ordinal,
    .root_symbol_count = 1,
    .kernel_request_sink = {
        .publish = publish_kernel_request,
        .user_data = &request_queue,
    },
};

loomc_cmd_program_product_t* product = NULL;
loomc_result_t* result = NULL;
loomc_status_t status = loomc_cmd_program_product_build(
    workspace, &options, loomc_allocator_system(), &product, &result);
```

On success, `loomc_cmd_program_product_program_at` returns each serialized command artifact and its root-local requirement ordinals. `loomc_cmd_program_product_entry_requirement_at` maps those ordinals to logical entry symbols and reports whether a source request was published. Requirements without source are the executable entries the host must bind from HIP, SPIR-V, precompiled HSACO, or another external provider.

Omit `kernel_request_sink.publish` for the body-blind form. The same command artifacts and requirement table are produced, but no kernel body is opened and no child module is allocated or transferred.

## Ownership and concurrency

The API makes the asynchronous boundary explicit:

| Object | Ownership and sharing |
| --- | --- |
| Frozen link index | Retains immutable source or materialized-module providers and may be shared across workers |
| Workspace | Mutable invocation scratch; each concurrent build uses a distinct workspace |
| Command product | Immutable owner of serialized programs, copied names, root projections, and product-wide requirements; retains no source module, index, plan, or analysis state |
| Kernel request module | Transfers independently at callback entry and follows ordinary `loomc_module_t` ownership |

Request publication is provisional until `loomc_cmd_program_product_build` returns an OK status with a succeeded result. A callback may enqueue accepted children immediately, but the embedding commits their use only when the parent transaction succeeds. If the callback fails, the parent product is cancelled while ownership of already transferred modules remains with their recipients.

## Select only the required frontier

Requester ownership and library visibility now apply consistently to whole symbols and kernel facets. Public definitions supplied only as libraries satisfy dependencies but do not leak into the output. A `kernel.launch` selects the contract and configuration facets required to map a logical workload to physical geometry; a `kernel.dispatch` needs only the configured entry contract. Device implementation facets remain behind the child frontier.

The command backend consumes prepared schedule rows and the existing source fact table as its complete launch-site inventory. It groups sites without another IR walk or symbol-name lookup, classifies each selected entry once across all command roots, and publishes one source product per live class. Dead command paths create no requests, repeated equivalent launches share a request, and direct external entries remain explicit requirements.

A frozen link index may now retain immutable source providers or already materialized modules. The latter avoids serializing and reparsing products composed in memory. The provider terminology reflects this wider role:

- `loomc_link_index_source_slot_t` becomes `loomc_link_index_provider_slot_t`.
- `loomc_link_index_source_options_t` becomes `loomc_link_index_provider_options_t`.
- `loomc_link_index_builder_reserve_source_slot` becomes `loomc_link_index_builder_reserve_provider_slot`.
- Existing source fill/add operations accept the provider types; new module fill/add operations accept ordinary materialized modules.

## Scaling and body blindness

The implementation removes two backend module walks by carrying source-retention policy in compact index metadata and selecting command roots directly from indexed symbols. Exact-root embeddings stay on ordinals throughout the hot path; scanning exported INPUT roots is an explicit opt-in. Configuration projection, target specialization, provider diagnostics, and Low representation setup use the same selective-materialization policy as ordinary linking.

Optimized locked measurements show that cost follows the selected frontier rather than the source catalog or hidden implementation size:

| Scenario | Small case | Large case | Result |
| --- | ---: | ---: | --- |
| Select one INPUT root with unrelated library exports | 706 ns at 1 export | 763 ns at 4,096 exports | Warm selection is effectively independent of unrelated exports |
| Plan a 110-byte kernel configuration | 910 ns with a 42-byte body | 912 ns with a 1 MiB body | Planning does not read implementation payloads |
| Materialize that configuration | 9.285 us with a 42-byte body | 9.326 us with a 1 MiB body | Projection cost is body-size insensitive |
| Prepare distinct body-blind kernel products | 19.6 us for 1 kernel | 7.829 ms for 1,024 kernels | Linear at about 7.6 us per additional kernel |
| Publish distinct source requests | 26.7 us for 1 kernel | 16.096 ms for 1,024 kernels | Linear at about 15.7 us per additional request |

At the 1,024-kernel endpoint, the body-blind path records 1,024 external requirements and opens zero implementation bodies. The enabled path publishes 1,024 independently owned modules. A poisoned-bytecode witness proves the distinction directly: command-only construction succeeds after implementation bytes are corrupted, while source publication observes and rejects the invalid body.

The same targetless child request round-trips through ordinary Loom bytecode and compiles independently for gfx1151/wave32 and gfx942/wave64. Target specialization therefore stays on the child side of the frontier and does not couple a portable command product to one device target.

## Review surface

The highest-value boundaries to inspect are the callback ownership transaction in `loomc_cmd_program_product_build`, the distinction between source-backed and external entry requirements, the reuse of selective materialization across ordinary links and command products, and the proof that body-blind construction never opens implementation payloads. The heterogeneous checked workflow under `loom/docs/examples/product-frontiers/` exercises the public CLI and bytecode boundaries end to end.


benvanik:
I am working through what it takes to make Loom own the complete XDNA/AIE compiler path. That target makes the product-frontier work in this PR extremely useful, but it also exposes two assumptions in the current public surface that I think we should remove before it lands: requests transfer live modules, and the reusable request abstraction is named and shaped specifically as a command program requesting a kernel.

XDNA needs to take one specialized source product, separate array-level work from independently compilable tile programs, compile or resolve those units from cache concurrently, and gather their artifact contributions into a final deployment. This is not a special XDNA workflow. The coming VM work has the same shape at another level: a host VM program may request command programs for reusable command buffers and may also launch kernels directly; those command programs may request some of the same kernels. Compiling any one requested source may expose another frontier beneath it.

That is a heterogeneous nested DAG rather than a command-to-kernel tree:

```text
VM product
  requirement C0 -> command-program source A
  requirement K0 -> kernel source K

command-program product A
  requirement K1 -> the same kernel source K
```

`K0` and `K1` are distinct edges in distinct parent products. `K` is one reusable compilation node and should have one content identity/cache entry when its specialization and target contract are the same. XDNA adds array-program and tile-program nodes to the same model; it should not add an XDNA-specific LoomC ownership API.

The conceptual documentation in this PR already says the hierarchy should not require a different request representation when a host-to-command producer is added. I agree with that. The current `loomc_cmd_kernel_request_t` and `loomc_cmd_kernel_request_sink_t` make the first exercised edge shape the foundational public abstraction, though. I think #455 should factor the source request/publication contract into a target-independent LoomC surface and make command-product construction its first user. The command byte payload and command-specific convenience views can remain command-specific; the canonical requirement edges and independently scheduled child-work representation should not be.

The generic surface needs to cover both sides of an edge: the parent's durable requirements and the optional source work published to satisfy them. Merely renaming the request sink while every product invents a different requirement table would still force a recursive orchestrator to understand command, VM, XDNA, and every future target individually. A product-specific payload should be able to refer to generic parent-local requirement ordinals. Each generic requirement needs a stable symbol/contract identity and an association with zero or one selected published source unit; one source unit may satisfy several requirements. Detailed ABI and target semantics can remain ordinary Loom declarations/IR rather than being flattened into a universal C descriptor. Command-specific queries may project convenient entry information from that canonical frontier.

This does not call for a graph-specific file format or an in-memory hierarchy of modules. The durable units remain ordinary `.loombc` recipes/sources and ordinary artifact definitions; a native frontier object, if there is one, is a compact ownership/query surface over immutable sources and copied edge descriptors.

I am not attached to names or an exact struct decomposition. The requirements for the generic frontier are:

- It is recursive. Any source unit received from one frontier can be compiled by the appropriate product builder and publish another set of source-backed or external requirements through the same mechanism.

- It is heterogeneous. One parent may request different product classes, and no generic type is semantically restricted to kernels, command programs, VM functions, XDNA tile programs, or any other fixed pair of producer/consumer kinds. A worker must be able to determine the appropriate compilation route from durable request/source metadata without depending on the producer's live IR. I would avoid a closed enum that has to grow for every target; this can be represented by ordinary Loom symbol/contract identity or another extensible identity.

- It models a DAG rather than ownership of a child by one parent. A parent-local requirement ordinal identifies an edge and is useful for fast gather/binding. It is not the identity of the requested compilation. The same serialized source node may satisfy several requirement edges, including edges from products built by different compiler layers.

- Requirements and source work are distinct. A requirement may intentionally have no source because its implementation is externally owned. A source unit may contain several mutually dependent functions, several roots, an SCC, or a target-chosen cluster, and may satisfy one or more compatible requirements. `one callback -> one authored kernel -> one parent edge` must not become a structural invariant of the generic API.

- Discovery and scheduling order do not affect identity. Parent-local ordinals provide deterministic association inside the parent product; canonical source contents plus all relevant compiler, target, ABI, and code-generation configuration provide the compilation/cache identity. A parent path, callback order, or discovery ordinal must not prevent sharing. An ordinal-derived filename or diagnostic label may remain presentation metadata, but it cannot make otherwise identical child work distinct.

- Publication remains transactional in the sense already established by this PR. Accepted work is owned by the receiver even if publication later fails, while parent edges and the completed parent product remain provisional until the producing operation returns OK with a succeeded result. The exact policy for retaining an independently valid compiled cache entry after its parent fails can remain an embedding concern; no invalid parent association should be committed.

There is also an important representation/lifetime requirement. The object crossing this frontier should be an immutable serialized Loom bytecode source, not a live `loomc_module_t*`.

Today every request module retains the workspace owning its arena storage. A host that streams requests into a thread pool, remote compiler, or cache pipeline must queue those modules, so the producer workspace and its blocks remain live until the slowest consumer finishes. The returned command product correctly retains no workspace state, but the work queue becomes the displaced lifetime root. A mutable module is also neither a transport representation nor a durable content identity. The current `root_symbol` view demonstrates the secondary lifetime problem because it borrows from the transferred module.

The strong handoff invariants are:

- A published source-backed request owns a self-contained immutable bytecode source. `loomc_source_t*` with `LOOMC_SOURCE_FORMAT_BYTECODE` appears to have the right existing lifetime and thread-sharing contract, but an equivalent owned source representation is fine.

- The bytecode is ordinary standalone Loom IR accepted by existing source, indexing, linking, query, and compilation paths. It is not an interface-only stub, a target-specific request language, or a lossy fact sidecar.

- Everything the producer established about the unit travels in that IR: the selected definitions, specialization assumptions, configuration, target facts, dependencies, and provenance needed to recover identical semantics in a fresh workspace. No consumer needs the producer's analysis state or a hidden companion object.

- After publication, no child module, producer workspace, link plan, or analysis lifetime escapes through the request. The producer may serialize a completed unit, publish it, release the mutable module and recover its arena blocks, then continue discovering later units while consumers queue or compile the source independently.

- Small routing and edge-association metadata has the same independent lifetime. No string view, ordinal array, or contract record may borrow from a released module. It may be copied into the request, owned by an immutable handle, or cheaply recovered from the source index; the spelling is flexible.

The generic boundary also cannot freeze an optimization boundary. Serialization of complete Loom IR is lossless; premature decomposition is not. The compiler producing a frontier retains authority over when and how it partitions work. If functions require fusion, inlining, joint placement, joint resource allocation, or joint scheduling, they remain in one unit or those transformations happen before publication.

I am treating the information around a frontier as three different contracts. The logical interface is usually source-known: symbols, argument/result types, logical buffers and streams, shapes, and layout predicates. The required physical contract becomes stable only after enough target preparation has fixed the externally visible ABI: address spaces, alignments, resource and patch slots, legal relocation forms, and allocation envelopes promised to a parent planner. The provided realization appears after complete child lowering: encoded payloads, exact segment sizes, stack/spill use, symbols, relocations, and feature requirements.

A target may publish at an early frontier when the physical contract is already implied by source, or wait for a target-Low frontier when it is not. Every fact that can change instruction selection, scheduling, allocation, layout, addressing forms, or code-size decisions must be resolved before publication, serialized as explicit target-owned IR visible to child lowering, or remain subject to a genuine later optimization/relaxation. Address-only decisions can remain symbolic fixups. The generic request mechanism should carry arbitrary ordinary Loom IR and must not whitelist only the source-level facts needed by the current command example.

For XDNA, potentially code-generating facts include tile kind and coordinate, local-memory bank/address class and alignment, buffering depth, stream endpoint and mode, routing constraints, synchronization topology, and code/data/stack budgets. Two tile sources can share a cache entry when their differing array facts are codegen-irrelevant or representable as fixups; a tightly specialized program can deliberately include exact placement facts and trade reuse for quality. The frontier representation must leave that decision with the target compiler instead of choosing it in LoomC.

I am flexible about whether this is a generic requirement/request struct plus sink, an immutable frontier handle, or a small product-frontier interface used by product-specific builders. It could be supplied directly in each builder's options or through a common invocation extension. The command-facing API may retain convenience wrappers and projected command-entry queries, but an orchestrator must be able to enumerate durable edges and published source work without a command-specific type switch. It is also fine for #455 to exercise only command-to-kernel publication as long as the public ownership and identity contract is genuinely usable by later VM and XDNA producers without renaming or wrapping a command-specific abstraction.

I am not asking this PR to implement the DAG scheduler, recursive orchestration, cache, final gather/link, VM compilation, XDNA lowering, artifact-contribution records, or bytecode resource storage. Those layers can follow. I would avoid shipping both live-module and serialized-source publication paths in this unreleased API: that would preserve two lifetime models and make the important invariant optional.

The end state I want from #455 is that command-program construction proves the first edge of a general model: a producer returns its own product and durable requirement edges while streaming independently owned, deterministic Loom sources for the requirements it can satisfy. Those sources can be deduplicated, queued, cached, compiled recursively in unrelated workspaces, and later bound back to every requesting parent. Command programs, VM programs, and XDNA array programs should all be able to participate without any one of them owning the abstraction.

benvanik:
Superseded by #491, which rebuilds this frontier around immutable request and product ownership while keeping scheduling and cache policy outside the compiler API. Closing this draft so the obsolete module-transfer surface is not reviewed or reused.

PR #462, opened 2026-08-28
[Infra] Run clang-tidy natively on Windows

benvanik:
The supported Windows LLVM distribution is statically linked and does not export the C++ symbols required by loadable clang-tidy modules. Build a dedicated `iree-clang-tidy.exe` with the IREE checks linked in, while preserving the existing module-based analyzer on other hosts.

Bazel imports the selected LLVM headers and static archives, builds the analyzer in its own graph, and supplies analysis actions with the configured compiler environment and toolchain inputs. CMake builds the equivalent executable for its compile-database lane. Presubmit also resolves Windows executable suffixes, launches extensionless LLVM Python tools correctly, and batches anchored file selectors below the Win32 command-line limit.

Wrapper-option tests now assert parsed wrapper state and structured backend argv instead of searching rendered shell commands. This keeps checkout paths and presentation formatting from impersonating flags while still proving that wrapper-only options do not leak into Bazel or CMake.

## Landing order

This branch remains based directly on `main`; it is not stacked on #461. The restored analyzer exposes the IOCP owning-status violation corrected by #461, so merge #461 first and then rerun this PR against the updated base.

## Reviewer notes

The highest-value review surfaces are the imported LLVM archive set in `llvm_repo.bzl`, the Windows-only analyzer selection in `BUILD.bazel`, and the presubmit command construction shared by the Bazel and CMake lanes.


benvanik:
This is unacceptable.

PR #506, opened 2026-09-01
[Runtime/VM] Introduce the replacement VM and Loom target

benvanik:
This introduces the replacement IREE VM as a compact, compiler-neutral runtime and makes Loom an ordinary producer of its Core ISA. The first production boundary is kernel launch configuration: one Loom compilation can produce a target-native kernel and a shared VM host companion, join them in one immutable kernel product by artifact-local ordinals, load and bind both once, evaluate launch geometry through the public VM API, and dispatch the native kernel without allocation, locking, lookup, or artifact traversal on the repeated path.

This is not a compatibility port of the retired VM. It replaces the old context/module-state/list model with a smaller composition model built around immutable modules and programs, process-local opaque state, caller-owned invocation storage, source-ordered 16-byte variants, verification at the external byte boundary, and one authoritative ISA description shared by the compiler, runtime, tooling, and documentation.

## Product boundary

The complete path established here is:

```text
authored Loom kernel + launch region
  -> one shared Loom module pipeline
  -> target-native executable artifact + VM launch companion
  -> immutable kernel product with exact artifact/function ordinals
  -> verify/load/link/create once
  -> invoke launch function
  -> dispatch the matching native kernel entry
```

Launch functions are deliberately the first consumer because they put both ends of the design under pressure. They are tiny enough that compiler and invocation overhead are immediately visible, but they require real target specialization, arithmetic, control flow, ABI lowering, serialization, external-byte verification, linking, process creation, and checked result publication. The product test crosses that entire boundary and verifies final device-visible output.

The retained value-facts launch evaluator is not removed here. This PR establishes and qualifies its replacement path without turning the initial landing into a deployment migration. Full HAL command-program execution and faithful multi-workitem kernel simulation build on the same Core runtime but are separate product boundaries.

## Runtime composition and lifetime model

The public runtime separates immutable work that can be shared across requests from mutable state that belongs to one execution lane:

| Object | Contract | Expensive work paid here | Concurrency and lifetime |
| --- | --- | --- | --- |
| `iree_vm_environment_t` | Thread-safe registry of host-owned ref-type provider tables | Namespace registration and module-load type lookup | Used while constructing modules; tables remain owned by their provider |
| `iree_vm_module_t` | Immutable, ref-counted implementation of the generic module ABI | Implementation validation, ref-type resolution, and immutable execution-table construction | Shareable across programs and threads |
| `iree_vm_program_t` | Immutable link of one executable module and zero or more library modules | Module sorting, callable-type canonicalization, import resolution, initializer selection, and process-state layout | Retains modules and is shared by any number of processes |
| `iree_vm_process_t` | One mutable instance of a linked program | One slab allocation, module-state attachment, and the executable module's optional `initialize` call | Contains no execution lock; the host serializes one process or creates independent processes |
| `iree_vm_invocation_t` | Reusable caller-owned execution stack and state machine | No dynamic growth; fixed capacity is selected by the caller | One externally serialized driver; placement initialization and pooled allocation are both supported |

Modules are pure interfaces. Bytecode, native C, dynamically loaded providers, and future implementations expose the same immutable descriptor and versioned vtable. Common VM code knows import/export/callable declarations, typed reflection metadata, and the byte count of each module's opaque process state; it does not know bytecode globals or any other implementation-private layout.

Program creation is the link boundary. It resolves every required import, permits explicitly optional imports to remain absent, interns structurally ordered callable types, and assigns stable slices in one process slab. Library modules have no implicit initialization. Process creation calls `attach_state` for every module, runs only the executable module's ordinary exported `initialize` function, seals the attached state, and publishes the process only after initialization succeeds. This makes initialization order a compiler/program concern and lets the initializer accept ordinary typed arguments such as device groups instead of relying on global host state.

Process failures retain the existing IREE status philosophy: a status is a terminal failure of the operation's scope, not suspension control flow and not `abort()`. The VM deliberately does not specify whether an invocation failure semantically taints application state; the program and embedding own that policy.

## Invocation ABI and ownership

Host calls use source-ordered arrays of fixed 16-byte `iree_vm_variant_t` values. A variant carries one scalar, one typed ref, or one function ref. Inline scalar constructors preserve exact bits, and ref construction names the lifetime transition explicitly (`borrowed`, `retained`, or `move`). Dynamic RPC-style callers can use the corresponding checked builders and extractors without constructing the implementer-only physical register packet.

A representative synchronous call is intentionally ordinary C:

```c
iree_vm_function_t function = iree_vm_function_null();
IREE_RETURN_IF_ERROR(iree_vm_process_lookup_function(
    process, IREE_SV("math"), IREE_SV("add"), &function));

iree_vm_variant_t arguments[] = {
    iree_vm_variant_from_i32(lhs),
    iree_vm_variant_from_i32(rhs),
};
iree_vm_variant_t results[1] = {};
iree_status_t status = iree_vm_invoke(
    invocation, function, iree_vm_variant_span_from_array(arguments),
    iree_vm_variant_span_from_array(results));
if (iree_status_is_ok(status)) {
  status = iree_vm_i32_from_variant(results[0], out_value);
}
iree_vm_variant_span_reset(iree_vm_variant_span_from_array(results));
return status;
```

The boundary rules are deliberately strong and simple:

- Argument and result counts must match exactly; there is no truncation or partial population.
- Once boundary preflight succeeds, every argument is consumed on every return, including failure and suspension.
- `iree_vm_invoke` touches result storage only when it returns OK; asynchronous start/resume touches it only on terminal completion.
- Successful publication initializes every result and transfers any result owners to the caller.
- A caller may zero-initialize result storage and use one unconditional cleanup path without inspecting partially published state.

The synchronous wrapper drives the same asynchronous core used by `iree_vm_invocation_start` and `iree_vm_invocation_resume`. Suspension is the ordinary `IREE_VM_EXECUTION_OUTCOME_SUSPENDED` value on an OK status. Resume accepts a new exact result span on every drive, validates it before mutation or provider entry, and permits spurious polling. Executors integrate the wake callback with their scheduler; a simple host may use the synchronous wrapper and accepts the same deadlock risks as any blocking wrapper over asynchronous dependencies.

Functions whose transitive implementation may suspend advertise that fact in their callable type. Passing a borrowed ref to such a root fails early, while an actual yield retains a runtime backstop against incorrect metadata. This keeps the common synchronous borrowed-ref path free of unconditional retain/release traffic without making suspension rely on compiler trust for memory lifetime.

## Generic module ABI and reflection

The module descriptor contains only common VM facts: stable module name, resolved canonical ref types, declaration counts, generic flags, and opaque process-storage size. The vtable contains execution, state attachment, and cold declaration/reflection queries. Bytecode-specific function records, globals, and decoded tables do not leak into the generic ABI.

Imports and exports remain static module declarations, while `iree_vm_function_t` is the resolved process-bound callable used for execution. Public reflection stays on the declarations where source aliases actually live: two exported names may refer to one deduplicated private function while retaining distinct documentation and metadata.

Reflection provides:

- sorted module/import/export typed metadata with `BOOL`, `I64`, `U64`, `F64`, UTF-8, byte-span, and forward-compatible opaque value encodings;
- exact machine signatures plus optional authored function types, field names, field authored types, and documentation;
- allocation-free structural queries when presentation is not needed; and
- the standard size-probe/caller-storage query pattern when a provider must synthesize presentation data.

Reference identity is canonical within an environment. A type is `(namespace, type_name)`, lowered from authored types such as `hal.buffer` to VM types such as `vm.ref<hal, buffer>` without imposing source-language version syntax. Provider tables are append-only, size-prefixed structures in read-only storage. A host compiled against an older table can load newer modules as long as every module agrees on the newer types; the host does not need fixed caller storage for type definitions it never touches.

## Authoritative bytecode specification

`runtime/src/iree/vm/bytecode/spec/model/` is the sole editable authority for the module format, Core ISA, optional HAL page, wire encodings, state effects, and revision history. Every entity carries a `since` revision, allowing one model to project historical ISA views without copying version directories.

One generation action produces purpose-built artifacts for each consumer:

- natural-layout C wire declarations and private layout assertions;
- compact Loom descriptor, lowering, constraint, and encoding tables;
- runtime verifier, interpreter, and inspection tables;
- disassembler tables; and
- the published human ISA reference.

Only the concise natural-layout wire headers are checked in so a runtime-only build does not require Python. Generated compiler/runtime implementation projections remain in build output, and freshness tests reject drift. The full Markdown ISA reference is generated into the Loom documentation site rather than checked into the repository.

Core 0.0 contains 183 executable records: 171 generated descriptor-backed instructions and 12 structural control/ABI encodings. The compiler, verifier, interpreter, disassembler, and coverage tests all derive their inventory from the same model. Core covers arbitrary CFG, scalar integer and floating operations, conversions, direct and indirect calls, optional imports, function values, value/ref/function registers, deterministic spills, overflow ABI carriers, mutable and set-once immutable globals, constants, rodata, buffers, atomics, explicit failure, and suspension. The separately specified HAL page is optional and does not enter a Core-only runtime link.

## Format, verification, and execution

The module container is little-endian, eight-byte aligned, and designed for mmap and natural C overlays on alignment-sensitive hosts. A fixed header and local section descriptors carry exact counts and byte ranges for strings, ref types, signatures, callable types, imports, exports, functions, constants, globals, rodata, reflection, and code. Local counts remain authoritative for local verification; the header summary exists to allocate runtime state without rescanning the image.

Module creation is the external trust boundary. It validates section structure, strict table ordering, local ordinals, callable nesting, function layouts, CFG branch targets, register and stack ranges, instruction semantics, ownership masks, and all other facts knowable from immutable bytes. It resolves ref types and builds one immutable module slab. Program creation then performs only cross-module work, and process creation performs only per-instance work.

The interpreter runs the verified immutable bytes directly. It does not decode a second instruction copy and does not repeat static validation on each instruction. Checks in the execution loop are limited to live state, such as a dynamic buffer access or indirect callable value. Mutating verified image storage violates the host-owned storage contract; defeating hostile concurrent mutation would require per-instruction re-verification and is intentionally outside this in-process VM model.

## Loom lowering and emission

The VM is a normal Loom target beside AMDGPU, SPIR-V, LLVM IR, WebAssembly, and x86. It reuses the shared source-to-Low machinery, scheduling, liveness, register allocation, move sequencing, and spilling. The target owns only VM contracts: source-to-opcode mappings, ABI layout, semantic materializers for operations that are not a direct mapping, and module emission.

Simple mappings are generated compact read-only tables. For example, the same source scalar operation maps by machine scalar width without a handwritten switch forest. More complex families such as vectors, memory, calls, CFG switches, constants, and launch ABI materialization use focused C implementations because they construct real plans rather than select one opcode.

Ordinary authored control flow:

```loom
func.def public target(@target) @conditional_return(
    %condition: i1, %true_value: i32, %false_value: i32) -> (i32) {
  cfg.cond_br %condition, ^then, ^else
^then:
  func.return %true_value : i32
^else:
  func.return %false_value : i32
}
```

passes through shared target-Low allocation and emits compact branch-oriented bytecode:

```text
+0000  control.block
+0004  control.branch.unless.s16 condition_v8=0 target_rel16=3
+0008  control.block
+0012  value.copy dst_v8=0 src_v8=1
+0016  control.return
+0020  control.block
+0024  value.copy dst_v8=0 src_v8=2
+0028  control.return
```

Target-neutral bulk memory syntax remains equally direct:

```loom
buffer.copy %source[%source_offset], %target[%target_offset], %byte_length
```

and function values stay explicit in Loom rather than becoming a hidden VM convention:

```loom
%target = func.address @identity : func.ref<(i32) -> (i32)>
%result = func.call.indirect %target(%value) : (i32) -> (i32)
```

`func.ref<yieldable (...) -> (...)>` is the distinct suspending type; synchronous function refs use the shorter default form. The emitted indirect-call record carries the expected callable type ordinal so the runtime can check a dynamic target locally without rebuilding signatures.

The serializer walks the selected module once. It emits into a segmented byte sequence using 32 KiB blocks, reserves directory and function rows, writes each payload once, and patches the reserved rows after final offsets are known. It has no sizing serialization pass, contiguous staging buffer, or grow-and-copy loop. A consumer that requires contiguous bytes explicitly clones the sequence; file and callback sinks enumerate segments directly.

## Kernel launch-config integration

A kernel launch region is ordinary Loom source and remains target-aware:

```loom
kernel.def target(@gfx1100) @unchecked_1d(%element_count: index) {
  %c1 = index.constant 1 : index
  %rounding = index.constant 255 : index
  %workgroup_size = index.constant 256 : index
  %rounded_count = index.add %element_count, %rounding : index
  %workgroup_count = index.div %rounded_count, %workgroup_size : index
  kernel.launch.config workgroups(%workgroup_count, %c1, %c1)
      workgroup_size(%workgroup_size, %c1, %c1) : index
} launch() {
  kernel.return
}
```

The mixed-target compiler keeps the device entry and extracted host launch function in one source module so parsing, linking, specialization, and module passes are shared. After the common pipeline, the target-native emitter and VM emitter select their own roots. Product construction joins those projections once by stable compiler function version and publishes artifact-local ordinals. Runtime loading uses those ordinals directly; name lookup remains available only for discovery and independently authored artifacts.

The public Loom C wrapper owns a loaded immutable VM program, one process, pre-bound launch functions, source-ordered argument type rows, and a reusable 16 KiB invocation inside one slab. A repeated call validates the token and exact raw workload-bit count, writes scalar variants, invokes the VM function, validates all launch results, and publishes the complete `loomc_launch_config_t` only on success. It performs no allocation, lock, name lookup, artifact scan, or cache lookup.

## Performance

The performance contract is deliberately split between compile/readiness latency and the repeated invocation path.

| Product boundary | Representative optimized median |
| --- | ---: |
| Device kernel compilation only | 10.780 us |
| Same compilation with VM launch companion | 35.231 us |
| Incremental companion compilation and publication | 24.451 us |
| Bytecode module creation with complete verification | 240.9 ns |
| One-module program linking | 105.9 ns |
| One-module process creation without an initializer | 23.6 ns |
| Loaded program and function binding | 0.765 us |
| Checked bytecode launch-config invocation | 57.7 ns |
| Warm allocation, locks, and name lookups | 0 |

The checked invocation is well below the 100 ns product stop line and faster than the retained value-facts evaluator on the matched witness. The compile path remains the larger cost and is reported honestly: the VM companion adds approximately 24.5 us for the five-operation launch function. The serializer and publication are not hiding a second pass or size-dependent copy; remaining time is primarily shared Loom normalization, facts, liveness, allocation, and frame construction. Those are general compiler optimization opportunities, not grounds for a launch-only assembler or a second semantic pipeline.

## Binary footprint

The canonical stripped `LoomC bytecode to HSACO (all AMDGPU)` witness grows from 8,124,600 to 8,509,240 bytes: +384,640 bytes (+375.6 KiB, +4.73%). The emitted HSACO remains byte-identical. During this audit, provider-owned legalizer rows were separated from composed registry entries and an unused per-entry payload was removed; that reduced the canonical witness by 5,440 bytes and also shrinks each composed registry allocation.

An exact symbol-bearing A/B under the same shipping build contract attributes the +385,009-byte allocated-section delta as follows:

| ELF contribution | Delta |
| --- | ---: |
| Executable code | +271,066 B |
| Read-only data | +77,968 B |
| Relocated writable data | +12,728 B |
| Relocations | +14,520 B |
| Unwind tables | +10,288 B |
| Alignment and small dynamic sections | -1,561 B |
| **Total allocated sections** | **+385,009 B** |

Normalized symbols attribute 247,760 bytes to the complete VM compiler target and 96,826 bytes to shared compiler prerequisites and refactors; 40,423 bytes are unnamed constants plus relocation, unwind, alignment, and ELF metadata. The VM-owned closure is 188,406 bytes of executable code, 51,322 bytes of read-only data, and 8,032 bytes of relocated/writable data.

The 44,128-byte generated Core descriptor projection is dense and index-based: 11,072 bytes of 64-byte descriptors, 10,012 bytes of BSTRING data, and compact operand/immediate/assembly/enum arrays. It is not a full-width pointer table per operation. All other VM lowering, constraint, provider, and ABI tables total 11,525 bytes. The remaining large symbols are real vector/memory/ABI/layout/emission implementations rather than duplicated opcode switches.

The separately linked runtime layers contain 145,609 named bytes for generic VM composition plus the structural bytecode module, verifier, and executor. Optional inspection and dump/disassembly add 25,926 bytes. Dependency checks prove that dump/inspection do not retain the executor. This keeps Core-only embeddings and tooling-disabled builds from paying for presentation tables or disassembly strings.

## Verification surface

The test shape follows production boundaries instead of maintaining a second bytecode assembler:

- specification tests validate revision rules, encoding constraints, and every generated projection;
- wire freshness tests reject stale or unexpected checked-in declarations;
- runtime unit tests exercise individual record semantics, ownership, cancellation, suspension, FPU scoping, module reflection, and malformed external bytes;
- common source-Low corpus tests run the same target-neutral programs through VM, AMDGPU, SPIR-V, x86, and other applicable targets;
- VM-specific `.loom-test` cases cover ABI layout, spills, imports, function values, metadata, CFG, suspension, and emitted module structure;
- compiler-to-runtime execution tests load actual Loom-produced modules and invoke them through the public API;
- the kernel-product test evaluates a VM launch companion and verifies the matching native kernel's device-visible result; and
- invocation, structured-instruction, and raw-module libFuzzer harnesses cover the public state machine, complete Core record surface, loader, verifier, inspection, and dump paths.

The retained-corpus qualification is complete. Twelve coverage-instrumented ASAN workers ran for 32,412 worker-seconds (9.003 aggregate CPU-hours) and executed 121,071,083 inputs across the invocation state machine, structured Core records, and raw module factories. Every worker exited 0 with no crash, timeout, leak, sanitizer report, unexpected status, or artifact. Campaign-final coverage counters were 833 edges and 3,283 features for invocation, 2,961 edges and 4,471 features for structured Core records, and 2,775 edges and 7,040 features for raw module factories.

LibFuzzer reduced the 7,229-file retained corpus (145,481 bytes) to 1,850 files (17,680 bytes). Every minimized corpus replays cleanly and preserves all normal-mode coverage edges. This sustained qualification complements the deterministic malformed-input and integration tests above; it does not replace them.

## Review map

The most useful review order is:

1. `runtime/src/iree/vm/bytecode/spec/` for the format/ISA authority and generation boundaries.
2. `runtime/src/iree/vm/{module,program,process,invocation,variant,ref}.*` for the generic composition, lifetime, and execution contracts.
3. `runtime/src/iree/vm/bytecode/` for one-time verification, immutable module construction, and direct interpretation.
4. `loom/src/loom/target/arch/vm/` for source-to-Low lowering and shared compiler integration.
5. `loom/src/loom/target/emit/vm/` for module selection, layout, ABI encoding, and one-pass segmented serialization.
6. `loom/binding/c/target/vm/` and the kernel-product tests for the user-visible launch boundary.

The commit stack is dependency ordered from the specification and generic runtime through complete Core execution, the Loom target, and product integration. The runtime and Loom VM READMEs provide shorter entry points, while the generated VM bytecode reference is the instruction-by-instruction authority.


benvanik:
The VM portion of this draft is superseded by #553.

#553 reconstructs and independently qualifies the standalone Core VM boundary: the authoritative Core 0.0 specification, generic environment/module/program/process/invocation APIs, bytecode format and verifier, trusted loader, interpreter, reflection, and `vm-as`/`vm-dis`. Its history intentionally excludes the Loom VM compiler target and HAL ISA.

Reviewers should use #553 as the canonical review surface for the VM/runtime work represented here. Loom lowering and emission, launch-config product integration, and the HAL ISA remain separate follow-on work; #553 does not supersede those portions of this draft.

PR #519, opened 2026-09-03
[Loom] Jointly select gfx11 result publication layout

benvanik:
RDNA3 f16 WMMA result publication is now selected together with the
instruction's physical fragment layout. Column-major f16 results can use the
all-lane scalar publication shape instead of paying for cross-lane packing and
a partial-wave store.

The matrix layout catalog now represents exact alternatives as a canonical
family. The gfx11 f16 alternative retains the source-role payload layouts,
reverses the physical WMMA operands, and transposes the accumulator/result
coordinates. Interpreting the semantic RHS as the hardware LHS produces
`B^T`; interpreting the semantic LHS as the hardware RHS produces `A^T`.
The instruction therefore computes `B^T A^T = (A B)^T`, and the transposed
result coordinates restore the semantic `A B` mapping.

Publication planning scans the function's actual matrix contracts and exact
fragment stores, then ranks compatible layouts by subgroup memory-region cost
and target instruction issue cost. Descriptor availability supplies the target
capability boundary. Naturally contiguous packed publication remains preferred,
cross-lane publication remains available when it is genuinely cheaper, and an
ambiguous source contract conservatively retains its canonical layout.

The tracked source-low proof uses real f16 fragment loads and WMMA, checks the
reversed instruction operands, and checks direct b16 stores from all 32 lanes.
The full High workload consequently lowers through the default pipeline to 48
scalar conversions and 48 b16 stores without DPP/DS exchange or publication
exec-mask manipulation.

## Reviewer Notes

The highest-value review surfaces are the symmetry validation in the generated
layout catalog, the shared instruction/publication selector, and the narrowed
store issue-cost accounting.


benvanik:
Closing because review exposed that this is an insufficiently designed fix.

The change adds whole-function recursive rescanning and generic compiler surface for a single gfx11/f16 WMMA realization. That violates Loom's single-owner analysis and planning model, and it does not establish a compiler-wide abstraction across fragment families, types, and targets. The validated reproducer and semantic transpose proof remain useful evidence, but this implementation will not be merged.

A replacement design must consume canonical analysis once, keep target packages as data and capability providers instead of nested compilers, and justify every change outside a target directory as an uplift used by the whole compiler before implementation resumes.

PR #528, opened 2026-09-03
[Loom] Retain source lowering decisions once

benvanik:
Source-to-Low lowering now builds one immutable view of a source function and retains the physical decisions derived from it. Target legality, descriptor selection, type mapping, and operation lowering can share source topology, dataflow results, and native-representation choices instead of independently walking or reinterpreting the source IR.

## One compiler-owned analysis boundary

The source-program index assigns stable ordinals to operations, blocks, regions, values, and uses and records typed control-flow, alias, tied-value, and operation-local value-flow edges. It is rebuilt only when rewriting invalidates the source function. Consumers use dense records and spans rather than recursive callbacks or target-owned CFG mirrors.

A target-independent monotone solver evaluates compact physical-value transfer tables over that retained program. It supports forward and backward strata, phase projections, cyclic structured control, and accumulating rejection evidence while keeping dependency scheduling, convergence, and result storage inside the common compiler.

## Native representation planning

The retained representation planner treats physical source values as connected components. Generated providers declare finite exact or opaque representations, operation-local candidates, capability predicates, and Low descriptor recipes. The planner intersects compatible domains, rejects inseparable alternatives, compares structured descriptor costs, and retains one value and operation decision before source types are mapped.

Candidate preparation happens once per handled operation. Compact table rows match against that immutable summary, choose any target-local realization, and retain ranked rejection evidence for diagnostics. Later lowering queries the same selected candidate and prepared facts; it does not reconstruct why the candidate won or why an optional group was unavailable.

Generated lower-rule guards can require either a representation group or one exact candidate. Contract-only legality defers those late-bound guards, while final source lowering resolves them against the retained plan.

## Shared descriptor evidence

Descriptor resource accounting moves out of the scheduler into a common Low cost component. Scheduling and representation selection now use the same capacity-normalized evidence for instruction multiplicity, dependency depth, memory effects, and durable register pressure. Unknown or incomparable models retain a deterministic canonical choice instead of accepting a target-authored scalar score.

## Scope

This is the common compiler and synthetic test-target foundation. It does not change a production target's matrix representation by itself. SPIR-V and AMDGPU integrations are stacked separately so each production policy and generated table can be reviewed against the same already-proven lifecycle.

## Reviewer Notes

The highest-value review surfaces are source-program invalidation and typed edge ownership, monotone phase semantics in the dataflow solver, component separability and cost comparison in the representation planner, and the selected-versus-rejected retention contract exposed to lowering.


benvanik:
Closing this draft because the design fails its required code-size and compile-throughput gates.

With optimized ThinLTO, the foundation alone adds 46,448 bytes of `.text` to targetless and SPIR-V compiler binaries and 46,192 bytes to AMDGPU; total ELF section growth is 49,132-53,240 bytes. The completed replacement remains larger: +53,220 bytes targetless, +69,640 bytes SPIR-V, and +155,628 bytes AMDGPU.

Balanced fixed-iteration benchmarks also reject it. Foundation is 7.71% slower on the SPIR-V compile workload and 2.59% slower on AMDGPU. The completed stack is 7.46% and 3.60% slower, respectively. It also raises SPIR-V workspace blocks from 13 to 15 and doubles AMDGPU oversized allocations per kernel.

The mechanism is structural: `source_program` duplicates topology already owned and traversed by the IR and `source_plan`; dataflow materializes a second runtime equation/watch graph; representation planning allocates dense state for every value and node; legalization rebuilds that stack after rewrites; source lowering builds it again; and the query API duplicates scope identity across a lowering context and environment, then validates the copies by pointer equality.

Resolution: this is an insufficient design, not a patch to optimize in place. The replacement will contract query ownership, extend the existing source plan instead of adding a parallel program, keep reusable solvers link-selective, and generate descriptor-derived candidate costs into build-verified tables. Each slice must pass size and throughput gates before target integration grows on top of it.

PR #557, opened 2026-09-09
[HAL] Realize specialized hardware queues and HIP execution domains

benvanik:
Adds exact, specialized hardware queues to HAL and uses them to implement HIP execution-resource domains without leaking AMD-specific concepts into generic interfaces. This is the first major user of the queue-object foundation from #550: callers can request a precise hardware realization, inspect what they received, and keep that queue as a normal HAL ref object.

## Newly supported capabilities

| Capability | HAL model | HIP/libhrx result |
| --- | --- | --- |
| Exact execution-resource confinement | Queue families advertise canonical execution resources; acquisition accepts an exact sorted resource set | CU-mask streams, split device resources, green contexts, and execution-context streams run only on the selected CUs/WGP groups |
| Native scheduling priority | Priority is an immutable acquired-queue property | Priority streams use native HSA queue priority instead of a software hint |
| Cooperative grid execution | Cooperative support is an immutable queue feature and cooperative execution is an explicit dispatch flag | `hipLaunchCooperativeKernel` and `hipModuleLaunchCooperativeKernel` use GWS or memory-backed grid synchronization as appropriate for the target |
| Executable-specific occupancy | A queue query combines the loaded function, exact queue realization, and launch shape | HIP occupancy APIs and cooperative-launch validation account for the selected resources, registers, LDS, dynamic shared memory, and target limits |
| Dynamic hardware queues | `iree_hal_device_acquire_queue` realizes an exact hardware queue on demand | Expensive and specialized HSA queues are created only when requested |
| Multi-device enqueue | Per-device work remains an ordered host-side transaction above HAL | `hipExtLaunchMultiKernelMultiDevice` validates and commits a complete set of launches without inventing a generic HAL group-dispatch ABI |
| Replayable queue identity | Acquisition records the family and exact immutable queue parameters | Replayed workloads reconstruct specialized queues instead of falling back to a default realization |

## HAL contract

Queue families now describe four independent facts: supported command categories, selectable execution resources and their constraint groups, scheduling priorities, and specialized queue features. Those are immutable hardware capabilities, not backend modes or stringly target names.

`iree_hal_device_acquire_queue` is exact and all-or-nothing. It returns a hardware queue with the requested family, priority, feature set, and execution-resource set, or fails without publishing an output. HAL performs no generic queue caching, pooling, leasing, or virtual scheduling. Drivers remain free to share a native realization where the exact requested properties permit it.

```c
const iree_hal_queue_family_t* queue_family =
    iree_hal_device_queue_family(device, family_ordinal);

const iree_hal_queue_execution_resource_ordinal_t resource_ordinals[] = {0, 1};
iree_hal_queue_params_t queue_params;
iree_hal_queue_params_initialize(&queue_params);
queue_params.priority = requested_priority;
queue_params.execution_resources =
    (iree_hal_queue_execution_resource_list_t){
        .count = IREE_ARRAYSIZE(resource_ordinals),
        .ordinals = resource_ordinals,
    };

iree_hal_queue_t* queue = NULL;
IREE_RETURN_IF_ERROR(iree_hal_device_acquire_queue(
    device, queue_family, &queue_params, &queue));

// Use explicit semaphore edges to order work on this queue.
// Queue acquisition selects hardware; it does not make HAL queues FIFO.

iree_hal_queue_release(queue);
```

An empty execution-resource list and an explicit canonical enumeration of every family resource both mean the complete family resource set; acquisition canonicalizes both to the empty representation. Acquired queues report their achieved immutable properties directly and have distinct frontier incarnations, so independently acquired queues cannot accidentally alias synchronization state.

Cooperative dispatch deliberately requires both sides of the contract:

| Contract | Meaning |
| --- | --- |
| `IREE_HAL_QUEUE_FEATURE_FLAG_COOPERATIVE_DISPATCH` | This exact hardware queue can execute cooperative grids |
| `IREE_HAL_DISPATCH_FLAG_COOPERATIVE` | This particular dispatch requires grid-wide synchronization |

This keeps an operation requirement on the operation instead of mutating global queue state, which is essential because HAL queues are not FIFO.

## HIP execution-resource model

libhrx projects HAL's canonical resource ordinals into the current HIP ABI. Copyable `hipDevResource` values carry pointer-free identities into immutable device-owned tables, and process-unique incarnations prevent stale values from aliasing a recreated device. Resource splitting preserves arbitrary exact subsets rather than rounding them into a count or affinity hint.

A HIP execution domain follows the native API shape while terminating in an exact HAL queue:

```text
hipDeviceGetDevResource
  -> hipDevSmResourceSplit / hipDevSmResourceSplitByCount
  -> hipDevResourceGenerateDesc
  -> hipGreenCtxCreate
  -> hipExecutionCtxStreamCreate
  -> exact iree_hal_queue_t retained by the stream
```

The corresponding stream/context APIs now have concrete behavior:

| HIP surface | Behavior |
| --- | --- |
| `hipExtStreamCreateWithCUMask` / `hipExtStreamGetCUMask` | Converts a valid CU mask to the queue family's exact canonical resource set and verifies native confinement |
| `hipStreamCreateWithPriority` and execution-context stream creation | Requests native queue priority and reports failure if the exact priority cannot be realized |
| Execution-resource split and descriptor APIs | Produce exact, copyable subsets with explicit remainder handling |
| Green/execution context events and synchronization | Build explicit semaphore fan-in and fan-out across the context's streams |
| Cooperative kernel launch APIs | Select a cooperative realization for the operation and bridge it through the logical stream timeline |
| Runtime and module occupancy APIs | Query the loaded executable on the exact target queue instead of duplicating AMD occupancy formulae in libhrx |
| `hipExtLaunchMultiKernelMultiDevice` | Captures arguments, locks participating streams in stable identity order, appends all launches without host-thread interleaving, and then submits the complete transaction |

Single-device cooperative launch may use a different physical queue than the logical HIP stream. Explicit semaphore edges bridge that work through the stream timeline, preserving HIP ordering without relying on FIFO submission.

The deprecated `hipLaunchCooperativeKernelMultiDevice` entry points remain unsupported. CUDA removed this API family, and implementing cross-device grid synchronization would distort both the HAL queue contract and the meaning of the live non-cooperative AMD extension.

## AMDGPU realization

AMDGPU lowers exact requests to native HSA queue descriptors, applies priority and CU/WGP confinement before publishing the queue, requires native mask programming to succeed, and sizes queue sidecars from achieved native capacity. Ordinary queues are created on demand. The expensive cooperative HSA queue is acquired lazily and shared at its physical-agent scope; requests that cannot currently be realized exactly, such as masked or non-normal-priority cooperative queues, fail instead of silently broadening the request.

The synchronization mechanism remains private to AMDGPU:

| Target family | Cooperative grid mechanism |
| --- | --- |
| Supported GFX9/GFX10 targets | Native GWS initialization |
| gfx942, gfx950, GFX11, and GFX12 | Memory-backed OCKL grid barrier |

Direct dispatches allocate epoch-reclaimed state per operation. Every reusable command-buffer execution materializes fresh queue-owned grid state. Explicit barrier chains prevent later initialization from overtaking an earlier grid without pretending the queue is FIFO.

## Prior work folded into this design

| Prior PR | What it established | How this PR carries it forward |
| --- | --- | --- |
| #351 by @AWoloszyn | Explored lazy AMDGPU queue provisioning, exact masks, and the HIP execution-context requirements | Replaces affinity slots and lease-shaped machinery with generic exact hardware-queue acquisition while preserving Andrew's adapted HIP execution-resource authorship |
| #444 by @zjgarvey | Implemented and hardware-tested fixed native CU-masked queues, including validation before queue publication | Folds the native mask realization and tests into the generic AMDGPU acquisition path while preserving Zach's adapted commit authorship |

Thank you to Andrew and Zach for doing the requirements discovery and hardware work that made this design concrete. #550 gave that work a generic HAL queue-object foundation; this PR supersedes #351's exploratory design and folds in #444 without carrying their AMDGPU-private API shapes forward.

## Scope boundary

This PR establishes the exact hardware-queue substrate and implements the current HIP compatibility consumers. It intentionally leaves the upcoming AMD GPU/XDNA placement-domain model, allocator split, engine taxonomy, and extensible target-specific queue operations to the separate HAL redesign, where they can replace or extend this substrate cleanly instead of being forced through affinity masks or private-driver reacharounds.

It also does not add virtual queues, mutable queue modes, generic grouped dispatch, or cross-device cooperative synchronization. Those are materially different contracts rather than hidden variants of hardware queue acquisition.

## Production coverage

| Layer | Exercised behavior |
| --- | --- |
| Generic HAL queue CTS | Provisioned-queue behavior on production Task, Vulkan, and AMDGPU devices; exact acquisition and validation; independent queue ordering; cooperative direct and reusable-command-buffer dispatch; exact dispatch-concurrency queries |
| AMDGPU CTS | Exact physical execution-resource IDs, priority/mask realization, replay, concurrency, and driver-specific failure paths; driver-owned HIP/OCKL payloads instantiate the generic cooperative queue CTS |
| libhrx HIP tests | Resource splitting and descriptors, execution contexts, stream priority and CU masks, events, occupancy, cooperative launch, and multi-device launch transactions |

## Reviewer guide

| Review area | Primary surface | Question to answer |
| --- | --- | --- |
| Generic contract | `runtime/src/iree/hal/queue.h`, `device.h`, and `device_spec.h` | Are exact queue requirements separated cleanly from placement visibility and semaphore ordering? |
| AMDGPU realization | Native queue construction, host queue dispatch, and cooperative grid-state ownership | Can any request publish a queue weaker than the exact contract, or can grid state outlive/reuse the wrong operation? |
| libhrx ownership | Execution-resource tokens, descriptors, contexts, streams, and events | Are identities copyable and stale-safe, and are queue/context lifetimes explicit? |
| HIP behavior | CU-mask, priority, cooperative, occupancy, and multi-device entry points | Does each HIP semantic terminate at the right layer without leaking AMD concepts into HAL? |
| CTS | Generic queue, AMDGPU, replay, and HIP coverage | Do tests exercise production devices and explicit dependency edges rather than mocks or FIFO assumptions? |


AWoloszyn (commented):
Looks good, and this should cover my future use-cases well too

zjgarvey:
Some review findings (1 high, 3 med).

1. **[HIGH] Normalize an explicitly enumerated complete resource set before cooperative queue acquisition.** The HAL contract says an empty execution-resource list selects the complete family (`runtime/src/iree/hal/queue.h:68-72`), but the HIP resource table expands that form to the family's count and stores every ordinal explicitly (`libhrx/src/binding/common/execution_resource.c:66-71,228-233`). `hipExecutionCtxStreamCreate` forwards those ordinals (`libhrx/src/binding/hip/execution_context.c:514-540`), and cooperative selection preserves them (`libhrx/src/binding/common/stream.c:355-384`). AMDGPU then rejects every cooperative request whose resource count is nonzero (`runtime/src/iree/hal/drivers/amdgpu/logical_device.c:2368-2373`). Consequently, a primary execution-context stream or a green-context stream covering the whole device returns `hipErrorNotSupported` for an otherwise supported cooperative launch solely because “all resources” has the explicit representation. Please canonicalize an explicit complete set to the empty/full form (or compare complete-set semantics in the backend), keep achieved-queue equality consistent with that choice, and add end-to-end cooperative launch coverage for primary and explicit-full green contexts while retaining rejection for genuinely partial sets.

2. **[MEDIUM] Run `hip_execution_resource_api_test` in an AMD-GPU test lane.** The new target has no `runtime-resource=amd-gpu` label in Bazel or CMake (`libhrx/src/binding/hip/BUILD.bazel:316-331`; `libhrx/src/binding/hip/CMakeLists.txt:365-384`). The GPU CMake jobs select only that label (`.github/workflows/ci_core_linux.yml:110-143`), whereas the unlabelled CPU run can pass after the fixture skips when `hipInit` fails (`libhrx/src/binding/hip/hip_execution_resource_api_test.cc:441-444`). This leaves the public resource/context API assertions compiled but not exercised with hardware. Please add the GPU resource labels in both build systems and make device initialization failure fatal in the resource-guaranteed lane.

3. **[MEDIUM] Require two devices in the test of `hipExtLaunchMultiKernelMultiDevice`.** The only test gathers up to two matching devices but accepts `launch_count == 1` (`libhrx/cts/tests/hip/module_execution_test.cpp:546-583`), and the declared gfx942 lane is a one-GPU runner (`.github/workflows/ci_core_linux.yml:110-122`). A one-entry call cannot exercise distinct-device validation, symbol resolution in a second device context, or a cross-stream batch transaction. Please add a case on a lane that guarantees two matching visible devices, require two launches there, and cover a later-member preflight failure without publishing an earlier member.

4. **[MEDIUM] Give the multi-device synchronization checks a non-synchronizing oracle.** The default-flags case creates fresh streams, so no pending prefix tests pre-synchronization, then reads each result with synchronous `hipMemcpy` (`libhrx/cts/tests/hip/module_execution_test.cpp:585-596`). That copy path synchronizes the context before reading (`libhrx/src/binding/hip/api.c:6786-6789`), so it supplies the post-synchronization the test is intended to verify. Please use deterministic gated work before and during the launches plus `hipStreamQuery` or another non-synchronizing observation to distinguish all four pre/post-sync flag combinations before cleanup.

benvanik:
Thanks, Zach — these were useful findings. I addressed the three issues we can prove on the current runner fleet:

1. Explicit complete resource sets are now canonicalized centrally by `iree_hal_device_acquire_queue` from `[0, N)` to the empty/full representation before driver dispatch, and achieved queue properties are checked against that canonical form. The shared queue CTS acquires and dispatches through an explicitly enumerated complete cooperative set. The HIP end-to-end test now covers cooperative launch on a primary execution-context stream and an explicit-full green-context stream, while genuinely partial cooperative sets remain unsupported.

2. `hip_execution_resource_api_test` now carries the AMD GPU resource label in both Bazel and CMake, and `hipInit` failure is fatal in that hardware-guaranteed test. It is running in the GPU lanes instead of passing through a CPU-side skip.

3. I retained the `numDevices == 1` case because it is a valid HIP call and still exercises registration, argument packing, enqueue, synchronization, and execution through `hipExtLaunchMultiKernelMultiDevice`. The test automatically uses two distinct matching devices when they are available. This repository did not previously have a guaranteed multi-GPU CI runner, so I am tracking that infrastructure addition together with a true later-member preflight-failure test that proves no earlier member is published. I do not want to fake that coverage by submitting twice to one physical device.

4. All four pre/post synchronization flag combinations now use deterministic host/device gates and `hipStreamQuery` as the non-synchronizing oracle. The synchronous copy happens only after those observations and explicit cleanup synchronization.

Thank you again for the careful pass.

PR #562, opened 2026-09-10
[HIP] Implement stream-ordered memory operations

AWoloszyn:
Implement 32-bit and 64-bit stream waits, writes, and batched memory operations using generic HAL queue atomics.

Value waits lazily acquire a context-owned exact queue so an unresolved predicate cannot block producer kernels multiplexed onto the ordinary operation queue. A nonblocking host-call gate carries wait completion back to the stream timeline without occupying that producer queue.

Resolve and retain each target allocation before submission, select device or system atomic scope from its visibility, validate every batch entry before enqueueing, and report support only when the queue family exposes the required atomic, host-call, and dynamic-acquisition capabilities.

### DIFF
--- a/libhrx/src/binding/common/BUILD.bazel
+++ b/libhrx/src/binding/common/BUILD.bazel
@@ -67,6 +67,7 @@ hrx_cc_library(
         "registry.c",
         "stream.c",
         "stream_host_call.c",
+        "stream_value.c",
         "tls.c",
     ],
     hdrs = [
@@ -82,6 +83,7 @@ hrx_cc_library(
         "module.h",
         "occupancy.h",
         "stream.h",
+        "stream_value.h",
         "tls.h",
     ],
     includes = [
@@ -151,6 +153,17 @@ hrx_cc_test(
         ":common",
         "//libhrx/src/libhrx:hrx_static",
         "//runtime/src/iree/base",
+        "//runtime/src/iree/hal",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
+    name = "stream_value_test",
+    srcs = ["stream_value_test.cc"],
+    deps = [
+        ":common",
         "//runtime/src/iree/hal",
         "//runtime/src/iree/testing:gtest",
         "//runtime/src/iree/testing:gtest_main",
--- a/libhrx/src/binding/common/CMakeLists.txt
+++ b/libhrx/src/binding/common/CMakeLists.txt
@@ -64,6 +64,7 @@ iree_cc_library(
     "module.h"
     "occupancy.h"
     "stream.h"
+    "stream_value.h"
     "tls.h"
   TEXTUAL_HDRS
     "${IREE_ROOT_DIR}/build_tools/amdgpu/elf_machine_map.inl"
@@ -88,6 +89,7 @@ iree_cc_library(
     "registry.c"
     "stream.c"
     "stream_host_call.c"
+    "stream_value.c"
     "tls.c"
   DEPS
     ::execution_resource
@@ -168,6 +170,19 @@ iree_cc_test(
     iree::testing::gtest
     iree::testing::gtest_main
     libhrx::src::libhrx::hrx_static
+    libhrx_defs
+)
+
+iree_cc_test(
+  NAME
+    stream_value_test
+  SRCS
+    "stream_value_test.cc"
+  DEPS
+    ::common
+    iree::hal
+    iree::testing::gtest
+    iree::testing::gtest_main
     libhrx_defs
 )
 
--- a/libhrx/src/binding/common/buffer_table_test.cc
+++ b/libhrx/src/binding/common/buffer_table_test.cc
@@ -271,6 +271,40 @@ TEST(BufferTableTest, LookupRangeOverflow) {
   iree_hal_streaming_buffer_table_free(table);
 }
 
+TEST(BufferTableTest, RetainedRangeSurvivesTableRemoval) {
+  iree_allocator_t allocator = iree_allocator_system();
+  iree_hal_streaming_buffer_table_t* table = nullptr;
+  IREE_ASSERT_OK(iree_hal_streaming_buffer_table_allocate(allocator, &table));
+
+  hrx_buffer_t buffer = nullptr;
+  IREE_ASSERT_OK(
+      iree_allocator_malloc(allocator, sizeof(*buffer), (void**)&buffer));
+  memset(buffer, 0, sizeof(*buffer));
+  iree_atomic_ref_count_init(&buffer->ref_count);
+  buffer->size = 4096;
+
+  constexpr uint64_t kDevicePointer = UINT64_C(0x100000000);
+  IREE_ASSERT_OK(BufferTableStatus(hrx_buffer_table_insert(
+      table, kDevicePointer, /*host_ptr=*/nullptr, buffer->size, buffer,
+      /*user_data=*/nullptr)));
+
+  hrx_buffer_t retained_buffer = nullptr;
+  size_t offset = 0;
+  IREE_ASSERT_OK(BufferTableStatus(hrx_buffer_table_find_range_retain(
+      table, kDevicePointer + 32, /*size=*/8, &retained_buffer, &offset)));
+  EXPECT_EQ(buffer, retained_buffer);
+  EXPECT_EQ(32u, offset);
+  EXPECT_EQ(2, iree_atomic_ref_count_load(&buffer->ref_count));
+
+  IREE_ASSERT_OK(
+      BufferTableStatus(hrx_buffer_table_remove(table, kDevicePointer)));
+  hrx_buffer_release(buffer);
+  EXPECT_EQ(4096u, retained_buffer->size);
+  EXPECT_EQ(1, iree_atomic_ref_count_load(&retained_buffer->ref_count));
+  hrx_buffer_release(retained_buffer);
+  iree_hal_streaming_buffer_table_free(table);
+}
+
 //===----------------------------------------------------------------------===//
 // Multiple operations
 //===----------------------------------------------------------------------===//
--- a/libhrx/src/binding/common/context.c
+++ b/libhrx/src/binding/common/context.c
@@ -147,6 +147,7 @@ iree_status_t iree_hal_streaming_context_create(
   context->stream_wait_frontier = NULL;
   context->event_record_timeline = (iree_hal_streaming_operation_timeline_t){0};
   iree_slim_mutex_initialize(&context->event_record_mutex);
+  context->stream_value_wait_queue = NULL;
 
   // Initialize default limits.
   // These are typical defaults matching CUDA/HIP behavior.
@@ -321,6 +322,11 @@ static void iree_hal_streaming_context_destroy(
 
   // Now release the context's reference to default stream.
   iree_hal_streaming_stream_release(default_stream);
+
+  // The value-wait queue is acquired lazily and can only be released after all
+  // stream work that may reference it has completed.
+  iree_hal_queue_release(context->stream_value_wait_queue);
+  context->stream_value_wait_queue = NULL;
 
   // Free stream tracking resources.
   if (context->streams) {
--- a/libhrx/src/binding/common/context.c
+++ b/libhrx/src/binding/common/context.c
@@ -147,7 +147,6 @@ iree_status_t iree_hal_streaming_context_create(
   context->stream_wait_frontier = NULL;
   context->event_record_timeline = (iree_hal_streaming_operation_timeline_t){0};
   iree_slim_mutex_initialize(&context->event_record_mutex);
-  context->stream_value_wait_queue = NULL;
 
   // Initialize default limits.
   // These are typical defaults matching CUDA/HIP behavior.
@@ -303,16 +302,20 @@ static void iree_hal_streaming_context_destroy(
     iree_hal_streaming_stream_t* stream = context->streams[i];
     iree_hal_queue_t* queue = NULL;
     iree_hal_queue_t* cooperative_queue = NULL;
+    iree_hal_queue_t* value_wait_queue = NULL;
     iree_slim_mutex_lock(&stream->mutex);
     if (stream->context == context) {
       queue = stream->queue;
       cooperative_queue = stream->cooperative_queue;
+      value_wait_queue = stream->value_wait_queue;
       stream->queue = NULL;
       stream->cooperative_queue = NULL;
+      stream->value_wait_queue = NULL;
       stream->context = NULL;
     }
     iree_slim_mutex_unlock(&stream->mutex);
     iree_hal_queue_release(cooperative_queue);
+    iree_hal_queue_release(value_wait_queue);
     iree_hal_queue_release(queue);
   }
   for (iree_host_size_t i = 0; i < detached_stream_count; ++i) {
@@ -322,11 +325,6 @@ static void iree_hal_streaming_context_destroy(
 
   // Now release the context's reference to default stream.
   iree_hal_streaming_stream_release(default_stream);
-
-  // The value-wait queue is acquired lazily and can only be released after all
-  // stream work that may reference it has completed.
-  iree_hal_queue_release(context->stream_value_wait_queue);
-  context->stream_value_wait_queue = NULL;
 
   // Free stream tracking resources.
   if (context->streams) {
--- a/libhrx/src/binding/common/graph.c
+++ b/libhrx/src/binding/common/graph.c
@@ -2665,14 +2665,42 @@ iree_status_t iree_hal_streaming_update_capture_dependencies(
   return iree_ok_status();
 }
 
+iree_status_t iree_hal_streaming_capture_set_last_node_locked(
+    iree_hal_streaming_stream_t* stream,
+    iree_hal_streaming_graph_node_t* node) {
+  IREE_ASSERT_ARGUMENT(stream);
+  IREE_ASSERT_ARGUMENT(node);
+  if (IREE_UNLIKELY(stream->capture_status !=
+                    IREE_HAL_STREAMING_CAPTURE_STATUS_ACTIVE)) {
+    return iree_make_status(IREE_STATUS_FAILED_PRECONDITION,
+                            "stream is not actively capturing");
+  }
+  if (IREE_UNLIKELY(!iree_hal_streaming_graph_node_is_active_in_graph(
+          stream->capture_graph, node))) {
+    return iree_make_status(
+        IREE_STATUS_INVALID_ARGUMENT,
+        "capture frontier node does not belong to the active graph");
+  }
+  if (stream->capture_dependency_capacity == 0) {
+    IREE_RETURN_IF_ERROR(
+        iree_hal_streaming_grow_capture_dependencies(stream, 1));
+  }
+  if (!stream->capture_origin) {
+    stream->capture_joined_to_origin = false;
+  }
+  stream->capture_dependencies[0] = node;
+  stream->capture_dependency_count = 1;
+  return iree_ok_status();
+}
+
 iree_status_t iree_hal_streaming_capture_set_last_node(
     iree_hal_streaming_stream_t* stream,
     iree_hal_streaming_graph_node_t* node) {
   IREE_ASSERT_ARGUMENT(stream);
   IREE_ASSERT_ARGUMENT(node);
-  if (!stream->capture_origin) {
-    stream->capture_joined_to_origin = false;
-  }
-  return iree_hal_streaming_update_capture_dependencies(
-      stream, &node, 1, IREE_HAL_STREAMING_CAPTURE_DEPENDENCIES_SET);
+  iree_slim_mutex_lock(&stream->mutex);
+  iree_status_t status =
+      iree_hal_streaming_capture_set_last_node_locked(stream, node);
+  iree_slim_mutex_unlock(&stream->mutex);
+  return status;
 }
--- a/libhrx/src/binding/common/internal.h
+++ b/libhrx/src/binding/common/internal.h
@@ -292,6 +292,10 @@ struct iree_hal_streaming_context_t {
   // streams. Immutable once published and guarded by |stream_list_mutex|.
   iree_hal_fence_t* stream_wait_frontier;
 
+  // Exact queue used only for memory waits that can block until an independent
+  // producer runs. Lazily acquired and owned by the context.
+  iree_hal_queue_t* stream_value_wait_queue;
+
   // Dedicated mutex for stream list access.
   iree_slim_mutex_t stream_list_mutex;
 
@@ -601,7 +605,7 @@ static inline iree_status_t iree_hal_streaming_stream_reserve_next_value_locked(
     iree_hal_streaming_stream_t* stream, uint64_t* out_wait_value,
     uint64_t* out_signal_value) {
   const uint64_t wait_value = stream->pending_value;
-  if (IREE_UNLIKELY(wait_value == UINT64_MAX)) {
+  if (IREE_UNLIKELY(wait_value >= IREE_HAL_SEMAPHORE_MAX_VALUE)) {
     return iree_make_status(IREE_STATUS_RESOURCE_EXHAUSTED,
                             "stream timeline value overflow");
   }
--- a/libhrx/src/binding/common/internal.h
+++ b/libhrx/src/binding/common/internal.h
@@ -292,10 +292,6 @@ struct iree_hal_streaming_context_t {
   // streams. Immutable once published and guarded by |stream_list_mutex|.
   iree_hal_fence_t* stream_wait_frontier;
 
-  // Exact queue used only for memory waits that can block until an independent
-  // producer runs. Lazily acquired and owned by the context.
-  iree_hal_queue_t* stream_value_wait_queue;
-
   // Dedicated mutex for stream list access.
   iree_slim_mutex_t stream_list_mutex;
 
@@ -553,6 +549,11 @@ typedef struct iree_hal_streaming_stream_t {
   // Lazily acquired cooperative realization of |queue| retaining its exact
   // family, priority, and execution-resource set. NULL until first use.
   iree_hal_queue_t* cooperative_queue;
+
+  // Lazily acquired exact queue dedicated to value operations that may block
+  // on an externally produced memory predicate. Each stream owns a distinct
+  // queue so one unresolved predicate cannot stall another logical stream.
+  iree_hal_queue_t* value_wait_queue;
 
   // Event dependencies that establish safe cross-stream allocation reuse.
   iree_hal_streaming_memory_reuse_dependency_t* memory_reuse_dependencies;
@@ -1043,6 +1044,19 @@ typedef struct iree_hal_streaming_buffer_ref_t {
   iree_hal_streaming_buffer_t* buffer;
   iree_device_size_t offset;
 } iree_hal_streaming_buffer_ref_t;
+
+// Immutable allocation metadata retained independently of the streaming
+// wrapper and buffer-table entry from which it was resolved.
+typedef struct iree_hal_streaming_retained_buffer_ref_t {
+  // HRX allocation retaining the HAL buffer and its physical backing.
+  hrx_buffer_t owner;
+  // HAL buffer borrowed from |owner|.
+  iree_hal_buffer_t* buffer;
+  // Byte offset of the requested pointer into |buffer|.
+  iree_device_size_t offset;
+  // Memory type captured while the allocation is retained.
+  iree_hal_memory_type_t memory_type;
+} iree_hal_streaming_retained_buffer_ref_t;
 
 static inline iree_hal_buffer_ref_t iree_hal_streaming_convert_buffer_ref(
     iree_hal_streaming_buffer_ref_t ref) {
@@ -2056,6 +2070,23 @@ iree_status_t iree_hal_streaming_memory_lookup_range_across_contexts(
     iree_hal_streaming_context_t** out_context,
     iree_hal_streaming_buffer_ref_t* out_ref);
 
+// Looks up and retains immutable allocation metadata for an address range.
+// |out_ref| must be deinitialized by the caller on success.
+iree_status_t iree_hal_streaming_memory_lookup_range_retain(
+    iree_hal_streaming_context_t* context,
+    iree_hal_streaming_deviceptr_t device_ptr, iree_device_size_t size,
+    iree_hal_streaming_retained_buffer_ref_t* out_ref);
+
+// Searches every live context and retains immutable allocation metadata for an
+// address range. |out_ref| must be deinitialized by the caller on success.
+iree_status_t iree_hal_streaming_memory_lookup_range_retain_across_contexts(
+    iree_hal_streaming_deviceptr_t device_ptr, iree_device_size_t size,
+    iree_hal_streaming_retained_buffer_ref_t* out_ref);
+
+// Releases a retained allocation reference and clears its metadata.
+void iree_hal_streaming_retained_buffer_ref_deinitialize(
+    iree_hal_streaming_retained_buffer_ref_t* ref);
+
 // Synchronization: none (allocates memory).
 iree_status_t iree_hal_streaming_memory_allocate_device(
     iree_hal_streaming_context_t* context, iree_device_size_t size,
@@ -2493,6 +2524,11 @@ iree_status_t iree_hal_streaming_update_capture_dependencies(
     iree_hal_streaming_capture_dependencies_mode_t mode);
 
 iree_status_t iree_hal_streaming_capture_set_last_node(
+    iree_hal_streaming_stream_t* stream, iree_hal_streaming_graph_node_t* node);
+
+// Updates the stream capture frontier to |node|. The caller must hold
+// |stream->mutex| and the stream must be actively capturing.
+iree_status_t iree_hal_streaming_capture_set_last_node_locked(
     iree_hal_streaming_stream_t* stream, iree_hal_streaming_graph_node_t* node);
 
 //===----------------------------------------------------------------------===//
--- a/libhrx/src/binding/common/memory.c
+++ b/libhrx/src/binding/common/memory.c
@@ -514,6 +514,74 @@ iree_status_t iree_hal_streaming_memory_lookup_range_across_contexts(
                : iree_status_from_code(IREE_STATUS_NOT_FOUND);
 }
 
+iree_status_t iree_hal_streaming_memory_lookup_range_retain(
+    iree_hal_streaming_context_t* context,
+    iree_hal_streaming_deviceptr_t device_ptr, iree_device_size_t size,
+    iree_hal_streaming_retained_buffer_ref_t* out_ref) {
+  IREE_ASSERT_ARGUMENT(context);
+  IREE_ASSERT_ARGUMENT(out_ref);
+  memset(out_ref, 0, sizeof(*out_ref));
+
+  size_t offset = 0;
+  hrx_buffer_t owner = NULL;
+  IREE_RETURN_IF_ERROR(HRX_CALL(hrx_buffer_table_find_range_retain(
+      &context->buffer_table, device_ptr, (size_t)size, &owner, &offset)));
+  if (IREE_UNLIKELY(!owner || !owner->hal_buffer)) {
+    hrx_buffer_release(owner);
+    return iree_make_status(IREE_STATUS_FAILED_PRECONDITION,
+                            "registered allocation has no HAL buffer");
+  }
+
+  out_ref->owner = owner;
+  out_ref->buffer = owner->hal_buffer;
+  out_ref->offset = (iree_device_size_t)offset;
+  out_ref->memory_type = iree_hal_buffer_memory_type(owner->hal_buffer);
+  return iree_ok_status();
+}
+
+iree_status_t iree_hal_streaming_memory_lookup_range_retain_across_contexts(
+    iree_hal_streaming_deviceptr_t device_ptr, iree_device_size_t size,
+    iree_hal_streaming_retained_buffer_ref_t* out_ref) {
+  IREE_ASSERT_ARGUMENT(out_ref);
+  memset(out_ref, 0, sizeof(*out_ref));
+
+  iree_hal_streaming_device_registry_t* device_registry =
+      iree_hal_streaming_device_registry();
+  if (!device_registry) {
+    return iree_make_status(IREE_STATUS_FAILED_PRECONDITION,
+                            "HAL stream layer not initialized");
+  }
+
+  iree_status_t result = iree_status_from_code(IREE_STATUS_NOT_FOUND);
+  iree_slim_mutex_lock(&device_registry->context_list.mutex);
+  for (iree_hal_streaming_context_t* context =
+           device_registry->context_list.head;
+       context; context = context->context_list_entry.next) {
+    iree_status_t status = iree_hal_streaming_memory_lookup_range_retain(
+        context, device_ptr, size, out_ref);
+    if (iree_status_is_ok(status)) {
+      iree_status_free(result);
+      result = iree_ok_status();
+      break;
+    }
+    if (iree_status_code(status) != IREE_STATUS_NOT_FOUND) {
+      iree_status_free(result);
+      result = status;
+      break;
+    }
+    iree_status_ignore(status);
+  }
+  iree_slim_mutex_unlock(&device_registry->context_list.mutex);
+  return result;
+}
+
+void iree_hal_streaming_retained_buffer_ref_deinitialize(
+    iree_hal_streaming_retained_buffer_ref_t* ref) {
+  if (!ref) return;
+  hrx_buffer_release(ref->owner);
+  memset(ref, 0, sizeof(*ref));
+}
+
 iree_status_t iree_hal_streaming_memory_allocate_device(
     iree_hal_streaming_context_t* context, iree_device_size_t size,
     iree_hal_streaming_memory_flags_t flags,
--- a/libhrx/src/binding/common/stream.c
+++ b/libhrx/src/binding/common/stream.c
@@ -314,6 +314,7 @@ iree_status_t iree_hal_streaming_stream_create(
   stream->queue = queue;
   iree_hal_queue_retain(stream->queue);
   stream->cooperative_queue = NULL;
+  stream->value_wait_queue = NULL;
   stream->memory_reuse_dependencies = NULL;
   stream->memory_reuse_dependency_count = 0;
   stream->memory_reuse_dependency_capacity = 0;
@@ -405,11 +406,14 @@ static void iree_hal_streaming_stream_destroy(
     if (stream->context == context) {
       iree_hal_queue_t* queue = stream->queue;
       iree_hal_queue_t* cooperative_queue = stream->cooperative_queue;
+      iree_hal_queue_t* value_wait_queue = stream->value_wait_queue;
       stream->queue = NULL;
       stream->cooperative_queue = NULL;
+      stream->value_wait_queue = NULL;
       stream->context = NULL;
       iree_slim_mutex_unlock(&stream->mutex);
       iree_hal_queue_release(cooperative_queue);
+      iree_hal_queue_release(value_wait_queue);
       iree_hal_queue_release(queue);
     } else {
       iree_slim_mutex_unlock(&stream->mutex);
--- a/libhrx/src/binding/common/stream_value.c
+++ b/libhrx/src/binding/common/stream_value.c
@@ -1,0 +1,301 @@
+// Copyright 2026 The HRX Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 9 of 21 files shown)


zjgarvey (commented):
1. **[High] Give each logical stream an independently progressing wait lane.**
   The context [caches one exact wait queue for every stream](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L66-L93), and [every memory wait is submitted to it](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L227-L253). HAL explicitly permits an atomic wait to stall its exact queue and requires its producer to progress independently ([queue.h](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/runtime/src/iree/hal/queue.h#L755-L767)). With `X=0` and `Y=1`, enqueue A: `wait(X==1)`, B: `wait(Y==1)`, then B: `write(X=1)`. A stalls the shared wait queue before B's already-satisfied wait; B's gate holds its write behind that wait; the only producer for X can never run. Please make the wait lane per stream, or use a scheduler/pool that guarantees unrelated unresolved waits never share a blocking lane, and add this sequence as a bounded two-stream regression. The capability attribute should remain false where that progress guarantee cannot be provided.

2. **[High] Acquire allocation ownership as part of target lookup.**
   `iree_hip_stream_value_target_t` [retains only the owner context and stores a raw wrapper pointer](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/hip/api.c#L11462-L11516). The buffer table [unlocks before returning that pointer](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/libhrx/buffer_table.c#L250-L277), while `hipFree` can synchronize, remove the entry, and destroy the wrapper before this call submits anything ([memory.c](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/memory.c#L1037-L1074)). Scope selection and enqueue then dereference freed storage; batch pre-resolution widens the interval across as many as 256 targets. Please retain an owning HAL buffer/refcounted wrapper and snapshot immutable metadata while the table lock still protects the entry, then release that ownership only after submission or cleanup. A latch-controlled lookup-versus-`hipFree` test should cover scalar and batch paths under ASAN/TSAN.

3. **[High] Make `hipStreamBatchMemOp` one stream transaction.**
   The batch validates all records first, but then [calls the scalar enqueue helper once per entry](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/hip/api.c#L11783-L11808); each helper [flushes and locks/unlocks the stream independently](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L201-L264). Another host thread can therefore insert a same-stream operation between two entries. For example, a batch `[write X=1, wait Y=1]` can be split by another thread's `write Y=1`, an execution that cannot result from either whole call preceding the other. A later enqueue failure also returns an error after an already accepted prefix, so retrying a batch containing increment/decrement can duplicate effects. ROCm instead [enqueues one batch command](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_stream_ops.cpp#L34-L40), whose contract says entries are enqueued in array order ([command.hpp](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/rocclr/platform/command.hpp#L1009-L1017)). Please represent and submit the array as one ordered batch operation, with all fallible preparation complete before its commit; add same-stream interleaving and failure-injection coverage, and benchmark count scaling instead of expanding a 256-entry call into 256 scalar flush/lock/submission paths.

4. **[High] Recheck capture state under the submission lock.**
   The new helper [reads the plain `capture_status` field without the stream mutex, flushes, and later locks without rechecking](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L161-L208). Capture begin [writes that field under the same mutex](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/graph.c#L2049-L2097). If begin-capture completes during the gap, the atomic is submitted directly after capture is active but is absent from graph replay; the concurrent plain accesses are also a C data race. Please perform the preliminary read under `stream->mutex` and repeat the attachment/capture checks after the flush at the actual locked submission point, following the existing [semaphore-wait pattern](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream.c#L850-L920). Add a latch-controlled capture-transition test under TSAN.

5. **[High] Do not commit an uncancellable wait before the fallible gate.**
   The atomic wait is [accepted and published before the host-call gate is submitted](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L227-L260), but gate submission can still fail while allocating its state or deferred operation ([host call](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/runtime/src/iree/hal/drivers/amdgpu/host_queue_host_call.c#L77-L111), [deferred operation](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/runtime/src/iree/hal/drivers/amdgpu/host_queue_pending_payload.c#L656-L675)). Failing the stream timeline does not cancel an unresolved GPU wait. During destruction, the AMD queue [closes admission, waits for the final hardware epoch, and only then signals the stop condition](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/runtime/src/iree/hal/drivers/amdgpu/host_queue.c#L1165-L1187); that epoch wait [uses an infinite timeout](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/runtime/src/iree/hal/drivers/amdgpu/host_queue.c#L605-L663), so a permanently false predicate can make teardown wait indefinitely. Please prepare a cancelable gate before committing the wait, or provide a transactional HAL primitive with no fallible step after acceptance. Fault-inject every gate allocation/clone site and verify API return, stream/context destruction, and final target release all terminate.

6. **[Medium] Preserve the existing graph-capture contract for default writes.**
   Before this change, a default write to device-only memory copied stable value storage ([base api.c](https://github.com/ROCm/hrx-system/blob/85fbcf374c7ee1f56f8e8d66ccc340bdb9425561/libhrx/src/binding/hip/api.c#L11603-L11617)) into the [graph-aware H2D path](https://github.com/ROCm/hrx-system/blob/85fbcf374c7ee1f56f8e8d66ccc340bdb9425561/libhrx/src/binding/common/memory.c#L2224-L2268). The new common path rejects every active-capture atomic as `UNIMPLEMENTED` ([stream_value.c](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value.c#L161-L169)), which [maps to `hipErrorNotSupported`](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/hip/api.c#L1029-L1055) and leaves capture active. A previously captured write is now absent from the eventual graph. Please either add retained/replayable atomic graph nodes or preserve the old captured default-store path; variants that intentionally remain unsupported should invalidate capture and return `hipErrorStreamCaptureUnsupported`. Add capture/end/instantiate/relaunch coverage for both widths.

7. **[Medium] Exercise the public HIP path in GPU CI.**
   The added tests call only translation helpers and a synthetic capability predicate ([HIP test](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/hip/stream_value_test.cc), [common test](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/stream_value_test.cc)); neither calls the five exported APIs. Their [HIP](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/hip/CMakeLists.txt#L343-L353) and [common](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/libhrx/src/binding/common/CMakeLists.txt#L176-L187) targets also lack the `runtime-resource=amd-gpu` label selected by the [gfx942/gfx1201 jobs](https://github.com/ROCm/hrx-system/blob/c34fbc623546e834d1f3c3381ac6ed1f94bac755/.github/workflows/ci_core_linux.yml#L104-L143). Please add an installed-`amdhip64` GPU test covering both widths/scopes, every predicate/write mode, wait-before-producer progress, two independent streams, batch order/failure, concurrent free/teardown, and capture. Add a deterministic fake-queue test for acquisition, atomic, and post-acceptance gate failures. Those are the paths needed to make the liveness, ownership, ordering, and recovery contracts above executable.

AWoloszyn:
Addressed in e2e02ea3d:
  
1. Each logical stream now lazily owns an independent exact value-wait queue. Capability reporting remains disabled unless the queue family supports dynamic acquisition and every required wait width, scope, and predicate.
  
2. Target lookup now retains the allocation while holding the buffer-table lock and snapshots immutable HAL buffer metadata. Scalar and batch concurrent-free coverage exercises this lifetime contract.

3. Batches are now one transaction: every entry is resolved and recorded before commit, followed by exactly one queue_execute. Added maximum-size, same-stream interleaving, and invalid-batch/no-partial-commit coverage. I did not add a timing benchmark because submission count is now structurally constant; timing thresholds here would be environment-dependent.
  
4. Capture state is read under the stream mutex and checked again under the submission lock after flushing. Concurrent capture-begin and capture-end coverage exercises both outcomes.

5. The fallible post-acceptance host-call gate has been removed. The atomic operation and stream timeline signal are submitted in one HAL operation; queue acquisition and command recording complete before that commit point.
  
6. Default 32-bit and 64-bit stores retain their capture behavior through graph-owned staging and replayable copy nodes. Waits, arithmetic updates, and batches invalidate capture and return the capture-specific error.
  
7. Added an installed-DSO GPU test that calls the public HIP APIs and covers both widths and scopes, every predicate/write mode, independent wait progress, batch ordering and validation, concurrent free, teardown, and graph capture.

zjgarvey (changes requested):
Request changes.

I re-reviewed head `e2e02ea3daf6333025ec3755660fb9695dda7697`. The earlier shared wait lane, borrowed ordinary-device target, scalarized batch, missing final capture recheck, fallible post-acceptance gate, absent default-store capture, and missing installed-DSO target are no longer findings. The remaining issues are:

1. **[High] Retain the physical backing of imported host targets, not only the HAL import.**
   The new retained reference owns an `hrx_buffer_t` and borrows its HAL buffer ([internal.h](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/internal.h#L1048-L1059)). Runtime-owned host storage is separately allocated, imported with a null HAL release callback, and owned only by `wrapper->owns_host_ptr` ([memory.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/memory.c#L1458-L1525)); wrapper destruction releases the HRX/HAL objects and then independently frees that raw pointer ([memory.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/memory.c#L323-L377)). A value call can pause after retained lookup but before submission, while `hipFreeHost`, managed/signal `hipFree`, or `hipHostUnregister` synchronizes only already-submitted work and removes the wrapper. Resuming then submits a live HAL object over freed or repurposed storage. Registered memory uses the same null-callback import, and captured stores later re-resolve the numeric pointer ([capture path](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11624-L11638)), adding an ABA wrong-target path. Please make one refcounted allocation identity own the complete physical backing, make unregister wait for in-flight preparers, and carry the retained identity into capture. Add latch-controlled scalar, batch, and capture races for host, managed, signal, and registered allocations.

2. **[High] Preserve atomic release and scope semantics in captured default writes.**
   Ordinary default writes become HAL atomic stores with `RELEASE`, plus `SYSTEM_SCOPE` for host-visible targets ([translation](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/stream_value.c#L10-L35), [scope selection](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11514-L11535)). Capture instead preserves only the scalar bytes and destination in a `MEMCPY` node ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11601-L11638)), and replay lowers it to `iree_hal_command_buffer_copy_buffer` ([node recording](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/graph_exec.c#L226-L286), [replay](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/graph_exec.c#L1626-L1631)). The copy contract carries neither atomicity nor the release/system-scope parameters, so captured and ordinary execution differ for the publication-flag use case; a 64-bit copy may also be observed non-atomically. Please add an atomic-store graph node that retains width, flags, and the exact target and replays through `iree_hal_command_buffer_atomic_store`. Test the recorded operation/flags and a payload-then-flag handoff without an intervening stream synchronization.

3. **[High] Validate and materialize cross-context targets for the stream's device.**
   When the stream context misses a pointer, target initialization searches every context and accepts the owner's HAL buffer directly ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11486-L11511), [memory.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/memory.c#L542-L575)). It neither identifies an enabled peer owner nor creates an object owned by the executing context. The existing kernel-pointer fallback is not a sufficient guard: `context_has_peer_contexts` returns true when any other context is registered ([context.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/context.c#L847-L864)), although enabled peer owners are recorded separately by `context_enable_peer_access` ([context.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/context.c#L613-L675)); its import predicate also accepts managed or host-local/device-visible storage, not ordinary device peer allocations ([kernel_arguments.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/kernel_arguments.c#L43-L138)). The atomic backend then extracts a raw device address from whichever object it receives ([host_queue_atomic.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/host_queue_atomic.c#L89-L109)). A device-1 stream can therefore return success for an inaccessible device-0 allocation, leading to a fault, stalled wait, or unchanged target. Please add an owner-specific resolver: require the allocation's owner in the executing context's enabled-peer set for device-local memory, materialize and retain a HAL buffer valid in the executing context for each supported host/managed/peer class, apply `SYSTEM_SCOPE` to accepted cross-device atomics as defined by the HAL contract ([atomic.h](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/atomic.h#L55-L63)), and reject every unsupported target before stream mutation. Add two-GPU cases before enablement, after enablement, and after disablement, plus cross-context host/managed cases.

4. **[High] Make legacy default-stream ordering part of the capture transaction.**
   Every new value path performs legacy ordering before its final capture decision. The helper reads plain capture fields without either stream mutex ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L1652-L1670)), while capture begin/end write them under the mutex. Deeper in the path, `stream_wait_streams` takes an early capture snapshot, flushes work, and later accepts a real dependency barrier without rechecking capture ([stream.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream.c#L640-L801)). A capture can begin in that gap, producing a real barrier plus a captured/rejected value operation; capture can also end after the helper skips legacy ordering and before an ordinary value operation is accepted. Besides the C data race, one HIP call can cross the capture boundary and violate legacy ordering. Please serialize legacy dependency setup and capture/ordinary disposition as one operation, using a defined lock order or capture epoch, and add controlled begin- and end-capture interleavings under TSan.

5. **[High] Recycle completed wait lanes instead of pinning a full AMDGPU queue per stream.**
   The first wait lazily acquires a queue and stores its owning reference in `stream->value_wait_queue` ([stream_value.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L44-L86)); the reference is released only when the stream is detached or destroyed ([stream destruction](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream.c#L392-L417), [context detachment](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/context.c#L299-L319)). Dynamic acquisition creates a new AMDGPU host/HSA queue ([dynamic creation](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/logical_device.c#L2281-L2315)), and queue identity has 256 total slots, with provisioned queues consuming part of that namespace ([logical_device.h](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/logical_device.h#L81-L98), [slot allocation](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/logical_device.c#L624-L655)). The defaults also request a 64K-packet AQL ring and a 16 MiB kernarg ring for every such queue ([physical_device.h](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/physical_device.h#L108-L119), [queue initialization](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/host_queue.c#L954-L1033)). Thus 200 long-lived streams that each issue and synchronize an already-satisfied wait retain about 3.1 GiB of kernarg-ring capacity, and later waits fail once the remaining identity slots are exhausted. Please use completion-aware, priority/resource-set-keyed pooling: keep a lane exclusive while its unresolved wait segment is live, then asynchronously recycle it. Give atomic-only lanes appropriately small rings, and add a greater-than-256-live-stream regression that proves bounded queue/memory counts without reintroducing cross-stream blocking.

6. **[Medium] Roll back failed captured-store graph mutations.**
   Captured-store construction attaches staging, inserts a copy node, and then updates the capture frontier ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11624-L11643)). Staging is linked into the graph immediately ([graph.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/graph.c#L160-L183)), while node insertion and initial frontier growth remain fallible. The error path invalidates capture but does not undo completed mutations. With `hipStreamBeginCaptureToGraph`, the caller-owned graph survives invalidation, so a failed API call can retain unreachable staging or leave an inserted write node. Please preflight the frontier or use a graph-mutation checkpoint that removes the node and staging on every later failure; fault-inject each allocation boundary and assert that the supplied graph is unchanged.

7. **[Medium] Return `hipErrorStreamDetached` for a live stream whose execution context was destroyed.**
   Registry lookup followed by failed attachment currently produces `hipErrorContextIsDestroyed` ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L1568-L1583)), and the value-specific wrapper preserves it ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11469-L11477)). ROCm distinguishes an invalid handle from a still-live detached stream and has an exact regression for all five APIs ([reference test](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/hip-tests/catch/unit/executionContext/hipExecutionCtxStreamDetached.cc#L960-L1017)). Please preserve that distinction in stream resolution and port the detached-stream case through the installed DSO.

8. **[Medium] Preserve the public invalidated-capture result at the final rechecks.**
   The HIP precheck returns `hipErrorStreamCaptureUnsupported` when it invalidates an active capture and `hipErrorStreamCaptureInvalidated` when the capture is already invalidated ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11573-L11599)). Both common-layer locked rechecks instead return the same `ABORTED` status for any non-NONE state ([submission recheck](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L200-L218), [preparation recheck](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L290-L314)), and the HIP wrapper maps every `ABORTED` result to Unsupported. If one thread passes the precheck, capture begins, and another operation invalidates it before the first reaches its final recheck, the first returns Unsupported rather than Invalidated. Please preserve the state distinction across the common/HIP boundary with distinct status results or one shared locked helper, and force both final-recheck interleavings in a deterministic test.

9. **[Medium] Make the installed wait suite follow the public capability and allocation contract.**
   The runtime's attribute is derived from the same zero-compute wait requirements used by submission ([query](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L2431-L2441), [requirements](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L13-L42)). The architecture table omits PM4 atomic wait from CDNA and includes it for RDNA ([capabilities](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/runtime/src/iree/hal/drivers/amdgpu/physical_device_capabilities.c#L12-L40)), so gfx942 does not advertise the zero-compute wait lane while gfx1201 does. The DSO fixture resolves neither `hipDeviceGetAttribute` nor `hipExtMallocWithFlags` ([symbol set](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L35-L131), [setup](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L169-L249)); its allocation helper calls ordinary `hipMalloc` ([allocation](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L276-L280)), and wait cases unconditionally expect success ([predicate/system cases](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L341-L417), [independent-stream case](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L532-L569)). The public HIP declaration specifies `hipMallocSignalMemory` storage and directs callers to query `hipDeviceAttributeCanUseStreamWaitValue` ([ROCm header](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/hip/include/hip/hip_runtime_api.h#L3284-L3317)). This is why five of fifteen tests currently fail with 801 on [gfx942 CI](https://github.com/ROCm/hrx-system/actions/runs/34530040913/job/103050682936), while gfx1201 passes. Please allocate 8-byte signal words, assert the public unsupported result when the attribute is zero, and run wait behavior only when it is true; keep write, validation, and capture-store coverage active on both families.

10. **[Medium] Make the new race/transaction tests deterministic.**
    `InvalidBatchDoesNotPartiallyCommit` uses a null second target, which both the scalarized and transactional implementations reject during preflight before entry 1 can commit ([test](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L720-L744)). The capture/free races synchronize only call entry and accept outcomes available without hitting the repaired windows ([capture races](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L643-L718), [free races](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L746-L809)); the batch-interleaving test observes only final values after uncontrolled host scheduling. Required DSO load and `hipInit` failures are reported as skips, so the job can remain green ([setup](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L169-L249)). Please use fake-queue/test-hook latches at retain, capture, record, and commit boundaries; assert exact submission counts and mutation state; and make required DSO loading/initialization fail closed. Keep the stress loops as supplementary coverage.

11. **[Medium] Keep wait-free operations in normal stream batching.**
    `contains_wait` affects only queue selection; every value operation, including a scalar store/update and a write-only batch, flushes the streaming command buffer and submits a separate direct atomic or fresh command buffer ([stream_value.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L137-L198), [submission path](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L280-L350)). A kernel/write/kernel loop therefore forces an extra submission boundary at every write, and every multi-entry write-only batch pays command-buffer create/begin/end/release. The separate lane and pre-flush are necessary for a potentially blocking wait, not for a store or update. Please record wait-free atomics into the stream's existing command buffer under its normal ordering rules, and add submission-count and host-latency benchmarks for scalar writes plus 2/16/256-entry write-only batches interleaved with kernels.

12. **[Medium] Resolve batch targets once per unique allocation.**
    Each batch allocates two heap arrays and independently resolves every entry ([api.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/hip/api.c#L11877-L11929)). Each retained range lookup linearly scans the allocation table while holding its mutex ([buffer_table.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/libhrx/buffer_table.c#L250-L274), [retained lookup](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/libhrx/buffer_table.c#L307-L335)); a miss additionally walks every context and table under the global context-list lock ([cross-context lookup](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/memory.c#L542-L575)). The maximum-size test targets one allocation 256 times, but production repeats all 256 scans/retains/releases. With 10,000 live allocations, that shape can perform about 2.56 million entry probes before recording. Please add a batch lookup that takes each relevant table lock once, retains each unique allocation once, and maps duplicate entries to it; replace or index the cross-context scan, and use inline/combined scratch storage. Benchmark duplicate and distinct targets at batch sizes 1/2/16/256 across 1K and 10K allocations and concurrent submitters.

13. **[Medium] Update the PR description to describe the implementation being reviewed.**
    The current description still says the wait queue is context-owned, a nonblocking host-call gate carries completion back to the stream, and support requires host-call capability. At this head the queue is stream-owned, the atomic submission advances the timeline directly, the host gate is gone, and the capability gate does not require host calls ([stream_value.c](https://github.com/ROCm/hrx-system/blob/e2e02ea3daf6333025ec3755660fb9695dda7697/libhrx/src/binding/common/stream_value.c#L13-L87)). Please rewrite the design and testing sections around the per-stream lane, one-submit batch, retained-target boundary, capture policy, actual capability query, and precise supported/unsupported GPU results.

zjgarvey (changes requested):
Request changes.

Review target: [`f14dce5e`](https://github.com/ROCm/hrx-system/commit/f14dce5e8802e61658820b70cc1e60536f91ecfb). The remaining issues are:

1. **[High] Give every successful write-only call a bounded submission path.**
   The write-only helper flushes older work, creates and begins a retained command buffer, records the stores/updates, assigns it to `stream->command_buffer`, and returns without ending or executing it, reserving a timeline value, or flushing its queue ([recording path](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L330-L365), [branch](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L539-L543)). With the default flush interval disabled ([defaults](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream.c#L150-L167)), there is no autonomous step that submits the last such buffer. For two nonblocking streams, `write(producer, signal=1)` followed by `wait(consumer, signal==1)` can therefore leave the consumer's submitted wait blocked forever if the application only polls `hipStreamQuery(consumer)`; a host poller has the same problem. A later producer call masks this by flushing the prior write, while also making N scalar writes require N command-buffer creations/submissions by the final sync. Please add a bounded eventual-submission policy that retains target backing without forcing one submission per scalar write—for example, a retained-resource list on the normal stream buffer plus a real autonomous flush boundary. Add a bounded `hipStreamQuery(consumer)` polling test that never reuses or synchronizes the producer, a host-visible publication test with no follow-up producer API call, and submission-count benchmarks for consecutive and dispatch-interleaved writes.

2. **[High] Never drain target users or synchronize streams while holding the target-acquisition mutex.**
   Target lookup takes `stream_value_target_mutex` and increments one context-wide active count ([lookup](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/memory.c#L530-L577)). A batch retains distinct targets one at a time and releases them only after the full operation has queued ([batch preparation](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11950-L11995)), while device free holds that same mutex across an infinite wait for the context-wide count and `context_synchronize_all` ([device teardown](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/memory.c#L1285-L1308)); host free and unregister repeat the pattern ([host teardown](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/memory.c#L1899-L1941), [unregister](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/memory.c#L2008-L2044)). If a batch retains A, a concurrent free acquires the mutex, and the batch next resolves B, the free waits for A while the batch waits for the mutex and cannot reach cleanup. Even with no active preparer, synchronizing under this mutex can deadlock a queued host callback: callbacks may submit work to another stream ([callback contract](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L14478-L14490)), but its target lookup needs the mutex held by the synchronization waiting for that callback. Please use a per-allocation closing state and preparation leases: mark only the target unavailable under the lookup/table lock, release all lookup locks, wait for that allocation's leases, synchronize, then reclaim it. Add barrier-controlled two-distinct-target/free and callback-on-unrelated-target tests for device free, host free, and unregister.

3. **[High] Represent captured writes as HIP-visible graph operations.**
   Capture accepts a default scalar write and makes its atomic-store node the stream frontier ([capture path](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11608-L11646)), but that node is marked `HIDDEN` even though it is the call's entire observable effect ([node construction](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/graph.c#L1657-L1704)). `hipGraphGetNodes` omits it, `hipGraphNodeGetType` rejects it, and `hipStreamGetCaptureInfo_v2` can still return its raw handle as a dependency ([enumeration](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L18853-L18904), [type query](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L19365-L19377), [capture info](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L23441-L23471)). Graph-exec update compares only visible nodes and dependencies ([compatibility check](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/graph_exec.c#L2621-L2699), [update](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/graph_exec.c#L2810-L2949)). A store-only graph consequently reports zero nodes although launch writes memory, and updating its executable to an empty graph can be accepted as topology-compatible and silently remove the write. Please expose a supported semantic node—naturally a one-entry `hipGraphNodeTypeBatchMemOp`—with complete enumeration, dependency, clone, parameter, and update handling; reject capture until that contract exists if necessary. Test node count/type, capture-frontier handles, clone, dependency chains, replay, and store-graph-to-empty-graph update rejection.

4. **[Medium] Reuse outstanding same-stream wait lanes and bound the idle cache.**
   Acquisition recycles only completed lanes, so every unresolved wait gets a new dynamic queue even when later waits are already serialized behind it on the same stream timeline; each acquisition also scans every pending lane ([pool](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L57-L148), [submission](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L497-L538)). Idle lanes are kept for reuse and their queues are released during context teardown ([idle retention and teardown](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L151-L192)); reuse requires an exact family/priority/resource-domain match ([match key](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L44-L55)). AMDGPU has a 256-entry queue-identity namespace and returns resource exhaustion when its remaining dynamic slots are live or retired ([limit](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/runtime/src/iree/hal/drivers/amdgpu/logical_device.h#L81-L98), [allocation](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/runtime/src/iree/hal/drivers/amdgpu/logical_device.c#L624-L659)). More unresolved waits than available slots on one stream can therefore fail even though one lane is sufficient, and a completed high-water mark can starve other domains or contexts. Please lease a compatible pending lane to its stream, impose device-aware admission for independently progressing streams, and release surplus idle queues with a small per-domain/global bound. Add deterministic small-limit tests for same-stream outstanding waits, independent streams, peak/drain recovery, domain churn, and a second context; benchmark enqueue growth to catch the current full pending-list scans.

5. **[Medium] Narrow the context-wide capture transition hot path.**
   Every scalar write, scalar wait, and batch takes the same context mutex, and keeps it across legacy ordering plus the complete queue operation ([write/wait](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11680-L11752), [batch](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11977-L11991)). That queue operation can scan/acquire a dynamic lane, allocate and record a command buffer, flush older stream work, submit, and flush the hardware queue ([queue path](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L436-L551)). Calls from different host threads to independent nonblocking streams in one context therefore serialize behind any slow allocation or queue flush. Please use a shared/exclusive transition epoch or per-stream gate so ordinary independent submissions can proceed concurrently while begin/end capture and actual legacy-stream interactions remain exclusive. Add 1/2/4/8-thread enqueue benchmarks and a concurrent capture-transition stress test.

6. **[Medium] Avoid multiplicative target lookup and import work on the hot path.**
   For each distinct batch target, the HIP loop scans all earlier targets and then performs a retained lookup ([batch loop](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11950-L11975)); each range lookup linearly scans the allocation table ([table lookup](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/libhrx/buffer_table.c#L251-L273)). A local miss snapshots and scans every other context, allocating snapshot storage above eight contexts, and a peer hit imports a new HAL view ([cross-context path](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/memory.c#L596-L734)). Thus U unique entries add O(U²) prior-target comparisons plus U allocation/context scans, while repeated scalar calls to one peer allocation repeat the search and import. Please add an address-range index and bulk retained lookup for batches, and cache imported views by execution context/allocation identity for the peer-access lifetime. Benchmark duplicate versus unique targets at 1/2/16/256 entries across allocation/context counts, and count import calls for repeated peer operations.

7. **[Medium] Preserve target-error provenance at the HIP boundary.**
   AMDGPU atomic validation returns `FAILED_PRECONDITION` when the selected width/scope cell is unavailable ([validator](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/runtime/src/iree/hal/drivers/amdgpu/atomic_memory.c#L271-L304)). The stream-value adapter special-cases only capture statuses and otherwise uses the global mapping, where `FAILED_PRECONDITION` becomes `hipErrorNotInitialized` ([adapter](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L11575-L11587), [mapping](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/api.c#L1029-L1056)). An initialized call against an imported or registered target lacking the required atomic cell can therefore masquerade as runtime initialization failure. Please preserve whether the failure came from stream state, device-wide capability, or this target, mapping target incompatibility to the HIP target error and device-wide absence to `hipErrorNotSupported`. Add negative width/scope/import tests that assert the public result.

8. **[Medium] Require the positive GPU runner to exercise waits.**
   The installed test derives `supports_value_waits_` from the implementation's attribute, and `CheckWaitSupport` treats false plus `hipErrorNotSupported` from the same implementation as success before each positive test returns normally ([setup](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L235-L267), [guard](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L310-L315), [example](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/hip/hip_stream_value_api_test.cc#L619-L640)). Independently, the AMDGPU table gives RDNA targets PM4 atomic wait/store support and derives zero-compute atomic operations from those bits ([capability table](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/runtime/src/iree/hal/drivers/amdgpu/physical_device_capabilities.c#L22-L51), [queue capability derivation](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/runtime/src/iree/hal/drivers/amdgpu/device_spec_builder.c#L325-L338)). If gfx1201 capability reporting or submission regresses to unsupported, the installed job stays green without running its wait, lane, capture-race, or teardown behavior. Please pass each runner's independently expected capability into the test, require true on gfx1201, use `GTEST_SKIP()` only where unsupported is expected, and keep system-scope write coverage active even on a wait-unsupported device.

9. **[Medium] Update the PR description to match the current implementation.**
   The body still describes a nonblocking host-call completion gate and host-call capability requirement, but the support predicate requires atomic operations plus dynamic queue acquisition, not host calls ([support predicate](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L13-L42)). Wait queues now come from a pooled lane path ([lane pool](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L85-L171)), and accepted atomic submissions signal the stream timeline directly ([submission](https://github.com/ROCm/hrx-system/blob/f14dce5e8802e61658820b70cc1e60536f91ecfb/libhrx/src/binding/common/stream_value.c#L368-L433)). Please describe the pooled wait-lane ownership, retained-target/free protocol, write submission policy, capture representation, capability split, and exact verification matrix reviewers should evaluate.

Verification for this head: all 12 GitHub checks are successful. The focused host targets `//libhrx/src/binding/common:stream_value_test`, `//libhrx/src/binding/common:buffer_table_test`, `//libhrx/src/binding/common:graph_exec_test`, and `//libhrx/src/binding/hip:stream_value_test` also pass; Bazel-to-CMake registration checks and `git diff --check` pass. Those checks do not force the query-only consumer progress case, the target/free lock cycles, the hidden-node update case, the finite-lane boundary, or multithreaded enqueue scaling above.

zjgarvey (approved):

zjgarvey (approved):
Approved.

Two non-blocking test nits, optional before merge:

- `libhrx/src/binding/common/stream_value_test.cc` uses `__atomic_store_n` and `__ATOMIC_RELEASE` in a target built by MSVC, which does not provide those builtins. IREE's portable atomic helpers would keep the release stores buildable on Windows.
- `FailedWaitLaneDoesNotPoisonOtherStreams` assumes the first owner-stream query reports `hipErrorInvalidValue`, although the following check permits `hipErrorNotReady`. Making the assertion cleanup-safe and synchronizing the deferred error deterministically would avoid leaving the blocked signal unresolved after a failed assertion.

Neither item blocks this HIP-layer change.

PR #563, opened 2026-09-10
[HIP] Implement module, library, linker, and kernel metadata APIs

AWoloszyn:
## Summary

  Implement the HIP module, library, linker, symbol, and kernel metadata APIs on top of the streaming binding’s module representation.

  This change:

  - Loads raw code objects, bundled images, fat binaries, and linked images through the streaming module path.
  - Implements module function and global lookup, including managed globals.
  - Implements HIP library loading, unloading
  - Implements kernel enumeration, naming, parameter metadata, function lookup, library lookup, and attribute access.
  - Implements linker state creation, input addition, completion, and destruction.
  - Exposes the corresponding APIs through dynamic entry-point lookup.
  - Adds the per-thread asynchronous symbol-copy entry point already declared by the export map.
  - Retains module ownership across library and kernel queries so concurrent unloads cannot invalidate active operations.
  - Stores mutable function configuration atomically.

  ## Uniform Workgroups

  Some module launch APIs receive global and local work sizes rather than a grid expressed directly in workgroups. Correctly validating those launches requires knowing whether the kernel permits
  a partial final workgroup.

  This change adds `IREE_HAL_EXECUTABLE_FUNCTION_FLAG_REQUIRES_UNIFORM_WORKGROUPS` as a generic executable-function property. The AMDGPU loader reflects the existing HSACO uniform-workgroup
  metadata into that property. The HIP binding then uses the generic property to distinguish invalid launch dimensions from unsupported partial-workgroup execution.

  The HAL addition is limited to the generic function flag. HSACO parsing remains in the AMDGPU loader, while module, library, linker, and launch behavior remains in the HIP and streaming binding
  layers.


benvanik (approved):

zjgarvey:
### [HIGH] Return the native kernarg offset from `hipKernelGetParamInfo`

`libhrx/src/binding/hip/library.c:572-609` returns each operation's
`source_offset`. The metadata model explicitly distinguishes that dense staging
offset from `native_abi_destination_offset`
(`libhrx/src/binding/common/internal.h:644-669`), and extraction fills them
independently (`libhrx/src/binding/common/module.c:402-504`). The sparse layout
already used by `executable_metadata_hsaco_test.cc:157-179` has native offsets
`0, 8, 20, 32`, while this query reports the dense offsets `0, 2, 10, 16`. A
client packing arguments from this successful query writes pointers and values
to the wrong bytes. ROCm returns the code-object parameter
offset in its [corresponding query](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L296-L315).
Return `native_abi_destination_offset` for both copy and pointer parameters, and
add a production-DSO test that queries and launches a padded scalar/pointer
signature.

### [HIGH] Preserve the lazy, first-query library state

`hipLibraryLoadData` immediately calls `hipModuleLoadDataEx` and defers only
`hipErrorInvalidImage`; it retains no source from which a later build could be
attempted. `hipLibraryLoadFromFile` is fully eager
(`libhrx/src/binding/hip/library.c:345-390`). ROCm instead stores the image or
path and synchronizes a one-time build on the first code-dependent query
([implementation](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L118-L171)).
Consequently, loading on device 0 and switching to device 1 before the first
query cannot select an image available only for device 1 at this head. Model
`unbuilt/building/built/failed` explicitly, retain or own the source according
to the API lifetime contract, and cache the complete first-query result under
the library mutex. Cover malformed input, a valid other-device bundle, and
concurrent first queries.

### [HIGH] Execute permitted non-uniform tail workgroups

For a non-divisible global extent, `libhrx/src/binding/hip/api.c:14236-14244`
returns `hipErrorInvalidValue` when the kernel requires uniform groups and
`hipErrorNotSupported` otherwise. The second case is the one the new metadata
marks as permitted: ROCm rejects only the uniform case and otherwise continues
with the exact global size
([reference path](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_module.cpp#L499-L523)).
A bounds-safe 100-work-item kernel with local size 64 and
`uniform_work_group_size: 0` therefore fails instead of running its 36-item
tail. Carry exact global extents or remainders through the streaming/HAL
dispatch contract and AMDGPU implicit arguments. Test the same non-divisible
launch with metadata values 0 (success) and 1 (rejection).

### [HIGH] Validate both extended-launch events before any launch side effect

The runtime helper submits/captures the kernel before its first validation of
`stopEvent` (`libhrx/src/binding/hip/api.c:13568-13624`); the module helper has
the same ordering (`api.c:13995-14104`). That validation occurs inside the later
public `hipEventRecord` call (`api.c:12068-12100`). A destroyed or
cross-context stop event can thus return an error after a kernel was accepted
or a graph node appended, and retrying the failed call can duplicate work.
Preflight-retain both events and check their contexts before legacy
dependencies, event recording, capture mutation, or dispatch. Record through a
private helper using those retained objects and converge all releases at one
cleanup exit. Add normal-stream and capture tests asserting that invalid stop
events create neither work nor a node.

### [HIGH] Connect static managed registration to the managed allocation

`__hipRegisterManagedVar` records the compiler pointer/value as lookup keys but
does not retain its `void **pointer` as a publication slot
(`libhrx/src/binding/hip/api.c:25507-25535`). Module construction then allocates
new managed storage and writes only its device address into the executable's
pointer slot; the eager visitor discards the returned host pointer
(`libhrx/src/binding/common/module.c:1087-1200`). The compiler's host variable
therefore remains on the old storage while GPU code dereferences the new one.
Static `hipGetSymbolAddress`/`hipGetSymbolSize` also return the pointer slot and
its pointer-sized extent (`api.c:9881-9970`), so `hipMemcpyToSymbol` can
overwrite the device pointer rather than the payload. ROCm retains the compiler
slot and publishes the allocation through it
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_platform.cpp#L241-L280)).
Make the registration own that publication edge, map static symbol APIs to the
managed buffer/logical size, and test host/GPU coherence plus symbol copies.

### [HIGH] Reject a library kernel on a foreign device before enqueue

`hipLibraryGetKernel` returns a tagged symbol from the library's loaded module
(`libhrx/src/binding/hip/library.c:408-420`). The common resolver checks only
the tag and symbol type; launch then validates dimensions against the stream's
device and submits the symbol's executable without comparing module and stream
contexts (`libhrx/src/binding/hip/api.c:13491-13619`). AMDGPU kernel objects are
agent-specific, so a kernel loaded on device 0 can reach a valid device-1
stream instead of returning `hipErrorInvalidDevice`. ROCm performs this device
check before launch
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_module.cpp#L295-L340)).
Centralize symbol/stream affinity validation before capture or enqueue and add
a two-device test for explicit and default streams.

### [MEDIUM] Release the resolved stream on every module-launch error

After `iree_hip_resolve_registered_stream` retains the explicit wrapper, common
stream, and context, the missing-parameter return at
`libhrx/src/binding/hip/api.c:14026-14030` and failed start-event return at
`api.c:14079-14084` omit `iree_hip_resolved_stream_release`. Repeated invalid
calls therefore retain stream/context resources indefinitely. Route every
post-resolution exit through one cleanup block and add lifetime-sensitive tests
for missing arguments and failed start-event recording on explicit and default
streams.

### [MEDIUM] Break the retired-module/context ownership cycle

The context symbol map owns each lazily created module
(`libhrx/src/binding/common/registry.c:767-770`), and that module retains the
same context (`libhrx/src/binding/common/module.c:624-625`). Unregister now
clears only `module_entry->registration` while leaving the module linked
(`registry.c:844-848`); the map releases it only from the context destructor
(`registry.c:535-542`), which the retained context can never reach. A normal
register/resolve/unregister cycle therefore permanently retains the context and
compiled resources without any stale-handle use. Give queued/graph/function
consumers their own explicit owner, then unlink and release the map entry after
those users drain. Add a repeated DSO registration/unregistration teardown
test.

### [MEDIUM] Handle library and JIT options consistently

`iree_hip_library_validate_options` rejects every nonempty library-option array
(`libhrx/src/binding/hip/library.c:252-257`), including
`hipLibraryHostUniversalFunctionAndDataTable` and
`hipLibraryBinaryIsPreserved`. Meanwhile the memory form forwards recognized
JIT options to `hipModuleLoadDataEx`, which accepts them as ignored hints, while
the file form rejects every nonempty JIT array (`library.c:345-390`). ROCm
accepts the defined library-option arrays and rejects nonempty JIT options
consistently in both forms
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L140-L171)).
Use one typed option parser, accept/apply the defined library options, reject
malformed or unknown values with the documented error, and cover both load
forms.

### [MEDIUM] Do not report unknown constant-memory use as zero

`hipKernelGetAttribute` unconditionally returns success and zero for
`hipFuncAttributeConstSizeBytes`
(`libhrx/src/binding/hip/library.c:638-640`), while the common attribute record
contains no constant-memory field. A kernel using constant storage is therefore
reported as using none. ROCm returns the actual value and has a public
128-byte-constant test
([implementation](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L318-L377)).
Plumb constant-memory metadata end to end, or return an explicit unsupported
error until it is available; add the 128-byte library-kernel case.

### [MEDIUM] Keep the HIP binding buildable with the AMDGPU driver disabled

The Linux `amdhip64` CMake target unconditionally links
`iree::hal::drivers::amdgpu`
(`libhrx/src/binding/hip/CMakeLists.txt:298`), but that target exists only when
`IREE_HAL_DRIVER_AMDGPU=ON`. The following configure-only matrix succeeds at
the merge base and fails at this head with the missing namespaced target:

```sh
cmake -S <source> -B <build> -G Ninja \
  -DLIBHRX_BUILD=ON \
  -DLIBHRX_BUILD_HIP_BINDING=ON \
  -DLIBHRX_BUILD_CTS=OFF \
  -DLIBHRX_BUILD_PASSTHROUGH=OFF \
  -DLOOM_BUILD=OFF \
  -DIREE_BUILD_TESTS=OFF \
  -DIREE_HAL_DRIVER_AMDGPU=OFF
```

No HIP source directly includes the AMDGPU driver or calls its symbols. Remove
the unused direct dependency and regenerate CMake; if it is intentional, gate
the whole binding with an explicit option error. Add this configure-only
combination to CI.

### [MEDIUM] Avoid a full-size anonymous copy for file module loads

`libhrx/src/binding/common/module.c:719-749` replaces the base's file-backed
`iree_io_file_handle_open` mapping with `iree_io_file_handle_preload`, which
allocates and reads the entire image before mapping that allocation. HAL's
executable-load contract requires implementations to finish consuming or copy
borrowed input before return, so the preload does not provide a required
lifetime extension. A large HSACO or multi-image fat binary now has an
additional O(file-size) allocation and failure mode. Restore file-backed
mapping, or fix any backend that violates the HAL borrowing contract. Add
cold/warm `hipModuleLoad` measurements for wall time and peak RSS; no magnitude
is claimed here.

### [MEDIUM] Extend shipped-DSO coverage to the remaining new behavior

Existing CTS coverage opens the candidate DSO and exercises
`hipModuleLoadData` -> `hipModuleGetFunction`/`hipFuncGetAttribute` ->
`hipModuleLaunchKernel` -> `hipModuleUnload` with embedded code objects;
dynamic-lookup tests also resolve and invoke `hipGetDriverEntryPoint` and
`hipGetDriverEntryPoint_spt`. Repository search finds no test call to the new
`hipLibrary*` or `hipLink*` implementations, `hipKernelGetParamInfo`,
`__hipRegisterManagedVar`, or `hipExtModuleLaunchKernel`. It also finds no
production-DSO test for `hipModuleLoadDataEx`, `hipModuleLoadFatBinary`,
`hipModuleGetGlobal`, `hipModuleGetTexRef`, `hipModuleGetFunctionCount`, or
`hipGetFuncBySymbol`. The parser/metadata tests stop before padded parameter
packing, managed host/device coherence, and HIP-visible non-uniform launch
behavior, so those uncovered paths can return unsupported or incorrect results
without failing current tests. Extend the existing production-DSO CTS with
library enumeration/metadata/load/unload, linker
create/add/complete/load/execute/destroy, the uncovered module variants, padded
parameter query-and-launch, managed coherence, non-uniform extended launch,
invalid-event failure atomicity, repeated-registration teardown, and optional
two-device affinity. Include targeted mutations that prove each test reaches
the intended path.

AWoloszyn:
Addressed the review in 980080fee:
  1. hipKernelGetParamInfo now returns native ABI destination offsets, with padded scalar/pointer query-and-launch coverage.
  2. Libraries now retain their source and use synchronized UNBUILT, BUILDING, BUILT, and FAILED states. The first query selects the current device image and caches either success or failure.
  3. Non-uniform tail workgroups now carry exact work-item extents through streaming, HAL, replay, and AMDGPU AQL emission. Profiled and ordinary submission paths produce equivalent packets.
  4. Extended launches retain and validate both events before capture or submission, with unified cleanup on every exit.
  5. Static managed registrations now publish shared host/device-visible backing through the compiler slot. Registrations and retained modules share refcounted ownership, including graphs surviving registration teardown.
  6. Library kernels are checked against explicit and default stream devices before capture or enqueue.
  7. Every post-resolution module-launch path now releases its resolved stream and context ownership.
  8. Unregistration removes context-map module entries while functions, queued work, and graphs retain explicit ownership.
  9. Memory and file library loads share typed option validation and consistent errors.
  10. Unknown constant-memory usage now returns hipErrorNotSupported instead of incorrectly reporting zero.
  11. The unused direct AMDGPU binding dependency was removed, and the HIP DSO builds with IREE_HAL_DRIVER_AMDGPU=OFF.
  12. File module loading now uses file-backed mappings rather than preloading a full anonymous copy.
  13. Added production-DSO coverage for module/library metadata, lazy loading, linker ownership and execution, exact tail launches, invalid-event atomicity, managed globals, repeated teardown, and optional multi-device affinity.

zjgarvey:
Re-review of `881e21c0f`, ordered by severity.

The original thirteen mechanisms have changed at this head: native parameter
offsets, lazy source/result caching, direct/replay/AQL exact extents,
stop-event preflight, managed publication, ordinary function/stream affinity,
launch cleanup, the module/context cycle, option parsing, scalar constant-memory
reporting, AMDGPU-off build wiring, file-backed module mapping, and production
DSO coverage are all present in their revised forms. The following
current-head issues remain.

### [HIGH] Acquire a live owner before dereferencing tagged function handles

`iree_hip_resolve_function_symbol` treats the tag bits as proof that an
arbitrary value is a live `iree_hal_streaming_symbol_t *`, untags it, and
reads `symbol->module` before acquiring any registry-protected owner
(`libhrx/src/binding/hip/api.c:13535-13570`). Module unload can free the
symbol array concurrently, and a stale or fabricated tag-shaped value follows
the same invalid read. `hipModuleLaunchCooperativeKernel` also untags and
reads the symbol without retaining its module at all
(`api.c:14495-14534`). The new library kernel/function handles use this same
representation (`library.c:540-555,659-669`).

Resolve exact live membership and retain the owner under the same lock/state
that retires it, before reading any symbol field. A stable, generation-bearing
handle table would also avoid allocator-address ABA; keep lookup expected O(1)
so the repair does not turn launch into a global linear scan. Cover stale,
fabricated, allocator-reused, and launch/query-versus-unload handles from both
module and library APIs under ASAN/TSAN.

### [HIGH] Synchronize lazy library module publication with every reader

The first query writes the plain `library->module` field while holding only
`build_mutex` (`libhrx/src/binding/hip/library.c:258-302`). A kernel
metadata query scans every live library and reads that field while holding only
the process registry mutex (`library.c:314-345`); unload also reads it
without `build_mutex` (`library.c:525-535`). The atomic wrapper refcount
keeps the allocation alive but does not create a happens-before edge for this
field.

Two independent valid handles are sufficient to trigger the race: one thread
first-builds library A while another queries a kernel already obtained from
library B, causing B's owner scan to read A's concurrently published field.
Publish the completed module with release/acquire semantics or guard all state
and module accesses with one documented lock order; make unload wait for the
terminal build state. Add a barrier-controlled two-library TSAN case plus a
first-query/unload case.

### [HIGH] Make executable-graph kernel updates ownership-aware

`hipGraphExecKernelNodeSetParams` snapshots the complete kernel attributes by
shallow copy, calls the ordinary node setter, rebuilds the executable, and then
assigns the snapshot back (`libhrx/src/binding/hip/api.c:21656-21728`).
The revised common setter now retains the replacement module and releases the
previous module (`libhrx/src/binding/common/graph.c:1398-1415`). The
shallow restore neither reacquires the old ownership nor releases the temporary
new ownership. After a cross-module A-to-B update, the source graph can contain
freed A symbol/module pointers while retaining B indefinitely.

Use an ownership-aware snapshot/swap transaction for the symbol, module,
constants, and bindings. Add a same-context cross-module executable update
that unloads A's public handle before the update and then exercises launch,
query, executable destruction, and source-graph destruction under a lifetime
checker.

### [HIGH] Preserve captured exact extents across graph parameter updates

Extended-launch capture stores the exact 100-work-item extent for a 100/64
launch, and graph execution uses it to suppress the 28 padded work-items.
`hipKernelNodeParams` has no field for that hidden extent, but
`hipGraphKernelNodeSetParams` builds a zero-initialized streaming parameter
record and the common setter unconditionally copies its zero
`workitem_count` into the node
(`libhrx/src/binding/hip/api.c:21889-21966`;
`libhrx/src/binding/common/graph.c:1398-1405`). The executable setter
rebuilds from that temporarily cleared node. A no-op public update can
therefore change the next execution from 100 to 128 work-items and write past a
100-element logical range.

Keep the captured exact/remainder metadata when the public parameter type
cannot replace it; if changed geometry cannot preserve it safely, reject the
update. Test both source-graph and executable-graph no-op updates of a captured
100/64 launch and assert that elements 100-127 remain untouched. Pinned ROCm
keeps these remainders in separate graph-node fields
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_graph_internal.hpp#L1324-L1333)).

### [HIGH] Clamp extended-launch local dimensions before tail validation

`hipExtModuleLaunchKernel` detects a partial group and computes the grid from
the caller's local dimensions without first clamping a local dimension that is
larger than its global dimension (`libhrx/src/binding/hip/api.c:14290-14328`).
The new CTS consequently expects global X 17/local X 64 to expose local size
64 (`module_library_execution_test.cpp:763-778`). Pinned ROCm validates the
requested local shape, then clamps each effective local dimension to the
corresponding global dimension before uniformity and command construction
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_module.cpp#L493-L523)).

With the current code, a non-uniform kernel observes the wrong local size and a
uniform kernel is rejected even though clamping makes the launch divisible.
Derive `effective_local[i] = min(requested_local[i], global[i])` after
device/kernel-limit validation, then use it for partial detection, grid
calculation, uniform validation, and dispatch. Change the 17/64 assertion to
local 17 and add the uniform-kernel acceptance case.

### [HIGH] Complete device-affinity validation for cooperative, graph, and old-stream paths

Ordinary launch now compares the function's module context with the resolved
stream context, but the cooperative helpers and graph add/set paths do not
(`libhrx/src/binding/hip/api.c:14343-14534,17268-17366,21889-21966`).
They can combine a device-0 executable with a device-1 cooperative queue or
graph. A second gap remains when a device-0 function and explicit device-0
stream are retained, the current device changes to 1, and module launch is
called: the explicit-stream resolver uses the stream's attached context but
does not consult the current device (`api.c:1625-1664`). Function and stream
therefore match each other, so HRX silently enqueues on the non-current device.
Pinned ROCm checks both the selected function/device and the explicit stream
against the current device before enqueue
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_module.cpp#L295-L340),
[stream check](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_module.cpp#L477-L484)).

Centralize a retaining resolver that enforces module-context/target-context
identity before graph mutation, occupancy, or enqueue, and independently
enforce the current-device rule for ordinary module/library launches. Extend
the two-device DSO test to all cooperative and graph routes and to an old
explicit stream after a device switch; failed calls must leave graphs and
stream timelines unchanged.

### [HIGH] Reject exact dispatch on backends that do not implement it

The new generic `IREE_HAL_DISPATCH_FLAG_EXACT_WORKITEM_COUNT` contract says
that `workitem_count` replaces the padded
`workgroup_count * workgroup_size` extent
(`runtime/src/iree/hal/queue.h:356-361`). The task backend records only group
size/count
(`runtime/src/iree/hal/drivers/task/command/block_command_ops.c:262-397`)
and WebGPU likewise discards the flag/exact extent
(`webgpu_command_buffer.c:327-348`, `webgpu_queue.c:2037-2085`). Both can
return success for exact 100 over two groups of 64 while running 128 items.

Implement exact final-workgroup behavior in those backends or reject the flag
before recording/submission. Add backend-matrix coverage requiring either the
exact result or an explicit unsupported error; successful padded execution
must not be an allowed outcome.

### [HIGH] Do not return successful fabricated constant-memory metadata

`hipFuncGetAttributes` zeroes the result, assigns
`constSizeBytes = 0 /* Not tracked */`, and returns success
(`libhrx/src/binding/hip/api.c:13228-13269`). The scalar
`hipFuncGetAttribute` and `hipKernelGetAttribute` paths now correctly
return `hipErrorNotSupported` for the same field. A kernel with nonzero
constant storage therefore receives two contradictory answers, and the
aggregate answer is a plausible silent lie. Pinned ROCm reports the actual
value and tests a 128-byte constant
([implementation](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_global.cpp#L108-L133),
[test](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/hip-tests/catch/unit/executionControl/hipFuncGetAttributesBasic.cc#L11-L28)).

Plumb constant-memory metadata to both APIs. Until it exists, return
`hipErrorNotSupported` before modifying the aggregate output. Cover compiler,
module, and library-derived functions using nonzero constant storage.

### [HIGH] Build the missing-symbol status before releasing registration ownership

On the slow registered-symbol path, a selected module may load successfully
while the requested stub is absent. The terminal branch clears the output,
unlocks the context map and global registry, and only then reads
`registration->device_name` while formatting the error
(`libhrx/src/binding/common/registry.c:1095-1101`). Concurrent
`__hipUnregisterFatBinary` can acquire the released registry lock and free
that registration first, making the diagnostic itself a use-after-free.

Copy the name or construct an owned status while the registry lock still
protects it, then unlock. Add a controlled absent-symbol/unregister race under
ASAN/TSAN.

### [MEDIUM] Release graph-node ownership when argument packing fails

Graph kernel-node construction retains `symbol->module` before packing the
argument list, then returns directly if packing fails
(`libhrx/src/binding/common/graph.c:1147-1152,1176-1228`). A null parameter
element is one public trigger. Because the node has not yet been linked into
the graph, graph destruction never deinitializes its attributes and the new
module retain leaks. Retain only after failure-prone packing succeeds, or
deinitialize every unlinked node on failure; add a repeated invalid-parameter
case with leak detection.

### [MEDIUM] Measure the complete raw ELF before taking lazy ownership

Default `hipLibraryLoadData` clones size-less raw input for the later first
query (`libhrx/src/binding/hip/library.c:445-470`). The ELF estimator defines
the full copy as only `e_shoff + e_shentsize * e_shnum`, and the clone copies
that prefix (`libhrx/src/binding/common/fat_binary.c:308-317,348-369,588-608`).
The section table need not be last; valid file-backed sections or program
segments may follow it, and stripped ELF can have no section table. Such a
library is accepted initially and later builds from a truncated owned image.

Compute the checked maximum extent of the ELF header, program-header table,
every file-backed program segment, section-header table, and every non-NOBITS
section, including extended-count rules. Add raw-HSACO clone and production-DSO
tests with required data after the section table and with no section table.

### [MEDIUM] Validate a library file's existence at load time

`hipLibraryLoadFromFile` copies/registers any nonempty path and returns
success; only the first query attempts `hipModuleLoad`
(`libhrx/src/binding/hip/library.c:506-522,258-307`). A nonexistent file
therefore produces a live-looking handle and moves the failure to an unrelated
query. Pinned ROCm keeps code-object parsing lazy but validates existence in
the load call
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L160-L171)).

Check existence/access before registry insertion, return the established load
error without a handle, and keep only parsing/architecture selection lazy. Add
a unique nonexistent-path DSO case.

### [MEDIUM] Finish synchronous launch validation before re-recording the start event

Module launch records `startEvent` before entering common argument
preparation (`libhrx/src/binding/hip/api.c:14153-14190`). A short pre-packed
kernarg buffer is rejected later in
`libhrx/src/binding/common/stream.c:1643-1653`, with the size check in
`libhrx/src/binding/common/kernel_arguments.c:12-24`, after the event has
already been rebound to a newly submitted timeline point. Move every
synchronous, fallible argument validation/packing step before legacy ordering
or either event mutation. Test that a previously recorded event and the stream
timeline remain unchanged after a short-buffer rejection.

### [MEDIUM] Preserve HIP-facing parameters for stream-captured kernel nodes

Stream capture calls the common graph constructor directly
(`libhrx/src/binding/common/stream.c:1554-1563`). The zero-initialized node stores
the common symbol and copied bytes but not `hip_function`,
`hip_kernel_params`, or `hip_extra`; the explicit graph-add and graph-set
paths fill those fields (`libhrx/src/binding/hip/api.c:17364-17366,21963-21965`),
but common stream capture does not. `hipGraphKernelNodeGetParams`
subsequently returns success with null function/argument metadata
(`libhrx/src/binding/hip/api.c:21864-21885`), so its output cannot support a
get/edit/set round trip.

Store a graph-owned HIP-facing representation during capture, not pointers to
the caller's stack arrays. Cover compiler, module, and library kernel capture
followed by get/edit/set after the original argument array is gone.

### [MEDIUM] Fail the configured production-DSO test when its DSO cannot load or initialize

The new fixture receives the exact candidate DSO from both build systems, but
turns `dlopen` failure or any `hipInit` failure into `GTEST_SKIP`
(`libhrx/cts/tests/hip/module_library_execution_test.cpp:350-359`;
`libhrx/cts/BUILD.bazel:265-271`;
`libhrx/cts/CMakeLists.txt:417-430`). A missing dependency, bad version
script, unresolved symbol, or initialization regression can therefore skip
every body while the configured test is green. Make these hard assertions when
the candidate path is injected; reserve skipping for an explicit
unconfigured/local discovery mode.

### [MEDIUM] Require one compiler-enabled linker success lane

The only linker ownership/execution case now destroys its state and skips the
entire test when `hipLinkComplete` returns `hipErrorNotSupported`
(`module_library_execution_test.cpp:984-1061`). That also skips the later
file-input ownership half. Optional COMGR absence is a valid production result,
but it cannot verify the successful compiler boundary, copied inputs, output
lifetime, module load, or launch.

Split dependency-neutral create/add/destroy ownership checks from the optional
production compiler integration, split data and file cases, and require a
COMGR-provisioned lane to complete link/load/launch. Keep a separate
non-COMGR case that requires the production `hipErrorNotSupported` result.

### [MEDIUM] Run ownership claims under sanitizers and the required device topology

The entire DSO test binary sets `ASAN_OPTIONS=detect_leaks=0`
(`libhrx/cts/BUILD.bazel:266-273`; `CMakeLists.txt:417-436`), masking HRX
leaks outside COMGR. Passing executions of this binary therefore cannot
substantiate its module- and library-ownership claims with leak detection. The
new library threads join before unload, the managed kernel synchronizes before
unregister, the captured graph is never instantiated and launched after
unregister, and the affinity case may skip without two devices
(`module_library_execution_test.cpp:480-570,850-855,937-979`).

Put COMGR residency in a separate process/test or use a narrow suppression,
then run the remaining DSO lifecycle suite with leak detection plus ASAN/TSAN.
Add barrier-controlled query/unload overlap, unregister-before-completion,
instantiate/launch-after-unregister, and a required two-device lane.

### Lower-priority follow-ups

- Every warm compiler-stub launch now takes one exclusive context-wide
  symbol-map mutex (`libhrx/src/binding/common/registry.c:1017-1026`), and a
  cold miss holds the process registry lock across executable loading
  (`:1037-1069`). Preserve lifetime safety with a read-mostly O(1) scheme and
  independently retained registrations; benchmark 1/2/8/32 submitting threads.
  No slowdown magnitude is claimed here.
- The comment above `hipModuleLoad` still advertises NVIDIA PTX/SASS and
  generic architecture binaries even though the revised loader accepts the
  AMDGPU ELF/bundle/CCOB/HIP-wrapper set
  (`libhrx/src/binding/hip/api.c:12537-12540`). State the implemented formats.
- Module handles, loading, metadata, launch/event sequencing, occupancy, graph
  updates, dynamic lookup, and compiler registration remain coupled in the
  25k-line `api.c`. Split these invariant clusters behind the ABI wrappers so
  variants share one implementation/verification boundary.

### Verified clean

- Kernel-parameter queries now return native ABI destination offsets, including
  for padded arguments, and the DSO test queries and launches through those
  offsets (`libhrx/src/binding/hip/library.c:720-736`;
  `module_library_execution_test.cpp:521-567`).
- Managed compiler registration now retains its shared backing through
  publication, and module retirement removes context-map ownership before
  release (`libhrx/src/binding/common/registry.c:286-306,398-499,826-993`;
  `libhrx/src/binding/common/module.c:1322-1354`).
- Ordinary module launch now acquires both event handles before submission,
  checks function/stream context identity, and releases retained objects through
  unified cleanup (`libhrx/src/binding/hip/api.c:13586-13604,13655-13660,14067-14216`).
  The distinct start-event ordering and sibling-route affinity gaps are listed
  above.
- Implemented library/JIT options share typed validation, and file-backed
  module loading restores a read-only mapping instead of retaining a transient
  preload buffer (`libhrx/src/binding/hip/library.c:356-407,485-523`;
  `libhrx/src/binding/common/module.c:730-772`).

### Summary

9 HIGH, 8 MEDIUM, and 3 lower-priority follow-ups.

### Verification

- `git diff --check` and Buildifier checks pass at `881e21c0f`.
- A clean, AMDGPU-disabled CMake configuration of the HIP binding passed, and
  focused compilation of `fat_binary.c`, `graph.c`, `api.c`,
  `library.c`, and `linker.c` passed without using ROCm. No local GPU
  runtime or performance test was run.
- Current-head CPU ASAN/TSAN/MSAN/UBSAN, gfx1150, and
  [gfx1201 CMake GPU](https://github.com/ROCm/hrx-system/actions/runs/34607531094/job/103291493887)
  checks passed. The sole source-run red lane,
  [gfx942 CMake GPU](https://github.com/ROCm/hrx-system/actions/runs/34607531094/job/103291493907),
  stopped in its pre-build `rocminfo` hardware probe after a 30-second hang
  and SIGKILL; it did not configure, build, or test this change. The CMake
  summary is red because of that matrix result.

zjgarvey:
Re-review of `087e5d73c` (including the three commits since `881e21c0f`),
ordered by severity.

The following are distinct current-head issues introduced or worsened by this
PR.

### [HIGH] Accept `hipGetFuncBySymbol` results in every module-occupancy API

`hipGetFuncBySymbol` now returns the registered compiler host address as a
`hipFunction_t` (`libhrx/src/binding/hip/api.c:13060-13088`). The general
function resolver understands that representation, but the resolver used by
all four `hipModuleOccupancy*` entry points consults only the opaque function
table and returns `hipErrorInvalidHandle` on a miss
(`api.c:14747-14770,14858-14923`). Thus a successful result from the former is
deterministically rejected by the latter before occupancy is queried.

Pinned ROCm coverage composes these APIs in exactly this way for all four
occupancy variants and does not permit `hipErrorInvalidHandle`
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/hip-tests/catch/unit/hrr/hrr_workload.cc#L1970-L2022)).
Use one retaining resolver that accepts both compiler-symbol and opaque module
handle forms while preserving each API's error mapping. Extend the production
DSO test from merely launching the `hipGetFuncBySymbol` result to passing it
through all four module-occupancy calls.

### [HIGH] Attach captured function metadata to the node created by that launch

The common capture path adds a kernel node and publishes it as the stream's
last dependency before returning (`libhrx/src/binding/common/stream.c:1544-1565`).
The HIP wrapper then enters a second stream critical section and writes the
public function into whichever kernel is currently the last dependency
(`libhrx/src/binding/hip/api.c:13627-13640,13731-13735`). The launch API is
documented as thread-safe, but the two locked regions do not form one
transaction.

During relaxed capture, thread A can add node A and pause before the
annotation; thread B can add and annotate node B; then A overwrites B's
`hip_function`. `hipGraphKernelNodeGetParams` subsequently reports no function
for A and function A for B, so a get/edit/set round trip can silently retarget
B. Pinned ROCm coverage explicitly ends a relaxed capture from a different
host thread
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/hip-tests/catch/unit/graph/hipStreamEndCapture_old.cc#L183-L236)).
Pass the public function identity into the common capture operation, or
return the exact created node and annotate it before publishing the new stream
tail. Add a barrier-controlled two-thread capture test which verifies both
queried functions and executions.

### [HIGH] Preserve cooperative mode across graph kernel parameter updates

`hipKernelNodeParams` cannot represent whether a node is cooperative. The
public setter creates internal flags only for argument encoding
(`libhrx/src/binding/hip/api.c:22058-22088`), yet the common setter overwrites
the hidden `attrs->cooperative` bit from those flags
(`libhrx/src/binding/common/graph.c:1415-1416`). Graph compilation/submission
uses the bit to select cooperative dispatch
(`graph_exec.c:1922-1924,2145-2152`). Consequently even a get-then-set update
of an unchanged cooperative node succeeds and rebuilds/runs it as an ordinary
dispatch; a grid-wide barrier can hang or produce wrong output.

Pinned ROCm keeps cooperative mode separate from public kernel parameters and
does not change it in `SetParams`
([reference](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_graph_internal.hpp#L1316-L1337),
[setter](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_graph_internal.hpp#L1651-L1677)).
Preserve this node property independently of argument-format flags in both
source and executable setters. Extend the cooperative capture test to apply
unchanged parameters through each setter and verify cooperative replay.

### [HIGH] Revalidate a replacement function against a preserved exact extent

For a captured partial launch, the graph setter correctly freezes grid/block
geometry and preserves its hidden exact `workitem_count`, but it allows
`params->func` to change and performs only generic launch validation
(`libhrx/src/binding/hip/api.c:22012-22054,22075-22092`). It omits the direct
extended-launch rejection for a function marked
`REQUIRES_UNIFORM_WORKGROUPS` (`api.c:14210-14214`). Graph compilation then
emits the exact partial extent for that replacement function
(`libhrx/src/binding/common/graph_exec.c:1928-1937`).

Capture 100 work items in groups of 64 for a non-uniform-capable kernel, then
replace only `func` with a layout-compatible uniform-only kernel: either graph
setter accepts a launch which the direct path rejects. This violates the
replacement kernel's compilation assumption and can fault or silently compute
the wrong result. Validate the candidate's uniform-workgroup flag against the
preserved exact count before mutation. Test rejection and rollback in both
source and executable setters, plus a successful non-uniform replacement.

### [HIGH] Make graph insertion atomic through root-block allocation

`iree_hal_streaming_graph_add_node` appends a node and increments `node_count`
before allocating a root block for a dependency-free node
(`libhrx/src/binding/common/graph.c:550-580`). If that second allocation fails,
the new kernel cleanup treats the node as unlinked, releases its module, and
clears its module and symbol (`graph.c:1229-1237`). The node is still published
in `node_blocks`, and instantiation later dereferences
`attrs->symbol->executable` (`graph_exec.c:1932-1937`). The add therefore
returns an error without a node handle but leaves a later deterministic null
dereference.

Allocate every required node/root block before assigning identity or linking
the node, or completely roll back the first publication before clearing its
owners. Add allocator-fault coverage at the root-block allocation and require
unchanged node/root counts plus successful query, instantiate, and destroy of
the pre-existing graph.

### [MEDIUM] Use normalized ELF section counts in managed-global visitation

The new ELF measurement code supports extended numbering by reading the real
section count from section zero when `e_shnum == 0`
(`libhrx/src/binding/common/fat_binary.c:526-553`). The new global-object
visitor instead re-reads the raw header and loops a `uint16_t` only up to raw
`header.shnum`, also checking `sh_link` against that raw value
(`fat_binary.c:798-833`). A valid extended-numbering AMDGPU ELF is therefore
accepted by the visitor but has zero globals visited. Module/library load can
report success without initializing a `*.managed` pointer slot, which a later
kernel can observe as null or stale.

Parse one validated ELF view carrying normalized wide section/program counts
and use it for both sizing and traversal. Add an extended-section-table visitor
case containing a managed symbol and verify its backing is published before
module launch.

### [MEDIUM] Publish every new failure through HIP's last-error channel

The thread-local error setter is defined only in `api.c` and is reached by
`HIP_RETURN_ERROR` (`libhrx/src/binding/hip/api.c:962-987`). The PR's new
library/linker translation units return validation, file, state, and compiler
errors directly. The PR also changes both registered-symbol async-copy
paths to use an internal lookup plus new bounds checks, then returns their
accumulated error directly (`api.c:10012-10047,10101-10137`). These failures
therefore do not update `hipPeekAtLastError`, whose public contract is the last
error returned by any HIP runtime call
([reference](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hip_runtime_api/modules/error_handling.html)).

Move error publication into a shared internal return helper available to every
HIP implementation unit and route all new non-success exits through it. Add
table-driven cases which clear TLS, provoke symbol lookup/bounds,
library-file, and linker-state failures, and compare the direct return with
both peek and get-last-error.

### [MEDIUM] Return the library API's documented error for an invalid filename

`hipLibraryLoadFromFile` maps every failed `fopen` to
`hipErrorFileNotFound` (`libhrx/src/binding/hip/library.c:511-530`). The public
library API documents `hipErrorInvalidValue`, and pinned ROCm returns it when
the file does not exist
([API](https://rocm.docs.amd.com/projects/HIP/en/docs-7.2.4/doxygen/html/group___module.html),
[implementation](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_library.cpp#L160-L171)).
`hipErrorFileNotFound` is separately part of `hipModuleLoad`'s contract. Map an
invalid or inaccessible library path to `hipErrorInvalidValue` and add a
nonexistent-file conformance case distinct from module loading.

### [MEDIUM] Canonicalize one module function handle per symbol

Every successful `hipModuleGetFunction` calls the opaque-handle factory even
for the same module/name (`libhrx/src/binding/hip/api.c:12897-12924`). Each call
allocates a record, retains the module, consumes a token, inserts a registry
entry, and prepends a process-global retirement entry
(`libhrx/src/binding/hip/function_handle.c:81-110`). There is no per-function
destroy API; all duplicates remain until module unload, whose retirement scans
the full list (`function_handle.c:146-174`). Repeated read-only queries
therefore grow memory, registry capacity, retained references, and unload work
linearly, with eventual host OOM possible.

Intern a synchronized handle per module symbol/export ordinal and return it on
repeated lookup, as the library kernel table already does. Retire each record
once at unload. Add repeated and concurrent same-symbol queries with a counting
allocator and assert stable live-record/module-retain counts after warm-up.
Run that lifecycle case with leak detection enabled: the production-DSO target
currently sets `ASAN_OPTIONS=detect_leaks=0` for its entire process in both
build systems (`libhrx/cts/BUILD.bazel:256-273`;
`libhrx/cts/CMakeLists.txt:406-436`). Split the COMGR completion case if its
residency exception must remain, rather than masking leaks from the handle
lifecycle test.

### Verification

- `git diff --check` and Buildifier checks pass at `087e5d73c`.
- A clean host-only CMake configuration with AMDGPU, Vulkan, WebGPU, Loom, CTS,
  and passthrough disabled built the candidate HIP DSO and changed common/HIP
  units. Six focused host tests passed: common graph-exec, graph,
  kernel-arguments, and fat-binary plus HIP handle-registry and
  function-handle.
- Current public CI is green for every source-bearing job. The gfx942 and
  gfx1201 CMake GPU lanes each passed 114/114 tests, including the new
  production-DSO test.
- No local ROCm/GPU execution or performance benchmark was run.

zjgarvey (approved):

zjgarvey (approved):

PR #564, opened 2026-09-10
[HIP] Complete device query and error reporting contracts

AWoloszyn:
## Summary

  This change completes the HIP binding’s device-management and error-reporting behavior without modifying the HAL or AMDGPU driver.

  It adds consistent support for:

  - Device selection, capability-based selection, and current-device queries
  - Device properties, UUIDs, PCI identity, limits, cache configuration, and shared-memory configuration
  - Legacy versioned hipDeviceProp_t layouts
  - Supported device attributes, including unified addressing
  - Rejection of unknown or unsupported device attributes without modifying output values
  - Complete HIP error names and descriptions
  - Thread-local propagation of execution, handle, launch, and memset errors
  - Runtime initialization before device-dependent argument validation
  - Consistent kernel launch validation and error reporting

  The UUID path preserves the complete 16-byte HAL device identity. Kernel launch entry points now share one validation path for dynamic shared-memory limits, ensuring consistent errors and
  stream-reference ownership.

### DIFF
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -321,6 +321,57 @@ hrx_cc_test(
 )
 
 hrx_cc_test(
+    name = "hip_device_attribute_validation_api_test",
+    srcs = [
+        "api.h",
+        "hip_device_attribute_validation_api_test.cc",
+        "hip_device_attribute_validation_test_shim.c",
+        "hip_device_attribute_validation_test_shim.h",
+    ],
+    # Exercise validation and TLS behavior through the exact public shared
+    # object built by this target graph.
+    data = [":amdhip64"],
+    env = {
+        "HRX_TEST_LIBAMDHIP64": "$(location :amdhip64)",
+    },
+    tags = [
+        "iree-run-requirement=runtime.resource.amd_gpu",
+        "runtime-resource=amd-gpu",
+    ],
+    target_compatible_with = _LINUX_COMPATIBILITY,
+    deps = [
+        "//build_tools/bazel:dl",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
+    name = "hip_device_attribute_no_device_api_test",
+    srcs = [
+        "api.h",
+        "hip_device_attribute_no_device_api_test.cc",
+        "hip_device_attribute_validation_test_shim.c",
+        "hip_device_attribute_validation_test_shim.h",
+    ],
+    data = [":amdhip64"],
+    env = {
+        "HRX_TEST_LIBAMDHIP64": "$(location :amdhip64)",
+        "ROCR_VISIBLE_DEVICES": "-1",
+    },
+    tags = [
+        "iree-run-requirement=runtime.resource.amd_gpu",
+        "runtime-resource=amd-gpu",
+    ],
+    target_compatible_with = _LINUX_COMPATIBILITY,
+    deps = [
+        "//build_tools/bazel:dl",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
     name = "hip_execution_resource_api_test",
     srcs = ["hip_execution_resource_api_test.cc"],
     data = [":amdhip64"],
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -347,6 +347,30 @@ hrx_cc_test(
 )
 
 hrx_cc_test(
+    name = "hip_unified_addressing_attribute_api_test",
+    srcs = [
+        "api.h",
+        "hip_unified_addressing_attribute_api_test.cc",
+    ],
+    # Exercise the public entry points from the exact shared object built by
+    # this target graph instead of linking a test-only implementation.
+    data = [":amdhip64"],
+    env = {
+        "HRX_TEST_LIBAMDHIP64": "$(location :amdhip64)",
+    },
+    tags = [
+        "iree-run-requirement=runtime.resource.amd_gpu",
+        "runtime-resource=amd-gpu",
+    ],
+    target_compatible_with = _LINUX_COMPATIBILITY,
+    deps = [
+        "//build_tools/bazel:dl",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
     name = "hip_device_attribute_no_device_api_test",
     srcs = [
         "api.h",
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -414,6 +414,28 @@ endif()
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 iree_cc_test(
   NAME
+    hip_unified_addressing_attribute_api_test
+  SRCS
+    "api.h"
+    "hip_unified_addressing_attribute_api_test.cc"
+  DATA
+    ::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+  LABELS
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+iree_cc_test(
+  NAME
     hip_device_attribute_no_device_api_test
   SRCS
     "api.h"
@@ -578,7 +600,8 @@ foreach(_hrx_hip_dso_test
     libhrx_src_binding_hip_hip_device_attribute_no_device_api_test
     libhrx_src_binding_hip_hip_device_attribute_validation_api_test
     libhrx_src_binding_hip_hip_get_proc_address_test
-    libhrx_src_binding_hip_hip_launch_validation_api_test)
+    libhrx_src_binding_hip_hip_launch_validation_api_test
+    libhrx_src_binding_hip_hip_unified_addressing_attribute_api_test)
   if(TARGET ${_hrx_hip_dso_test})
     target_compile_definitions(${_hrx_hip_dso_test}
       PRIVATE
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -390,6 +390,55 @@ endif()
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 iree_cc_test(
   NAME
+    hip_device_attribute_validation_api_test
+  SRCS
+    "api.h"
+    "hip_device_attribute_validation_api_test.cc"
+    "hip_device_attribute_validation_test_shim.c"
+    "hip_device_attribute_validation_test_shim.h"
+  DATA
+    ::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+  LABELS
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+iree_cc_test(
+  NAME
+    hip_device_attribute_no_device_api_test
+  SRCS
+    "api.h"
+    "hip_device_attribute_no_device_api_test.cc"
+    "hip_device_attribute_validation_test_shim.c"
+    "hip_device_attribute_validation_test_shim.h"
+  DATA
+    ::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+    "ROCR_VISIBLE_DEVICES=-1"
+  LABELS
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+iree_cc_test(
+  NAME
     hip_get_proc_address_test
   SRCS
     "hip_get_proc_address_test.cc"
@@ -526,6 +575,8 @@ endif()
 # The build-tree path is only valid for the in-tree CTest run.
 foreach(_hrx_hip_dso_test
     libhrx_src_binding_hip_hip_api_name_test
+    libhrx_src_binding_hip_hip_device_attribute_no_device_api_test
+    libhrx_src_binding_hip_hip_device_attribute_validation_api_test
     libhrx_src_binding_hip_hip_get_proc_address_test
     libhrx_src_binding_hip_hip_launch_validation_api_test)
   if(TARGET ${_hrx_hip_dso_test})
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -362,31 +362,6 @@ iree_cc_test(
 )
 endif()
 
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND IREE_HAL_DRIVER_AMDGPU)
-iree_cc_test(
-  NAME
-    hip_execution_resource_api_test
-  SRCS
-    "hip_execution_resource_api_test.cc"
-  DATA
-    ::amdhip64
-  DEPS
-    ${CMAKE_DL_LIBS}
-    ::execution_resource
-    iree::testing::gtest
-    iree::testing::gtest_main
-    libhrx_defs
-  ENV
-    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
-  LABELS
-    "iree-run-requirement=runtime.resource.amd_gpu"
-    "runtime-resource=amd-gpu"
-  SANITIZER_SUPPRESSIONS
-    lsan
-    hsa
-)
-endif()
-
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 iree_cc_test(
   NAME
@@ -455,6 +430,31 @@ iree_cc_test(
   LABELS
     "iree-run-requirement=runtime.resource.amd_gpu"
     "runtime-resource=amd-gpu"
+)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND IREE_HAL_DRIVER_AMDGPU)
+iree_cc_test(
+  NAME
+    hip_execution_resource_api_test
+  SRCS
+    "hip_execution_resource_api_test.cc"
+  DATA
+    ::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    ::execution_resource
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+  LABELS
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+  SANITIZER_SUPPRESSIONS
+    lsan
+    hsa
 )
 endif()
 
--- a/libhrx/src/binding/hip/abi_stubs.c
+++ b/libhrx/src/binding/hip/abi_stubs.c
@@ -730,10 +730,16 @@ HIPAPI hipError_t hipExtEnableLogging(void) { return hipErrorNotSupported; }
 HIPAPI hipError_t hipExtGetLinkTypeAndHopCount(int device1, int device2,
                                                uint32_t* linktype,
                                                uint32_t* hopcount) {
-  (void)device1;
-  (void)device2;
-  (void)linktype;
-  (void)hopcount;
+  if (!linktype || !hopcount || device1 == device2 || device1 < 0 ||
+      device2 < 0) {
+    return hipErrorInvalidValue;
+  }
+  int device_count = 0;
+  hipError_t result = hipGetDeviceCount(&device_count);
+  if (result != hipSuccess) return result;
+  if (device1 >= device_count || device2 >= device_count) {
+    return hipErrorInvalidDevice;
+  }
   return hipErrorNotSupported;
 }
 
--- a/libhrx/src/binding/hip/abi_stubs.c
+++ b/libhrx/src/binding/hip/abi_stubs.c
@@ -39,7 +39,118 @@ typedef struct hipArrayMemoryRequirements {
   size_t alignment;
   size_t size;
 } hipArrayMemoryRequirements;
-typedef struct hipDeviceProp_tR0000 hipDeviceProp_tR0000;
+typedef struct hipDeviceProp_tR0000 {
+  // Device name.
+  char name[256];
+  // Global memory size in bytes.
+  size_t totalGlobalMem;
+  // Workgroup-local memory size in bytes.
+  size_t sharedMemPerBlock;
+  // Registers available per workgroup.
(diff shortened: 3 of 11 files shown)


zjgarvey:
Requesting changes for the following 6 high-, 4 medium-, and 2 low-severity issues.

1. **[HIGH] Make every legacy device-property entry point use the R0000 ABI.** The `hip_4.2` export node lists the unsuffixed `hipChooseDevice` and `hipGetDeviceProperties` (`libhrx/src/binding/hip/amdhip64.map:25,105`), but there is no unsuffixed choose-device definition, and the unsuffixed property function clears and fills the larger current/R0600 type (`libhrx/src/binding/hip/api.c:2139,2183`). `hipGetProcAddress_spt` also discards `hip_version` before falling back to that unsuffixed symbol (`libhrx/src/binding/hip/abi_stubs.c:2173-2178`). A pre-6.0 caller can therefore fail to resolve `hipChooseDevice` or pass an R0000-sized object to a function that writes the R0600 layout, corrupting adjacent memory. Upstream keeps the unsuffixed facade on R0000 and aliases newer source to R0600 ([property facade](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L902-L904), [choose facade](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L878-L880)). Please add bounded R0000 wrappers for both unsuffixed symbols, route both proc-address APIs through the same version-aware selector, and add exact-DSO canary tests below and at version 600.

2. **[HIGH] Normalize `_spt` lookup flags without overriding explicit choices.** `hipGetDriverEntryPoint_spt` forwards default flag 0 unchanged, while the helper probes `<name>_spt` only for flag 2 (`libhrx/src/binding/hip/abi_stubs.c:517-555,2166-2170`), so default resolves the legacy-stream function. `hipGetProcAddress_spt` has the opposite problem: it forces flag 2 even when the caller explicitly asks for the legacy stream (`libhrx/src/binding/hip/abi_stubs.c:2173-2178`). Upstream normalizes only `DEFAULT` to `PER_THREAD_DEFAULT_STREAM` in each `_spt` wrapper ([driver lookup](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L772-L776), [proc lookup](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L891-L898)). With the current driver lookup, two host threads can obtain the unsuffixed launch, enqueue a waiting kernel and its producer onto the one legacy stream, and hang because the producer is ordered behind the waiter. Please share one default-only normalization helper, preserve explicit legacy/per-thread flags, reject invalid flags, and test both pointer identity and two-thread null-stream ordering.

3. **[HIGH] Keep device flags in synchronized public state.** `hipSetDeviceFlags` now writes `primary_context->flags` while holding `primary_context_mutex` (`libhrx/src/binding/hip/api.c:3747-3772`; `libhrx/src/binding/common/device.c:249-269`), but `hipGetDeviceFlags` reads that non-atomic enum/bitfield structure without the mutex (`libhrx/src/binding/hip/api.c:3791-3806`). Concurrent set/get on the shared primary context is therefore a C data race. The same representation also conflates public flags with backend capabilities: initialization sets `map_host_memory = true` (`libhrx/src/binding/common/init.c:269-273`), so the initial getter reports `hipDeviceMapHost`, while a scheduling-only set constructs a zeroed internal value and clears that supposedly fixed bit. Upstream reports separately stored public flags ([getter/setter](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L715-L727)). Please store the user-visible schedule flags under the device mutex, use the existing locked snapshot accessor for reads, preserve fixed backend capabilities, and add default, round-trip, and barrier-controlled concurrent tests.

4. **[HIGH] Put fatal illegal-address state in shared device/runtime state and publish it from every observer.** The new fatal latch is part of an `IREE_THREAD_LOCAL` object (`libhrx/src/binding/hip/api.c:901-904`), and only `hipDeviceSynchronize` marks `hipErrorIllegalAddress` sticky (`libhrx/src/binding/hip/api.c:3650-3659`). Stream/event synchronize and query return the same mapped backend failure directly (`libhrx/src/binding/hip/api.c:11518-11604,12338-12417`), without updating even the observing thread's last-error state. Thus thread B can continue returning cached device data or mutating runtime state after thread A observes a fatal GPU error; a stream/event observer can also receive `hipErrorIllegalAddress` while `hipPeekAtLastError` remains `hipSuccess`. Current ROCm keeps a process-shared GPU error and checks it at API entry ([shared state](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/rocclr/device/device.cpp#L339), [entry checks](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L142-L175)). Please keep fatal persistence, but move the latch to synchronized shared state, publish it after releasing retained handles from every completion observer, and test stream/event plus cross-thread behavior.

5. **[HIGH] Treat an empty valid-device list as a no-op.** `hipSetValidDevices` currently computes `selected_device = len > 0 ? device_arr[0] : 0` and installs it unless an explicit `hipSetDevice` occurred (`libhrx/src/binding/hip/api.c:1961-1980`). A thread that selected device 1 with a preference list can therefore call `hipSetValidDevices(NULL, 0)` and silently move later allocations and launches to device 0. ROCm returns success immediately for length zero without changing selection ([reference behavior](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L848-L875)). Please return after required initialization when `len == 0` and add a repeated-call test on a two-device configuration.

6. **[HIGH] Publish errors produced by public functions in `abi_stubs.c`.** The only last-error setter is private to `api.c` (`libhrx/src/binding/hip/api.c:984-988`), so newly implemented public paths in `abi_stubs.c` return their own validation, unsupported, and lookup errors without updating TLS. For example, after clearing last error, `hipDeviceComputeCapability(NULL, &minor, 0)` returns `hipErrorInvalidValue` at `libhrx/src/binding/hip/abi_stubs.c:841-848`, but `hipPeekAtLastError` still returns `hipSuccess`. Driver lookup and the new choose/property paths have the same split. Please expose a sticky-preserving internal result publisher (or put thin public wrappers in `api.c`), route every new public failure through it, and add exact-DSO tests covering direct validation, delegated failure, unsupported, missing-symbol, and cross-thread isolation cases.

7. **[MEDIUM] Populate the R0000 `gcnArch` field.** The converter zeros the legacy object, copies `canMapHostMemory`, skips the intervening numeric `gcnArch`, and then copies `gcnArchName` (`libhrx/src/binding/hip/abi_stubs.c:618-685`). A successful query on gfx942 consequently returns a valid name with `gcnArch == 0`, which can misroute legacy architecture selection. ROCm derives the numeric value from the ISA components ([reference assignment](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L775-L777)). Please populate it from structured architecture metadata or validated `gfx` parsing and cover both three- and four-digit targets.

8. **[MEDIUM] Return the documented result for a missing driver symbol.** `hrx_hip_driver_entry_point_lookup` correctly nulls the function and sets status to symbol-not-found, but returns `hipErrorNotFound` (`libhrx/src/binding/hip/abi_stubs.c:517-555`). The upstream driver-entry contract returns `hipErrorInvalidValue` while retaining the detailed status ([reference lookup](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L721-L766)). Please translate the missing-symbol result at the public lookup boundary and test return value, null output, status, and last-error publication together for regular and `_spt` entry points.

9. **[MEDIUM] Make limit validation cover the declared domain.** The header declares `hipExtLimitScratchMin`, `Max`, and `Current` before `hipLimitRange` (`libhrx/src/binding/hip/api.h:536-547`), but the mapper supports none of them. Get/set reject only equality with the sentinel, so values above it return unsupported rather than invalid, and these checks occur before initialization (`libhrx/src/binding/hip/api.c:4800-4831,4862-4890`). Upstream rejects the complete out-of-range domain and implements the scratch queries/current setter ([get](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L522-L549), [set](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L614-L639)). Please either implement the declared scratch values or narrow the surface, reject every value outside the range, and preserve initialization-before-argument-validation behavior.

10. **[MEDIUM] Make the GPU tests exercise the production behavior they assert.** The modified launch test expects `hipErrorInvalidResourceHandle` for all three stale-stream launches (`libhrx/src/binding/hip/hip_launch_validation_api_test.cc:358-381`), while head converts that result to `hipErrorInvalidValue` only in regular `hipLaunchKernel` (`libhrx/src/binding/hip/api.c:13520-13527`). Its target lacks the `runtime-resource=amd-gpu` tag selected by the GPU jobs (`libhrx/src/binding/hip/BUILD.bazel:439-458`; `.github/workflows/ci_core_linux.yml:95-143`), and the fixture skips whenever `hipInit` fails (`libhrx/src/binding/hip/hip_launch_validation_api_test.cc:189-192`), so the contradiction need not fail CI. The three new attribute tests have GPU resource tags but use `_LINUX_COMPATIBILITY` instead of the defined AMDGPU backend compatibility constraint (`libhrx/src/binding/hip/BUILD.bazel:30-38,323-390`). Please align the stale-stream contract, tag and constrain all GPU-dependent targets on AMDGPU, make initialization failure fatal in a resource-guaranteed lane, and add the canary, identity/PCI round-trip, fatal/TLS, and memset/error failure cases needed by the new public behavior.

11. **[LOW] Avoid scanning physical identities for non-PCI attributes.** `hipDeviceGetAttribute` unconditionally calls `iree_hip_physical_device_identity` before its switch (`libhrx/src/binding/hip/api.c:85-110,2420`), although only the three PCI cases consume the result (`libhrx/src/binding/hip/api.c:2515-2534`). Ordinary cached attribute queries therefore gain an O(P) scan over physical identities. Please move the lookup into the PCI cases or cache the selected identity during registry initialization, and record a representative multi-device attribute-sweep measurement.

12. **[LOW] Consolidate the exact-DSO test harness.** Candidate-path resolution, `dlopen` ownership, typed symbol lookup, initialization, teardown, and last-error helpers are duplicated across four fixtures (`libhrx/src/binding/hip/hip_device_attribute_validation_api_test.cc:27`, `libhrx/src/binding/hip/hip_unified_addressing_attribute_api_test.cc:24`, `libhrx/src/binding/hip/hip_device_attribute_no_device_api_test.cc:24`, `libhrx/src/binding/hip/hip_launch_validation_api_test.cc:20`). The copies already differ in path canonicalization and `dladdr` symbol verification. Please extract a test-only RAII DSO owner/resolver so loader-lifetime and exact-binding fixes apply uniformly.

zjgarvey:
Reviewed commit `946cc2837cd5c988aaa9d9941a8c65286e591be7`. Requesting changes for the following 7 high-, 3 medium-, and 1 low-severity issues.

1. **[HIGH] Make fatal-state enforcement and last-error publication a mandatory public API boundary.** At ordinary public API entry, the shared latch is enforced only through `iree_hip_ensure_initialized`; publication and the error-inspection getters also consult it, but public paths invoke those helpers inconsistently (`libhrx/src/binding/hip/api.c:1422-1426`; `libhrx/src/binding/hip/error_state.c:34-73`). Explicit streams are retained without the initialization gate (`libhrx/src/binding/hip/api.c:1591-1606`), version queries return raw success (`libhrx/src/binding/hip/api.c:1780-1808`), `hipCtxSynchronize` and terminal launch results return raw mapped errors (`libhrx/src/binding/hip/api.c:4607-4620,13518-13536`), zero-byte memcpy/memset return early (`libhrx/src/binding/hip/api.c:6875-6881,10291-10298`), and public stubs such as `hipProfilerStart` return unsupported directly (`libhrx/src/binding/hip/abi_stubs.c:1708-1715`). Thus an unsupported stub does not update last error; after thread A latches an illegal address, thread B can still get success from a version query or zero-byte operation; and `hipCtxSynchronize` can return the first illegal-address observation without latching it for other threads. ROCm applies one entry/return error-state contract to ordinary APIs ([entry and return macros](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L141-L186)). Please route every `hipError_t` public path through common entry and final-result wrappers, with explicit exceptions for error inspection and teardown/recovery, and publish only after releasing retained resources. Add exact-DSO tests for a raw unsupported stub, raw-success query, explicit-stream call, and first-observer completion failure on same and different threads.

2. **[HIGH] Keep ordinary last error separate from the extended last-command result.** The remediation newly routes `hipStreamQuery` and `hipEventQuery` through `iree_hip_error_state_publish`, which stores every non-success result in one TLS slot and therefore publishes the normal polling result `hipErrorNotReady` into `hipPeekAtLastError` (`libhrx/src/binding/hip/error_state.c:39-52`; `libhrx/src/binding/hip/api.c:11544-11558,12363-12371`). The error-state rewrite also preserves the pre-existing implementation of `hipExtGetLastError` as another getter for that same ordinary-error slot (`libhrx/src/binding/hip/api.c:25049-25072`). ROCm excludes `NotReady` from ordinary last-error updates and records every command's result separately ([update rules](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L165-L175), [extended getter](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_error.cpp#L12-L23)). Currently a pending-event poll turns a clean ordinary last error into `NotReady`; an invalid call followed by a successful call also makes the extended getter report/clear the earlier ordinary error instead of the latest command. Please maintain separate ordinary and last-command TLS fields, exclude `NotReady` from the ordinary slot, update the command slot on every public return, and test both getters around pending queries and a failure-success sequence.

3. **[HIGH] Make fatal reset generation-safe across threads.** A fatal observation is copied into sticky TLS, but `iree_hip_error_state_reset` clears only the process latch and the calling thread's TLS (`libhrx/src/binding/hip/error_state.c:28-80`), and `hipHALDeinit` invokes it only on the teardown thread (`libhrx/src/binding/hip/api.c:1753-1765`). A barrier-controlled host test against this commit made one thread sticky, reset from another, and deterministically read `hipErrorIllegalAddress` (700) rather than `hipSuccess` afterward. That surviving thread cannot clear the old runtime's value with `hipGetLastError`, and later nonfatal publications are ignored because its sticky bit remains set. Please associate shared and TLS error state with a monotonically increasing runtime generation (or avoid caching shared fatal state in TLS), advance it during teardown, and lazily invalidate every thread's old state. Add a two-thread public-DSO test in which the fatal observer survives deinit/reinit performed by the other thread.

4. **[HIGH] Represent valid-device preference independently from active context and runtime lifetime.** `hipSetValidDevices` calls the same helper as `hipSetDevice`, so a nonempty preference eagerly creates and installs a primary context (`libhrx/src/binding/hip/api.c:1859-1889,1923-1942`). Its precedence is a thread-lifetime TLS Boolean that survives runtime deinit/reinit on every surviving thread because `hipHALDeinit` never resets or generation-stamps it (`libhrx/src/binding/hip/api.c:1753-1765,1857`). On two GPUs, `hipSetDevice(0)`, full deinit/reinit, then `hipSetValidDevices({1}, 1)` returns success but ignores device 1 because the stale Boolean remains true; subsequent work defaults to device 0. Even without teardown, the preference call makes the selected primary context active, whereas ROCm changes the preferred/current-device TLS pointer without creating or marking the primary device context active ([valid-device path](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L848-L874)). Please keep a runtime-generation-scoped preference/provenance record separate from the active context and add primary-context-state plus same/surviving-thread deinit tests.

5. **[HIGH] Reset public device flags and report only the ROCm-visible scheduling state.** The PR now persists caller flags in per-device state (`libhrx/src/binding/common/device.c:249-300`), but `hipDevicePrimaryCtxReset` releases the context and pools without restoring that state (`libhrx/src/binding/hip/api.c:4082-4160`). Setting blocking synchronization, calling `hipDeviceReset`, and querying again therefore returns/reuses blocking rather than the initial default, despite the reset contract saying all device state returns to initial values (`libhrx/src/binding/hip/api.c:3627-3632`). The converter also returns `hipDeviceMapHost` and `hipDeviceLmemResizeToMax` bits (`libhrx/src/binding/hip/api.c:4168-4234`), while ROCm accepts those options but retains/reports the scheduling mask and resets it with the device ([getter](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L715-L722), [setter](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L800-L845), [reset](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L137-L153)). Please restore the synchronized public snapshot during reset, keep capability/option bits separate, mask the getter to the public scheduling state, and add reset plus non-scheduling-bit cases to the exact-DSO flag test.

6. **[HIGH] Complete the supported attribute table before rejecting the default case.** The new default correctly aims to reject unknown values, but 36 numeric selectors handled by the pinned ROCm switch have no case and now return `hipErrorInvalidValue` (`libhrx/src/binding/hip/api.c:2375-2583`). Examples include `hipDeviceAttributeEccEnabled`, `AsyncEngineCount`, `ComputeMode`, `MaxRegistersPerMultiprocessor`, `VirtualMemoryManagementSupported`, and `PciChipId`; many corresponding facts are already populated in `hipDeviceProp_t` (`libhrx/src/binding/hip/api.c:2183-2294`). A valid `AsyncEngineCount` query therefore errors even though this implementation reports `asyncEngineCount = 2`. The header also stops before newer selectors that the pinned runtime handles, while the negative test labels numeric values 93 and 10021 as gaps even though they are `DmaBufSupported` and `ExpertSchedMode` (`libhrx/src/binding/hip/hip_device_attribute_validation_api_test.cc:235-238`; [ROCm attribute switch](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L145-L470)). Please retain the rejecting default, add the complete supported numeric set using shared property/backend facts, update the public enum, replace those two false gaps, and add table-driven positive coverage including supported-zero results. Also remove `hipDeviceAttributeAmdSpecificBegin` from the unsupported-marker table: it aliases the real `ClockInstructionRate` value 10000 and cannot be distinguished at runtime (`libhrx/src/binding/hip/api.h:410-411`).

7. **[HIGH] Decode alphanumeric GPU targets before publishing architecture results.** The new R0000 conversion exposes a wrong result for `gfx90a` while reusing an existing wrong compute-capability cache. The pre-existing cached parser uses `sscanf("gfx%d")`; for `gfx90a` it accepts only `90` and selects the 7.5 fallback despite its own comment specifying 9.0 (`libhrx/src/binding/common/init.c:57-85`). The newly added legacy parser accepts decimal digits only and rejects the trailing `a`; its caller ignores that failure after zeroing the R0000 object (`libhrx/src/binding/hip/device_properties.c:13-28`; `libhrx/src/binding/hip/abi_stubs.c:469-502`). Thus MI200 callers receive successful but wrong results: compute capability 7.5, and `gcnArchName="gfx90a"` with `gcnArch == 0`. Pinned ROCm represents `gfx90a` as ISA components `(9,0,10)` and derives legacy `gcnArch` as 910 ([ISA table](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/rocclr/device/device.cpp#L153-L164), [legacy assignment](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L775-L777)). Please use one structured ISA decoder for compute capability and legacy encoding, handle alphabetic stepping and feature suffixes, propagate parse failure, and cover `gfx90a`/`gfx90a:xnack+` alongside numeric targets.

8. **[MEDIUM] Preserve the device-fault class before making it sticky.** The AMDGPU event bridge maps both memory-access faults and hardware exceptions to `IREE_STATUS_ABORTED` (`runtime/src/iree/hal/drivers/amdgpu/system_event.c:113-133`), while this PR maps every aborted status to `hipErrorIllegalAddress` and latches it (`libhrx/src/binding/hip/api.c:1046-1076`; `libhrx/src/binding/hip/error_state.c:39-48`). A hardware exception or queue reset is consequently reported as an illegal address for the rest of the runtime lifetime. ROCm preserves the event distinction and maps a hardware exception to launch failure rather than illegal address ([fault conversion](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/rocclr/device/rocm/rocdevice.cpp#L3654-L3708), [HIP mapping](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/trace_helper.h#L140-L176)). Please carry typed event provenance through HAL, map each class before publication, publish and latch each result according to its correct fault class, and add injected tests for memory fault, hardware exception, and unknown device failure.

9. **[MEDIUM] Translate missing symbols at both rewritten proc-address boundaries.** This error-code mismatch predates the PR, but the rewritten shared lookup and both proc wrappers preserve it while the driver wrappers now translate it. The lookup correctly nulls the output, sets symbol-not-found status, and returns internal `hipErrorNotFound` (`libhrx/src/binding/hip/api.c:24901-24927`); `hipGetProcAddress` and `hipGetProcAddress_spt` publish that code unchanged (`libhrx/src/binding/hip/api.c:24971-24975`; `libhrx/src/binding/hip/abi_stubs.c:2059-2071`). Pinned ROCm returns `hipErrorInvalidValue` while retaining the detailed status ([lookup boundary](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device.cpp#L863-L880)). Please apply the same translation to both proc wrappers and test direct result, null pointer, status, and last error for regular and `_spt` calls.

10. **[MEDIUM] Implement the advertised scratch-limit operations or narrow the compatibility surface.** `hipExtLimitScratchMin`, `Max`, and `Current` remain declared, but the mapper sends all three to an invalid sentinel, and the new test codifies `hipErrorUnsupportedLimit` (`libhrx/src/binding/hip/api.h:536-548`; `libhrx/src/binding/hip/api.c:854-878,4756-4787,4818-4845`; `libhrx/src/binding/hip/hip_device_contracts_api_test.cc:211-227`). Pinned ROCm returns all three values and permits setting Current ([get](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L522-L550), [set](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_device_runtime.cpp#L614-L640)). Please plumb immutable min/max and a validated mutable current budget through HRX/HAL, or remove the declarations if that ABI choice is acceptable, then replace unsupported expectations with bounds and round-trip tests.

11. **[LOW] Avoid retrying a strong compare-exchange after the fatal latch is set.** Every publication of `hipErrorIllegalAddress` attempts `compare_exchange_strong`, even when the global latch already contains that value; the common entry gate feeds the latched value back into the publisher (`libhrx/src/binding/hip/error_state.c:39-48`; `libhrx/src/binding/hip/api.c:1422-1425`). After a fault, each rejected call therefore performs a failed RMW on the same cache line and then another acquire load, causing avoidable contention while many host threads unwind. Please load once, attempt CAS only for a transition from success, reuse the observed value on failure, and add a counter-backed multithreaded test. No healthy-path timing regression is claimed here.

zjgarvey (approved):

zjgarvey (approved):

PR #565, opened 2026-09-10
[HIP] Complete per-thread array copy entry points

AWoloszyn:
The change:

  - Implements hipMemcpy2DFromArrayAsync_spt, hipMemcpy2DToArrayAsync_spt, and hipMemcpyFromArray_spt.
  - Resolves null streams for asynchronous _spt calls to the per-thread default stream.
  - Routes synchronous _spt copies through the per-thread stream and waits only for that stream’s timeline.
  - Preserves the existing device-wide synchronization behavior of the legacy non-_spt APIs.
  - Corrects the existing synchronous 2D _spt wrappers, which previously forwarded through the legacy default-stream entry points.
  - Adds DSO-level coverage for symbol availability and GPU-backed array-copy behavior.

### DIFF
--- a/libhrx/src/binding/hip/BUILD.bazel
+++ b/libhrx/src/binding/hip/BUILD.bazel
@@ -321,6 +321,29 @@ hrx_cc_test(
 )
 
 hrx_cc_test(
+    name = "hip_array_copy_spt_api_test",
+    srcs = [
+        "api.h",
+        "hip_array_copy_spt_api_test.cc",
+    ],
+    data = [":amdhip64"],
+    env = {
+        "HRX_TEST_LIBAMDHIP64": "$(location :amdhip64)",
+    },
+    sanitizer_suppressions = hsa_suppressions,
+    tags = [
+        "iree-run-requirement=runtime.resource.amd_gpu",
+        "runtime-resource=amd-gpu",
+    ],
+    target_compatible_with = _HIP_GPU_COMPATIBILITY,
+    deps = [
+        "//build_tools/bazel:dl",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
+hrx_cc_test(
     name = "hip_execution_resource_api_test",
     srcs = ["hip_execution_resource_api_test.cc"],
     data = [":amdhip64"],
--- a/libhrx/src/binding/hip/CMakeLists.txt
+++ b/libhrx/src/binding/hip/CMakeLists.txt
@@ -365,6 +365,31 @@ endif()
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND IREE_HAL_DRIVER_AMDGPU)
 iree_cc_test(
   NAME
+    hip_array_copy_spt_api_test
+  SRCS
+    "api.h"
+    "hip_array_copy_spt_api_test.cc"
+  DATA
+    ::amdhip64
+  DEPS
+    ${CMAKE_DL_LIBS}
+    iree::testing::gtest
+    iree::testing::gtest_main
+    libhrx_defs
+  ENV
+    "HRX_TEST_LIBAMDHIP64=$<TARGET_FILE:libhrx::src::binding::hip::amdhip64>"
+  LABELS
+    "iree-run-requirement=runtime.resource.amd_gpu"
+    "runtime-resource=amd-gpu"
+  SANITIZER_SUPPRESSIONS
+    lsan
+    hsa
+)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND IREE_HAL_DRIVER_AMDGPU)
+iree_cc_test(
+  NAME
     hip_execution_resource_api_test
   SRCS
     "hip_execution_resource_api_test.cc"
@@ -526,6 +551,7 @@ endif()
 # The build-tree path is only valid for the in-tree CTest run.
 foreach(_hrx_hip_dso_test
     libhrx_src_binding_hip_hip_api_name_test
+    libhrx_src_binding_hip_hip_array_copy_spt_api_test
     libhrx_src_binding_hip_hip_get_proc_address_test
     libhrx_src_binding_hip_hip_launch_validation_api_test)
   if(TARGET ${_hrx_hip_dso_test})
--- a/libhrx/src/binding/hip/abi_stubs.c
+++ b/libhrx/src/binding/hip/abi_stubs.c
@@ -1979,20 +1979,53 @@ HIPAPI hipError_t hipMemcpy2DAsync_spt(void* dst, size_t dpitch,
                           resolved_stream);
 }
 
+HIPAPI hipError_t hipMemcpy2DFromArrayAsync_spt(void* dst, size_t dpitch,
+                                                hipArray_const_t src,
+                                                size_t wOffset, size_t hOffset,
+                                                size_t width, size_t height,
+                                                hipMemcpyKind kind,
+                                                hipStream_t stream) {
+  hipStream_t resolved_stream = NULL;
+  hipError_t result = hrx_hip_spt_stream_or_explicit(stream, &resolved_stream);
+  if (result != hipSuccess) return result;
+  return hipMemcpy2DFromArrayAsync(dst, dpitch, src, wOffset, hOffset, width,
+                                   height, kind, resolved_stream);
+}
+
 HIPAPI hipError_t hipMemcpy2DFromArray_spt(void* dst, size_t dpitch,
                                            hipArray_const_t src, size_t wOffset,
                                            size_t hOffset, size_t width,
                                            size_t height, hipMemcpyKind kind) {
-  return hipMemcpy2DFromArray(dst, dpitch, src, wOffset, hOffset, width, height,
-                              kind);
+  hipStream_t stream = NULL;
+  hipError_t result = hrx_hip_spt_default_stream(&stream);
+  if (result != hipSuccess) return result;
+  result = hipMemcpy2DFromArrayAsync(dst, dpitch, src, wOffset, hOffset, width,
+                                     height, kind, stream);
+  return result == hipSuccess ? hipStreamSynchronize(stream) : result;
+}
+
+HIPAPI hipError_t hipMemcpy2DToArrayAsync_spt(hipArray_t dst, size_t wOffset,
+                                              size_t hOffset, const void* src,
+                                              size_t spitch, size_t width,
+                                              size_t height, hipMemcpyKind kind,
+                                              hipStream_t stream) {
+  hipStream_t resolved_stream = NULL;
+  hipError_t result = hrx_hip_spt_stream_or_explicit(stream, &resolved_stream);
+  if (result != hipSuccess) return result;
+  return hipMemcpy2DToArrayAsync(dst, wOffset, hOffset, src, spitch, width,
+                                 height, kind, resolved_stream);
 }
 
 HIPAPI hipError_t hipMemcpy2DToArray_spt(hipArray_t dst, size_t wOffset,
                                          size_t hOffset, const void* src,
                                          size_t spitch, size_t width,
                                          size_t height, hipMemcpyKind kind) {
-  return hipMemcpy2DToArray(dst, wOffset, hOffset, src, spitch, width, height,
-                            kind);
+  hipStream_t stream = NULL;
+  hipError_t result = hrx_hip_spt_default_stream(&stream);
+  if (result != hipSuccess) return result;
+  result = hipMemcpy2DToArrayAsync(dst, wOffset, hOffset, src, spitch, width,
+                                   height, kind, stream);
+  return result == hipSuccess ? hipStreamSynchronize(stream) : result;
 }
 
 HIPAPI hipError_t hipMemcpy2D_spt(void* dst, size_t dpitch, const void* src,
--- a/libhrx/src/binding/hip/api.c
+++ b/libhrx/src/binding/hip/api.c
@@ -8139,36 +8139,38 @@ static hipError_t iree_hip_array_row_device_pointer(
 
 static hipError_t iree_hip_array_enqueue_packed_row_copy(
     bool array_is_destination, iree_hal_streaming_deviceptr_t array_ptr,
-    void* external_ptr, size_t byte_count, hipMemcpyKind kind) {
+    void* external_ptr, size_t byte_count, hipMemcpyKind kind,
+    hipStream_t stream) {
   if (array_is_destination) {
     return hipMemcpyAsync((void*)array_ptr, external_ptr, byte_count, kind,
-                          NULL);
+                          stream);
   }
   return hipMemcpyAsync(external_ptr, (const void*)array_ptr, byte_count, kind,
-                        NULL);
+                        stream);
 }
 
 static hipError_t iree_hip_array_enqueue_packed_rows_copy(
     bool array_is_destination, iree_hal_streaming_deviceptr_t array_ptr,
     size_t array_pitch, void* external_ptr, size_t row_width, size_t row_count,
-    hipMemcpyKind kind) {
+    hipMemcpyKind kind, hipStream_t stream) {
   if (array_is_destination) {
     return hipMemcpy2DAsync((void*)array_ptr, array_pitch, external_ptr,
-                            row_width, row_width, row_count, kind, NULL);
+                            row_width, row_width, row_count, kind, stream);
   }
   return hipMemcpy2DAsync(external_ptr, row_width, (const void*)array_ptr,
-                          array_pitch, row_width, row_count, kind, NULL);
+                          array_pitch, row_width, row_count, kind, stream);
 }
 
 // Copies the selected 2D array slice as a packed byte stream. The allocation
 // pitch is an implementation detail, so a range spanning rows must skip its
-// padding rather than expose that padding to the caller. The synchronous APIs
-// enqueue a prefix, complete-row range, and suffix on the default stream, then
-// synchronize once after all successfully queued work has completed.
+// padding rather than expose that padding to the caller. A null stream
+// preserves legacy synchronous completion across the device; other streams
+// synchronize only their own timeline after all successfully queued work has
+// completed.
 static hipError_t iree_hip_array_copy_packed_rows(
     hipArray_const_t array, size_t byte_offset, size_t row_offset,
     void* external_ptr, size_t byte_count, hipMemcpyKind kind,
-    bool array_is_destination) {
+    bool array_is_destination, hipStream_t stream) {
   struct hipArray_st* array_info = NULL;
   hipError_t result = iree_hip_array_retain(array, &array_info);
   if (result != hipSuccess) return result;
@@ -8212,7 +8214,7 @@ static hipError_t iree_hip_array_copy_packed_rows(
       may_have_enqueued_work = true;
       result = iree_hip_array_enqueue_packed_row_copy(
           array_is_destination, array_ptr, (uint8_t*)external_ptr,
-          first_row_byte_count, kind);
+          first_row_byte_count, kind, stream);
     }
     external_offset = first_row_byte_count;
     remaining -= first_row_byte_count;
@@ -8229,7 +8231,7 @@ static hipError_t iree_hip_array_copy_packed_rows(
       result = iree_hip_array_enqueue_packed_rows_copy(
           array_is_destination, array_ptr, array_info->pitch,
           (uint8_t*)external_ptr + external_offset, array_info->width_bytes,
-          full_row_count, kind);
+          full_row_count, kind, stream);
     }
     const size_t full_row_byte_count = full_row_count * array_info->width_bytes;
     external_offset += full_row_byte_count;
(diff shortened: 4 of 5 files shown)


zjgarvey:
1. **[HIGH] Map `hipStreamLegacy` to PTDS in the new async exports.**

   Both new wrappers call [`hrx_hip_spt_stream_or_explicit`](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/abi_stubs.c#L396-L403), which passes every non-null value except `hipStreamPerThread` through unchanged. Passing `hipStreamLegacy` therefore selects the legacy default stream, whereas ROCm's [`PER_THREAD_DEFAULT_STREAM`](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L267-L270) maps both null and `hipStreamLegacy` to PTDS, including in the [corresponding array exports](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_memory.cpp#L4442-L4476). This changes more than stream identity: for a contiguous copy, the downstream HRX legacy branch [synchronizes every blocking stream before enqueue](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L6902-L6910). A blocking stream callback waiting for a flag that the caller sets after this async call returns can consequently deadlock the caller inside the new API. Please normalize null, `hipStreamLegacy`, and `hipStreamPerThread` to PTDS while preserving real explicit handles, and cover `hipStreamLegacy` with a PTDS-capture test plus a bounded callback-progress case.

2. **[HIGH] Insert the legacy-default dependency on every packed and strided PTDS route.**

   PTDS is [created as a blocking stream](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L1555-L1561), and HRX's [default-stream ordering helper](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L1680-L1698) makes a blocking stream wait for the legacy default stream. The 1D copy path [calls that helper](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L6928-L6940), but `hipMemcpy2DAsync` does not. The new packed entry point sends its [complete-row segment](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8224-L8235) through [`hipMemcpy2DAsync`](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8152-L8162), and the new 2D exports reach [staged/fallback 3D branches](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7776-L7868) that also record without this edge. A legacy-default producer can therefore race a PTDS consumer, and a successful PTDS synchronization can return stale array data. Please make dependency insertion an invariant of one resolved-stream submission boundary, then test complete-row and genuinely strided array copies against a gated legacy producer.

3. **[HIGH] Keep each array retained through buffer lookup and command recording.**

   The new async exports enter [`iree_hip_resolve_memcpy3d_array_params`](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8524-L8593), which retains each array, copies out its raw device pointer, and releases both arrays before `hipMemcpy3DAsync` performs its later [buffer lookup and dereference](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7725-L7742). That lookup [retains only the owner context](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/common/memory.c#L479-L514), not the returned buffer wrapper. A concurrent [`hipFreeArray`](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L5503-L5510) can synchronize before this copy is recorded and then [remove/free the wrapper](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/common/memory.c#L1037-L1074) after lookup unlocks, leaving the first thread dereferencing freed metadata or recording against freed backing storage. Please return translated parameters together with retained source/destination ownership and release them from one cleanup exit after capture/enqueue completes. A latch-controlled copy-versus-free test under ASAN/TSAN should cover both directions.

4. **[HIGH] Use a failure-atomic synchronous core instead of public async plus conditional synchronization.**

   The rewritten synchronous wrappers [call the async API and synchronize only after an all-success return](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/abi_stubs.c#L1995-L2028). This moves synchronous capture rejection after the mutation point. A PTDS capture can receive a [graph node](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7687-L7704) before the following synchronization invalidates it; with a different visible capture, ordinary PTDS copy work can be recorded and [`hipStreamSynchronize` returns before waiting](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L11319-L11348). A registered-host D2H path can still hold the raw destination for its [later callback](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/common/memory.c#L2573-L2628) after that error. [Multi-row recording](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7427-L7506) can likewise accept a prefix before a later enqueue failure, which bypasses the success-only synchronization. Please route both wrappers through a private stream-aware synchronous core that follows ROCm's [pre-enqueue synchronous capture guard](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L233-L249), tracks any accepted work, quiesces every accepted prefix, and preserves the primary error. Cover PTDS/other-stream capture and an injected second-row failure.

5. **[MEDIUM] Preserve zero-extent and D2D host-completion behavior.**

   The same unconditional success synchronization drains earlier PTDS work after the underlying [zero-extent async no-op](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8596-L8605). It also waits for D2D copies even though HRX's existing [synchronous 3D path deliberately skips host synchronization for D2D](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7548-L7590), matching ROCm's [kind-dependent copy completion](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_memory.cpp#L2419-L2441). Besides losing overlap, a D2D call can block behind an earlier PTDS callback that expects the host to act after the copy returns. Please have the common synchronous core return immediately for a validated zero extent and choose host completion from the resolved direction. Add event/query-based no-op and D2D progress tests rather than timing thresholds.

6. **[MEDIUM] Match the reference 2D-array validation order and error codes.**

   The delegated HRX functions return `hipErrorInvalidValue` when pitch is smaller than width and return zero-extent success before validating direction ([to-array](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8596-L8617), [from-array](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8659-L8680)). ROCm validates direction and pitch first, returning [`hipErrorInvalidMemcpyDirection` and `hipErrorInvalidPitchValue`](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_memory.cpp#L2492-L2509), before its array-copy core. Please centralize validation shared by the ordinary and `_spt` forms and add production-DSO table cases for pitch, direction, offsets, arrays/pointers, zero extent, and invalid/destroyed streams.

7. **[MEDIUM] Make the DSO test prove stream identity and fail on mandatory setup errors.**

   The test [resolves only the three newly defined symbols](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/hip_array_copy_spt_api_test.cc#L31-L127), not the two synchronous 2D wrappers changed here. Its only behavioral case uses one thread, null streams, tight full-array geometry, and a pageable D2H destination; HRX [classifies that destination as pageable](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L6897-L6900) and [synchronizes the selected stream before return](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L6977-L6979). Forwarding both calls to the legacy stream therefore produces the same bytes and makes the later PTDS wait non-load-bearing. The test also uses `GTEST_SKIP` when the required DSO cannot load or when `hipInit` fails ([setup](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/hip_array_copy_spt_api_test.cc#L73-L115), [initialization](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/hip_array_copy_spt_api_test.cc#L130-L178)), allowing a GPU CTest to pass without reaching the intended assertions. Please make those mandatory preconditions fatal and exercise all five affected exports with pinned memory, independently gated legacy/PTDS/explicit streams, nontrivial offsets/pitches, capture, D2D/zero-size, concurrent free, and failure injection. Record fail-then-pass source perturbations for the central cases.

8. **[LOW] Collapse packed pageable D2H to one terminal stream wait.**

   `hipMemcpyFromArray_spt` can split a range into [prefix, complete-row, and suffix fragments](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L8200-L8257). Each nested public copy synchronizes a pageable D2H destination ([1D](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L6977-L6979), [2D](https://github.com/ROCm/hrx-system/blob/5b7f68585328e75e6314be5a7b8ffc087c6f0b32/libhrx/src/binding/hip/api.c#L7281-L7296)), and the outer helper then synchronizes once more. A three-fragment range can therefore issue four stream waits. Please use enqueue-only internal fragment helpers and retain the single terminal synchronous boundary. A trace or benchmark should compare aligned and three-fragment pageable/pinned cases; no latency magnitude is claimed here.

Verification note: all 12 public checks are green, and the gfx942 CMake job reports the new installed-DSO CTest passed. Generated CMake matches Bazel. A focused local build did not reach test execution because the pinned ROCm SDK headers lack `hsa_amd_queue_create_desc_t`; no alternate ROCm installation was used.

zjgarvey:
1. **[HIGH] Serialize captured-copy recording with `hipStreamEndCapture`.**

   The common copy path [checks `capture_status` and enters capture recording](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7691-L7695) without holding the stream mutex. The capture helper then [reads `capture_graph` and the dependency frontier, mutates the graph, and separately updates the frontier](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7084-L7111). In parallel, end-capture [holds that mutex while clearing those fields, then unlocks before returning or releasing the graph](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/graph.c#L2417-L2524); relaxed mode permits the ending thread to differ from the capture owner. An async array copy racing that end can therefore access plain capture fields concurrently, mutate a graph after it has been returned, or dereference it after the caller destroys it. Please make status validation, graph/dependency use, node insertion, and frontier commit one stream-mutex transaction, with a locked frontier helper. A TSAN regression should race a relaxed capture end against each new async array export.

2. **[HIGH] Roll back or invalidate partial captured copies on construction failure.**

   Captured H2D construction [links a hidden host callback before adding its dependent copy](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L18080-L18100), while D2H [links the copy before allocating and adding its host callback](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L18215-L18259). Any later failure returns from the copy API without removing the already linked node or invalidating capture; a failure in the subsequent [frontier update](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7106-L7110) has the same problem after the full copy was added. The caller can end and launch the graph after the reported failure, so an orphan H2D callback may read host memory the caller has already reused. Please preallocate the complete node/frontier state before linking, or remove all inserted nodes in reverse order; if rollback cannot be guaranteed, invalidate capture. Fault-inject every allocation, node-add, and frontier-growth boundary and assert that failure leaves either an unchanged graph or an invalidated capture.

3. **[HIGH] Carry selected-stream ordering through cross-context staging.**

   The core inserts the default-stream dependency into the resolved stream, but a cross-context D2D request is then [routed to the staged-row helper](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7854-L7863), whose two-device branch [passes a null stream to every D2H and H2D leg](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7501-L7521). Those direct transfers are submitted [with an empty wait list](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/libhrx/transfer.c#L32-L40). A producer already queued on the selected PTDS or explicit stream can therefore race the source read and yield stale destination bytes. Please import the selected stream's prior fence into the source transfer and publish staged completion back to that stream, retaining staging until that point. An exact-DSO two-context test should gate a producer before each new async array export and cover null, legacy, PTDS, and explicit stream arguments.

4. **[HIGH] Keep blocking array destruction on the `hipFreeArray` caller.**

   `hipFreeArray` [removes and releases the registry reference](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L5518-L5525), while normal copy ownership is released at [the common cleanup exit](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7951-L7957). If free races after retention, it can return while the copy owns the final reference; the copy then runs the array destructor, whose backing free [synchronizes every active context](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/memory.c#L1037-L1057). Put a callback ahead of that copy which waits for a flag set after the async API returns, and cleanup waits for the callback while the caller waits for cleanup: the API hangs. Please use two-phase destruction so `hipFreeArray` remains the blocking teardown owner and operation pins only signal/drop. A deterministic latch after retain, combined with the callback cycle, should exercise this exact interleaving.

5. **[HIGH] Wait only the selected stream for pageable async D2H completion.**

   The new common completion block [uses context synchronization for null or legacy streams](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7931-L7943). That context operation [includes nonblocking streams](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/context.c#L1243-L1247), even though default-stream ordering [explicitly excludes them](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/context.c#L1340-L1348) and the existing scalar async path [waits only its resolved stream](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7016-L7018). An unrelated nonblocking-stream callback waiting for a flag set after an ordinary default-stream `hipMemcpy3DAsync` or array D2H call returns now deadlocks that call. Please synchronize `stream_obj` for successful pageable async D2H regardless of the public sentinel; reserve context-wide waits for APIs whose contract requires device-wide completion. Add a gated nonblocking-callback regression for both the generic 3D and array routes.

6. **[MEDIUM] Match HIP's capture dispatch and synchronous preflight order.**

   The new 2D helpers [validate arrays, pointers, pitch, and kind before entering the stream/capture core](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L8831-L8932), and the packed helper [validates and returns zero-count success before checking capture](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L8283-L8323). The pinned HIP implementation dispatches an active async capture before the ordinary path, and its [capture helper handles zero size before operand validation](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_graph.cpp#L535-L548); synchronous APIs apply [a process-wide all-mode capture guard](https://github.com/ROCm/rocm-systems/blob/5a8a42ac5262dafc24aebd2f7c4b9f2dfe953cde/projects/clr/hipamd/src/hip_internal.hpp#L233-L249) before normal validation. HRX also limits its synchronous check to captures visible to this thread in one context. Please dispatch active async capture before ordinary preparation and run the synchronous all-capture guard before validation/no-op returns. Cover zero-size invalid operands plus thread-local and relaxed captures owned by another thread and context.

7. **[MEDIUM] Make the safety regressions force the vulnerable states.**

   `ConcurrentCopyAndFreeRetainArrayStorage` [starts both threads from one flag and accepts either copy success or invalid-value](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/hip_array_copy_spt_api_test.cc#L586-L629), so every iteration may serialize entirely before or after free. The new accepted-work cleanup at [the row loop and terminal quiescence block](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L7880-L7949) likewise has no test that fails after an earlier row was accepted. Please add a test seam that pauses after array retention, force free during that interval, and verify the copied bytes; separately inject a second-row enqueue failure and prove the accepted prefix is quiescent before the original error returns. The synchronous capture test should also seed/read back the array so submit-then-invalidate cannot satisfy status-only assertions.

8. **[LOW] Batch complete packed rows instead of staging each row separately.**

   The packed helper's complete-row segment now [loops through the scalar enqueue helper once per row](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/hip/api.c#L8389-L8404). For pageable D2H, each scalar call [allocates its own staging buffer and enqueues its own host callback](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/memory.c#L2462-L2488), while the existing 2D helper [packs all rows into one staging allocation and callback](https://github.com/ROCm/hrx-system/blob/8a6386589272ee8d7cf957da0e7f568176398307/libhrx/src/binding/common/memory.c#L2540-L2627). This replaces the old repeated waits with O(height) allocation/barrier/callback work on a full-row hot path. Please retain the single terminal wait while using a batched 2D transfer (or one contiguous transfer when layout permits), and compare resource counts for 1, 256, and 4096 rows.

Verification note: all 12 current checks are successful, and the exact-head gfx942 and gfx1201 jobs ran and passed `hip_array_copy_spt_api_test`. Bazel-to-CMake generation is current. A focused build did not reach test execution because the pinned SDK headers lack `hsa_amd_queue_create_desc_t`; no alternate ROCm installation was used.

zjgarvey (approved):
Approved.

Optional before merge: [AWoloszyn/hrx-system#7](https://github.com/AWoloszyn/hrx-system/pull/7) contains two test/build follow-ups. It holds `TestNotification`'s mutex while publishing the predicate to prevent a lost wakeup, and mirrors the existing Linux Bazel `--wrap` options in generated CMake for `memory_test`. Andrew is welcome to land it, but it does not block this HIP-layer change.

PR #568, opened 2026-09-11
[HIP] Implement virtual memory management

AWoloszyn:
Implements same-process HIP virtual memory management using the generic HRX allocator VMM contract.

  - Implements virtual-address reservation and release through `hipMemAddressReserve` and `hipMemAddressFree`.
  - Implements physical allocation creation, retention, property queries, and release through `hipMemCreate`, `hipMemRetainAllocationHandle`, `hipMemGetAllocationPropertiesFromHandle`, and
  `hipMemRelease`.
  - Implements mapping, partial unmapping, and per-location access tracking through `hipMemMap`, `hipMemUnmap`, `hipMemSetAccess`, and `hipMemGetAccess`.
  - Reports VMM support and allocation granularity from the selected device.
  - Publishes reserved address ranges through the binding pointer table so mapped addresses work with kernel launches and memory operations.
  - Retains physical backing while mappings or public handles remain live.
  - Drains active streams before unmapping because device pointers may be embedded in opaque kernel arguments or device memory and therefore cannot be tracked reliably by the binding.
  - Rejects host-NUMA and uncached placements when the underlying HAL cannot preserve their semantics instead of silently substituting ordinary device memory.
  - Keeps shareable-handle import and export explicitly unsupported because this change only implements same-process ownership.

### DIFF
--- a/libhrx/cts/tests/virtual_memory/virtual_memory_test.cpp
+++ b/libhrx/cts/tests/virtual_memory/virtual_memory_test.cpp
@@ -100,7 +100,7 @@ TEST_CASE_METHOD(HrxTestFixture, "virtual memory lifecycle",
   resources.is_mapped = true;
   REQUIRE_OK(hrx().allocator_virtual_memory_protect(
       alloc, resources.virtual_buffer, /*virtual_offset=*/0, resources.size,
-      HRX_MEMORY_PROTECTION_READ_WRITE));
+      HRX_VIRTUAL_MEMORY_ACCESS_SCOPE_ALL, HRX_MEMORY_PROTECTION_READ_WRITE));
 
   REQUIRE_OK(hrx().allocator_virtual_memory_unmap(
       alloc, resources.virtual_buffer, /*virtual_offset=*/0, resources.size));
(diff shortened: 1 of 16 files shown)


zjgarvey:
I found one compatibility issue that should be fixed before #568 merges: the
PR changes the existing five-argument
`hrx_allocator_virtual_memory_protect` C entry point in place.
[AWoloszyn/hrx-system#8](https://github.com/AWoloszyn/hrx-system/pull/8)
restores that ABI and adds a separately named scoped variant. With that repair
applied, I do not think the remaining items should block this PR.

Possible follow-ups (not requested for #568):

- define cross-device/shared-VA VMM mapping and access behavior;
- invalidate VMM objects cleanly across device reset and runtime
  deinitialization;
- distinguish mapped extents from unmapped reservation holes in ordinary
  pointer queries;
- add production `amdhip64` VMM API coverage; and
- support non-null requested virtual addresses through the HRX/HAL/backend
  layers.

GitHub currently reports conflicts with `main`, so the branch will also need
to be updated before merge.

zjgarvey (approved):

PR #617, opened 2026-09-17
[HAL/AMDGPU] Support both ROCr queue creation interfaces

zjgarvey:
## Motivation

The HSA headers available at build time can declare descriptor-based queue
creation even when a dynamically loaded ROCr runtime does not export
`hsa_amd_queue_create`. Treating that symbol as required prevents AMDGPU driver
initialization even though the stable `hsa_queue_create` interface is
available.

## Summary

Treat descriptor-based queue creation as an optional dynamic capability. Use
it when the loaded runtime exports it, and otherwise create through the stable
interface. Static builds use the stable interface directly, avoiding an
optional link dependency.

The fallback preserves requested priority and compute-unit mask behavior
before the queue is published. Dynamic HIP and CUDA export sets are unchanged.

## Design

Only `hsa_amd_queue_create` becomes optional in the dynamic HSA symbol table;
all required symbols retain strict lookup failure. The capability check is
false when the headers do not define the descriptor interface or when HSA is
linked statically, and otherwise follows the loaded function pointer.

Descriptor creation retains its 32-bit byte-length overflow check. The stable
interface uses its packet-count contract, then applies priority and the
compute-unit mask while the new queue is still private. Cooperative queues
retain normal-priority and full-resource restrictions and skip those
post-create mutations. A null queue on native success is rejected, and a
priority or mask failure destroys the unpublished queue without changing the
caller's output.

## Testing

- Release CTest: `hsa_queue_test` and `libhsa_test` (2/2 passed).
- Dynamic loader coverage: required-symbol failure remained fatal, while a
  hermetic HSA library omitting only `hsa_amd_queue_create` loaded
  successfully.
- Queue unit coverage: descriptor and stable selection, descriptor-only
  capacity overflow, null-success results, output preservation, priority and
  mask cleanup, and cooperative behavior.
- Static HSA configuration and final link passed using only
  `hsa_queue_create`; the optional descriptor symbol was not referenced.
- Bazel `//runtime/src/iree/hal/drivers/amdgpu/util:libhsa_test` passed against
  the repository-pinned HSA header revision.
- ROCm SDK `7.14.0a20260622` validation passed 27/27. Its package headers
  declare the descriptor interface while its runtime omits the optional
  symbol; the Release HSA load and queue tests passed with that package
  configuration.
- TSAN: hermetic required-symbol and missing-optional-symbol loader tests (2/2
  passed).


benvanik (changes requested):
Thanks for tracking down the initialization failure and adding focused coverage for symbol loading and queue setup rollback. Supporting a dynamically loaded ROCr library that omits `hsa_amd_queue_create` is useful. I’d like to keep that fix, with the build-time compatibility machinery removed.

The distinction here is between **the API we build against** and **the capabilities of the library we load**:

- **Build-time API:** We control our pinned HSA headers. The runtime should compile against those declarations unconditionally. If a new API requires newer declarations, we update the pinned headers. At this revision, the pin already contains the descriptor-based queue interface.
- **Dynamically loaded library:** The loaded ROCr library may lack a symbol declared by our headers. Making that symbol optional and choosing the appropriate queue-creation path at runtime is the right handling.
- **Static/link-time binding:** The selected library must satisfy our build’s API requirements. We don’t need to introduce an unconditional legacy queue-creation path to accommodate an older linked package.

An explicitly selected, outdated external header installation is the responsibility of the person selecting it. It shouldn’t introduce another supported configuration throughout the driver. Keeping declarations fixed lets us support older loaded runtimes without multiplying the source configurations we need to maintain and test.

Concretely, please:

1. **Remove `IREE_HAL_AMDGPU_HAVE_HSA_AMD_QUEUE_CREATE` and all guards derived from it**, including those in `libhsa.h`, `libhsa_tables.h`, `hsa_queue.c`, and the tests. The descriptor types, function-table entry, and call wrappers should always be available when building this driver.

2. **Keep `hsa_amd_queue_create` optional in the dynamic loader.** The dynamic capability check should reflect whether the loaded function pointer is present. Missing required symbols should continue to fail initialization.

3. **Separate the static dispatch distinction from capability detection.** With the existing static-link contract, the capability helper can return true in static mode and test the function pointer in dynamic mode. The normal `IREE_HAL_AMDGPU_LIBHSA_STATIC` distinctions needed for direct calls versus function-table dispatch remain appropriate.

4. **Preserve the fallback’s existing queue semantics.** In particular, retain the descriptor-only byte-length overflow check, the stable API’s packet-count/default-segment arguments, priority and mask configuration before publication, cooperative-queue restrictions, and destruction of an unpublished queue when setup fails.

The fallback should remain specifically for an **absent entry point**. If the descriptor function is present and returns an error, propagate that error; retrying through another API could hide a real allocation or configuration failure. The current implementation already makes this distinction correctly.

The tests should continue to demonstrate successful loading without the optional export, strict failure for missing required exports, both queue-creation paths, and cleanup/output preservation on setup failure. A focused case asserting that a present-but-failing descriptor function does **not** invoke the fallback would also help preserve this contract. The function-table fixtures can remain dynamic-only; static mode should have a separate compile/link check against the declared API.

PR #624, opened 2026-09-17
[HIP] Implement IPC memory and event handles

zjgarvey:
Depends on #623.

## Motivation

HIP exposes 64-byte IPC memory and event handles, but HRX currently leaves the corresponding entry points unsupported. That prevents independent processes from sharing AMDGPU allocations or native events and leaves the runtime-only IPC CTS uncovered. The implementation needs to preserve ROCr-specific ownership without making IPC a capability that every HAL driver must provide.

## Summary

- Implement `hipIpcGetMemHandle`, `hipIpcOpenMemHandle`, and `hipIpcCloseMemHandle` for exact, live `hipMalloc` allocations, including duplicate opens, cross-context aliases, and context teardown.
- Implement native `hipIpcGetEventHandle` and `hipIpcOpenEventHandle` support across independent processes and same-device contexts, including record, wait, query, synchronize, re-export, destruction, shutdown, and reinitialization behavior.
- Add focused driver, binding-common, HIP shared-library, unsupported-configuration, concurrency, lifetime, capture, graph, and shutdown coverage. `docs/hip_ipc.md` documents the wire formats, ownership model, transaction ordering, and supported boundaries.

## Design

The AMDGPU driver exposes typed private carriers over its existing allocator, buffer, semaphore, queue, and ROCr loader internals, and owns canonical per-token event carriers and their monitor. Binding-common code owns the process-wide memory registry, context-retirement gate, and prepare/commit/abort event transactions. The HIP binding is the public adapter: it validates API arguments, encodes and decodes the 64-byte HIP wire formats, and maps status values. No generic HAL vtable or backend obligation is added.

Memory imports use one process-wide attach anchor per token and view, with ordinary HAL buffer aliases for each importing context and reference counting for duplicate opens. Final close releases the aliases before the single detach. Context teardown first closes import admission and synchronizes submitted work, then commits all revocations together; a synchronization failure rolls the reservations back without disturbing mappings owned by other contexts.

Each IPC event token has one canonical process-local carrier. Record and wait operations reserve all fallible state before queue submission, publish only after the queue accepts the operation, and abort cleanly on rejection. A lazy low-priority monitor retains accepted operations and carrier/device ownership until completion or shutdown, so event destruction remains nonblocking and teardown cannot race the backing signal.

Handle export/open/close and one-time carrier construction are cold paths. Imported memory becomes an ordinary HAL buffer, so dispatch and graph replay continue through pre-resolved bindings and existing queue operations without an extra generic dispatch layer. Ordinary events retain only predicted-cold adapter checks and create no IPC monitor state. IPC event capture and direct graph event nodes are rejected instead of adding replay-time transport dispatch.

## Compatibility

This fills existing HIP ABI entry points and handle types; it does not add exported functions or change the generic HAL ABI. Handles are native-endian, same-host values for compatible 64-bit Linux processes. Memory uses the ROCr token layout, and events use the native type-1 ROCr signal form; shared-memory type-0 event handles are rejected. Native event wire and cross-process behavior are validated between HRX processes, but stock-runtime IPC event interoperability is not claimed.

Support is limited to Linux AMDGPU builds. Other driver configurations keep the ABI and report IPC as unsupported. Memory-pool IPC, VMM shareable handles, external memory/semaphore APIs, IPC event timing, event operations during capture, and IPC event graph nodes remain outside this change.

## Testing

- Pinned TheRock ROCm 7.14.0a20260622 SDK self-tests: 27/27.
- Bidirectional IPC memory interoperability on the selected gfx942 Linux configuration: HRX exporter to the TheRock ROCm 7.14.0a20260622 stock HIP importer passed 1/1, and the reverse direction passed 1/1.
- Fresh CMake Release build: 17/17 focused binaries and 194/194 inner cases. Fresh layering-enabled Bazel build: 17/17 targets and the same 194/194 inner cases.
- Exact-stack dependency checks: graph-layout case PASS, ROCr queue interface coverage 9/9, and HSA loader coverage 3/3.
- Fresh ASan+UBSan build: 3/3 binaries and 16/16 cases with no sanitizer diagnostics. AMDGPU-disabled production-shim coverage: 2/2 unsupported-API cases.
- Full-scope Bazel clang-tidy completed 5,598 actions with no emitted diagnostics. Fresh uncached TSAN ran 6/6 binaries: 98 PASS, 1 capability SKIP, 0 FAIL, with no TSAN diagnostics. A strict C++17 build with extensions disabled passed 99/99 cases across 6/6 binaries.
- The production DSO exposed the expected 554-symbol/version surface. The fake HSA DSO supplied all 120 required exports while omitting the optional `hsa_amd_queue_create` entry point.
- Selected runtime-only ROCm HIP IPC CTS: 14 PASS, 2 SKIP, 0 FAIL across 16 unique cases. `MultiProc::Unit_hipIpcEventHandle_Functional` skipped because fewer than two GPUs were available. `EventTest::Unit_hipEventIpc_shm_cleanup` skipped because native IPC signals make `/dev/shm` cleanup inapplicable.


benvanik (changes requested):
Thanks for working through the interoperability and lifetime questions here, and for documenting the boundaries of the results. The reported bidirectional memory interoperability and the independent-process tests give us useful evidence to build on. There is substantial work here, and I want to make sure the next steps lead toward something we can support over the long term.

The main issue is the architectural boundary. We are replacing the AMDGPU HAL on a libamdf foundation and moving away from ROCr. This implementation adds a driver-specific IPC framework, including process-wide event state, a separate completion monitor, and interfaces that expose private ROCr allocator and device details. Those would become obligations across that replacement. Keeping the interfaces private and leaving the generic HAL vtable unchanged does not remove that coupling. We cannot take this framework into the HAL in its current architectural form.

Additional support in the HAL and AMDGPU driver may well be necessary, and we are open to adding it. The goal is to identify the smallest coherent set of capabilities that composes with the rest of the system: native resource sharing and ownership where those belong, and asynchronous dependencies through our shared infrastructure. That should give libhrx and its bindings room to implement their own behavior without prescribing HIP’s object model to the runtime. HIP compatibility semantics and peer-protocol adaptation remain concerns of the HIP binding, not libhrx core or the HAL.

There is relevant work to coordinate with here. We already have async/timepoint and external-buffer import/export interfaces, and the remoting work being landed in stages includes a SHM transport with cross-process bootstrap, OS handle exchange, and wakeups integrated with `iree/async/`. TCP is landing first, with SHM following. The transport components are separate from the remote HAL, so there is an opportunity to share underlying mechanisms without adopting remote-device execution as the HIP IPC model. That work needs to be part of this design discussion before we add another IPC transport or lifecycle mechanism.

These pieces do not establish that HIP IPC is already solved. In particular, transporting a handle is different from importing a GPU allocation; a recorded timepoint is different from a reusable HIP event; and an unmodified stock-HIP peer imposes different constraints from two HRX processes. We need to identify the actual gaps and agree on where they belong. We should also keep memory sharing and event synchronization separate enough to understand whether the application needs both.

For the next discussion, please bring a short requirements description and the experiment results, covering:

1. **The concrete consumer and operation sequence.** Which application or framework is blocked, and what does each process need to do? A representative export/import/use/close sequence would help distinguish the minimum useful feature from additional API coverage. Are IPC events required for that use case, or does it already have a separate synchronization channel?
2. **The required interoperability.** Are both processes using HRX, or must one use unmodified stock HIP? Please separate memory and event requirements and identify the required platforms and runtime versions. The current report demonstrates memory interoperability on a selected stock-HIP configuration while explicitly leaving stock-HIP event interoperability unclaimed; that distinction matters to the design.
3. **The ownership and synchronization contract.** What must happen for repeated imports, event recording and reuse, outstanding queue waits, close/destruction, and producer or peer failure? Please distinguish requirements of the HIP API or actual consumer from behavior inherited from ROCr or chosen by this implementation. That will help us retain necessary guarantees without carrying implementation-specific policy into shared APIs.
4. **What the experiments establish and what remains uncertain.** Existing results are a useful starting point. Please identify the relevant performance requirements and any specific interoperability or lifecycle questions that need further experiments. We can agree on targeted experiments for those questions before asking for broader implementation or qualification.

Before implementation touching the HAL continues, we need to discuss these requirements together and agree on the design, including how it relates to the shared async and remoting work. A requirements note and focused evidence are enough for that next review; this is not a request to rebuild the feature around an assumed replacement. The current branch can remain useful for experiments while we resolve those questions.

Once the boundaries are agreed, any HAL/runtime support and the libhrx/HIP integration need separate, focused reviews. The current PR combines too many ownership and compatibility decisions to review as one change, but splitting it alone would not resolve the architectural concern. Agreeing on the contracts first should let us preserve the useful findings from this work and make the next implementation effort much more directed.

PR #625, opened 2026-09-17
[Experiment] AMDGPU SDMA transfer/compute overlap

Groverkss:
## Summary

Add native SDMA queue utilities to the IREE AMDGPU driver so callers can submit copy programs and synchronize them with compute using GPU-visible memory milestones. Queue selection, storage ownership, dependencies, and retirement remain explicit; this change does not introduce automatic HAL copy-routing policy.

- Add a gfx12.0 capability profile, bounded packet emitters, validated copy-program construction, native ROCr queue/ring ownership, wrap handling, and timestamp conversion.
- Add a PM4 memory-wait helper for compute/SDMA synchronization.
- Keep indirect buffers **disabled in reported capabilities**, with a TODO for KFD `IB_ENABLE` support and qualification of the kernel/runtime contract. Experimental IB diagnostics remain opt-in.
- Add `sdma_pipeline_benchmark` beside the PM4 benchmark. It compares direct-ring SDMA, experimental SDMA IBs, and shader copies with the same double-buffered host-to-device weight-copy/GEMM schedule. Preparation selects an optimized 4096-square gfx1201 GEMM from the caller's ROCm installation; generated code objects are not checked in. IB mode requires explicit opt-in and an execution probe.

The benchmark reports GPU intervals and CPU submission costs separately and includes preparation, execution, and plotting scripts. Packet definitions reference public ROCr/Linux interfaces. No firmware/RTL sources, kernel-module patch, internal documents, or machine-specific artifacts are included.

## Validation

Validated on the branch commit with gfx1201:

- CMake build and eight utility CTest targets.
- Both opt-in live IB tests on a kernel with IB processing enabled.
- Full-output and dependency checks for all three benchmark modes, with 60 measured samples per mode.
- Rejection of benchmark IB execution without explicit opt-in.
- Repository precommit hygiene, generated CMake consistency, and local Semgrep error rules.

## Limits

The initial SDMA profile is gfx12.0; the optimized benchmark recipe is fixed to gfx1201 and the 4096-square GEMM ABI. Experimental IB execution is not qualified for forced preemption or queue eviction. GPU power-state sensitivity affected timing comparisons, so performance results are not presented as a general speedup claim. Validation above covers the tested branch commit, not a rebuilt merge with the latest upstream changes.


benvanik (changes requested):
Thanks for exploring native SDMA submission and putting together the double-buffered copy/GEMM experiment. The explicit GPU dependencies, output checks, and documentation of the experimental limits give us useful groundwork for understanding transfer/compute overlap. I'd like to see you continue exploring this and developing the measurements and correctness evidence.

We're rewriting the AMDGPU HAL on the libamdf foundation, replacing ROCr underneath it. That work includes a different public transfer API intended to accommodate AMDGPU SDMA, XDNA DMA, and Vulkan transfer queues. The transfer, dependency, completion, and lifetime contracts need to work across those backends. Adding another path through the current AMDGPU driver would create integration and qualification work against the foundation and APIs we're replacing.

Given that direction, this implementation and its experiment tooling aren't a fit for landing in the tree. Please keep developing the experiment on its branch and use this PR to collect the findings. Understanding when SDMA helps, how it interacts with compute, and what its synchronization requirements are can directly inform the rewrite. Those findings are what we'd carry forward; production integration will take shape around the new transfer API.

For the next round of experiments, I'd suggest:

1. **Measure the performance tradeoffs.** Compare SDMA and shader copies with matched workloads over a few representative transfer sizes and copy/compute overlap conditions. Report benefits, regressions, and inconclusive results, with clock/power conditions and run-to-run variation. Separate preparation, steady-state submission, and complete end-to-end timing. In the current harness, [the submission routines](https://github.com/ROCm/hrx-system/blob/6cf85a47ab55ed8ca8a50049e01b6c4337bf74dc/runtime/src/iree/hal/drivers/amdgpu/sdma_pipeline_benchmark.cc#L224-L263) include different allocation/upload costs, and [`host_us` starts after submission](https://github.com/ROCm/hrx-system/blob/6cf85a47ab55ed8ca8a50049e01b6c4337bf74dc/runtime/src/iree/hal/drivers/amdgpu/sdma_pipeline_benchmark.cc#L500-L522). Making those boundaries explicit will let us distinguish execution overlap from host-side preparation and submission costs.

2. **Strengthen the evidence for the native execution contract.** Extend the correctness experiments around copy/compute dependencies, memory visibility, safe buffer reuse, ring wrapping, and completion before reclaiming storage. The current [memory-milestone contract](https://github.com/ROCm/hrx-system/blob/6cf85a47ab55ed8ca8a50049e01b6c4337bf74dc/runtime/src/iree/hal/drivers/amdgpu/util/sdma_program.h#L23-L53) is a useful place to investigate those mechanics without introducing a HAL integration. Record which constraints come from the hardware/kernel and which come from the ROCr-backed harness. Keep direct-ring and indirect-buffer conclusions separate; the IB execution probe does not establish the preemption or eviction behavior that remains unqualified.

3. **Make the results reproducible and their scope clear.** Preserve raw samples and timelines, correctness results, the prepared-data manifest, and the exact hardware, kernel, ROCr, and IB-enablement configuration. Explain how the fixed gfx1201/GEMM experiment relates to an intended application workload and where its conclusions stop. A bounded experiment on the available platform is sufficient for this stage; broad supported-platform qualification belongs with the future implementation.

An updated PR description that presents the question, baseline and method, observed results, and remaining uncertainties would make this a useful experimental reference for the rewrite. Both positive and negative results matter here: they help us understand which workloads benefit from SDMA and which costs or constraints the new HAL needs to account for. Please keep exploring along those lines and attach the results here.

PR #712, opened 2026-09-20
[Loom] Lower x86 remainder by seven with explicit multiply ownership

benvanik:
Lower unsigned remainder by seven on x86 using the high half of a native multiply. The source rule covers every i32 bit pattern and index values proven to fit u32; it does not narrow unbounded inputs or reinterpret signed remainder. The sequence works on the scalar x86-64 contract without requiring BMI2.

### Multiply ownership survives lowering

The multiply's RAX input, RDX high result, and RAX low-result overwrite are visible to the existing scheduler and allocator. Singleton RAX/RDX classes share the canonical physical GPR inventory with ordinary 32- and 64-bit values. Their unspillable contract keeps constrained-register pressure visible to scheduling. Native assembly omits the implicit operands but does not invent register assignments hidden from allocation.

Source lowering preserves the numerator before entering that constrained domain, and the multiply input depends on the preserved copy. This matters when the original arrives in RAX or RDX and remains live after the multiply. Copies between ordinary and singleton GPR classes retain physical aliasing and width semantics; a 32-bit write still clears dirty upper bits.

The reciprocal comes from the shared unsigned-divisor recipe. Its new 64-bit projection incorporates the correction and post-shift, so the high product directly yields the quotient for a zero-extended u32 numerator. The existing 32-bit recipes remain unchanged.

### Shared allocation behavior

Optional copy coalescing now uses the same retained physical-domain preferences as ordinary register search. A broad value therefore does not occupy a narrow domain merely because a copy can be eliminated, leaving constrained users without their required register. Mandatory ties and storage aliases retain their existing rules.

The shared storage queries also consume validated physical identities directly for whole-range comparisons and atomic-unit indexing. Direct alias lookup uses the generated reverse membership map; aggregate subranges still use their declared views. This removes repeated view reconstruction from the existing XDNA allocation path as well as the new x86 representation.

### Cost and coverage

The emitted remainder fragment is 31 bytes versus 28 bytes for the same-ISA Clang control, with matching measured dependent-chain latency and independent-lane throughput on Zen 5. The optimized stripped all-target compiler grows by 3,968 bytes (0.03%), including 80 bytes of text and 2,432 bytes of read-only data.

Explicit physical ownership adds 0.20–0.36 microseconds (3–8%) to the small ordinary x86 frame-construction controls; their native output and arena use are unchanged. Full AMDGPU compile-and-emit controls stay within 1%. XDNA gate/up compilation is neutral to about 2.1% faster across four interleaved comparisons: one neutral result and three roughly 2% improvements. The four AMDGPU control images and the XDNA gate/up image are byte-identical between baseline and candidate, so the compiler measurements do not represent a change to those device kernels.

The common source corpus covers word and bounded-index remainders with live-through inputs across all seven target families. Focused native cases cover RAX/RDX aliases and independent products under all four scheduling strategies. Generator tests check shared physical identities across x86 profiles; arithmetic tests compare reciprocal projections against integer division. Signed and wider source domains retain precise rejection diagnostics.

### Reviewer notes

The commits separate the independently useful physical-storage query improvement from the complete x86 ownership and source-lowering slice. The highest-value review points are the implicit RAX clobber, the numerator-copy dependency, and the distinction between optional coalescing preferences and mandatory placement ties.


benvanik:
Retracting this implementation for local redesign. The source lowering is specific to one divisor; the replacement must support a general constant-division/remainder family with divisor-independent production code and demonstrate correctness and native performance across recipe classes.

PR #738, opened 2026-09-20
[Loom] Retain native x86 SysV calls

benvanik:
X86 previously made inlining a correctness requirement: any source call
surviving the inline pipeline could not reach native emission. This change
carries ordinary x86-64 SysV calls through source lowering, scheduling,
allocation, ABI materialization, scalar encoding, format-neutral native
objects, and ELF64 relocatable serialization. The resulting object links and
executes with normal C code; ordinary functions use the platform convention
without kernel wrappers.

The first commit is the rebased implementation from #704. It remains visible
here because the direct-call slice proves that a selected view crosses the ABI
as its target-native pointer representation. #704 itself is left untouched.

### Design

- Classify and retain SysV layouts before source types are erased. Schedule
  call and return points once, then derive frame preservation from retained
  sparse liveness and physical extents.
- Represent incoming stack loads and outgoing stack stores directly in Low as
  `low.func.stack_arg` and `low.func.call_arg`. The existing x86 ABI
  materialization pass creates these after inlining and symbol DCE, so the
  allocator sees their real dependencies without rescanning calls.
- Compose incoming arguments, outgoing arguments, spills, caller saves,
  callee saves, and parallel-move scratch in one SysV frame plan. Leaf and
  register-only paths avoid stack-argument storage.
- Interpret generated scalar encoding recipes over the accepted schedule. The
  encoder's two linear passes measure bytes and then emit them; this adds no
  CFG or callgraph pass.
- Keep section contents, symbols, and fixups format-neutral until an ELF64
  relocatable adapter serializes them. This keeps COFF, Mach-O, and executable
  library packaging as separate output concerns.
- Verify calls against provider-produced concrete target-context ordinals.
  Refined functions in one concrete context remain callable, while calls
  across genuinely distinct contexts fail at the target-aware boundary.
  Module target predicates keep x86 ABI materialization off other targets.

### Results

The linked production witness passes 16 integer arguments in each direction,
including ten incoming and ten outgoing stack arguments, preserves live values
across the call, and returns 251 through a host C function.

Locked 50-million-call AB/BA soaks measured the retained path at 1.975 ns/call,
down from 2.104 ns/call for the previous retained representation. Forced
inlining of the deliberately tiny helper remains faster at 1.668 ns/call,
which preserves the expected distinction between direct-call correctness and
inline profitability.

At 4096 callers, register-only compilation changed by -1.20% wall time and
+1.52 MiB peak RSS. Stack-argument compilation changed by +2.32% wall time and
+3.07 MiB peak RSS, with no quadratic knee and identical emitted object sizes.
The optimized compiler grows by 24,096 bytes of text, 3,328 bytes of read-only
data, and 29,968 bytes in the stripped executable.

Coverage includes a final 160-test ASAN matrix across Low, XDNA, AMDGPU, VM,
Wasm, x86, native emission, linked execution, and compiler tooling. The full
presubmit selection also passed 532 tests, clang-tidy, and Semgrep.

### Reviewer notes

The highest-value review surfaces are the shared call-point and liveness
contracts, the SysV frame plan and ABI materialization boundary, the numeric
x86 recipe interpreter, the generic native-object/ELF split, and concrete
target-context verification.


benvanik:
Retracted as a rejected backend design. This change attempted full native x86 backend enablement; its ABI contracts, implementation size, and review scope do not meet the project requirements. The measured 100,688-byte stripped executable growth is unacceptable. PR #704 remains the independent view-selection change and will be prepared separately. This branch is retained locally only as a reference for a future redesign.

PR #763, opened 2026-09-21
[Loom] Import first-class C++ views and layouts

benvanik:
C++ kernels can now use Loom rank-two views and address layouts as ordinary typed values. A view can be constructed from an interior pointer and a runtime stride, passed through normal helpers, selected by structured control flow, cropped with a deduced subview, and loaded or stored without flattening its geometry back into pointer arithmetic. Import preserves the dynamic extent, layout, and storage identity as durable Loom SSA, so the result remains analyzable, editable High IR.

```cpp
#include <loomcxx/kernel.h>
#include <loomcxx/view.h>

namespace loomt = loom::type;
using rows8 = loomt::view<const float, loomt::dynamic, 8>;

static rows8 preserve_view(rows8 source) { return source; }

[[loom::kernel, loom::workgroup_size(8, 8, 1),
  loom::workgroup_count(1, 1, 1)]]
void copy_strided(const float* input, float* output, unsigned rows,
                  unsigned input_stride, unsigned input_origin,
                  unsigned output_origin) {
  auto input_layout =
      loom::encoding::layout::strided(input_stride, 1u);
  auto output_layout = loom::encoding::layout::dense<2>();

  auto source = preserve_view(
      loom::buffer::view<loomt::dynamic, 8>(
          input + input_origin, {rows}, input_layout));
  auto destination =
      loom::buffer::view<loomt::dynamic, 8>(
          output + output_origin, {rows}, output_layout);

  loomt::size_type row = loom::workitem_id.y;
  loomt::size_type column = loom::workitem_id.x;
  if (row < rows) {
    float value = loom::view::load(source, row, column);
    loom::view::store(value, destination, row, column);
  }
}
```

The source above lowers to ordinary Loom operations. The dynamic row count and layout are explicit values, and the view type refers to those exact identities:

```loom
%input_layout = encoding.layout.strided [%input_stride, %one]
    : encoding<layout>
%source = buffer.view %input[%input_byte_offset]
    : buffer -> view<[%source_rows]x8xf32, %input_layout>

%returned_rows, %returned_layout, %returned_source =
    func.call @preserve_view(
        %source_rows, %input_layout, %source)
    : (index, encoding<layout>,
       view<[%source_rows]x8xf32, %input_layout>)
   -> (index, encoding<layout>,
       view<[%returned_rows]x8xf32, %returned_layout>)

%value = view.load %returned_source[%row, %column]
    : view<[%returned_rows]x8xf32, %returned_layout> -> f32
view.store %value, %destination[%row, %column]
    : f32, view<[%destination_rows]x8xf32, %output_layout>
```

### What this enables

Static shape information stays in the C++ type while runtime shape and layout stay live in Loom:

```cpp
using rows32 =
    loomt::view<const float, loomt::dynamic, 32>;

auto source = loom::buffer::view<loomt::dynamic, 32>(
    input, {rows},
    loom::encoding::layout::strided(input_stride, 1u));
```

Subview calls can preserve that static/dynamic pattern without repeating template arguments. The first call keeps the dynamic row count and static width of 32; the second deliberately changes the static width to 16:

```cpp
auto tail =
    loom::view::subview(source, {first, 0}, {remaining});

auto narrow_tail =
    loom::view::subview<loomt::dynamic, 16>(
        source, {first, column}, {remaining});
```

Views can cross ordinary abstraction boundaries. Helper returns and conditional joins receive fresh dependent result identities instead of accidentally referring back to an argument's SSA values:

```cpp
static rows32 choose(bool use_right,
                     rows32 left, rows32 right) {
  return use_right ? right : left;
}

auto selected = choose(rows > 2, source, tail);
if (rows > 3) {
  selected = tail;
}
float value = loom::view::load(selected, 0, column);
```

That conditional becomes an `scf.if` returning the extent, encoding, and dependent view together:

```loom
%selected_rows, %selected_layout, %selected =
    scf.if %condition
      -> (index, encoding<layout>,
          view<[%selected_rows]x32xf32, %selected_layout>) {
      scf.yield %tail_rows, %tail_layout, %tail
    } else {
      scf.yield %source_rows, %source_layout, %source
    }
```

These are useful beyond writing the final kernel directly in C++. An author can import familiar C++ to inspect how source concepts map into Loom, edit the resulting High IR, add scheduling or specialization, and compare it with an authored Loom implementation. Kernel ports can retain typed helper structure while replacing compatibility APIs incrementally. The same value-projection boundary also gives future HIP-, CUDA-, or domain-specific facade headers a place to map their public vocabulary onto Loom operations without making their runtime or header stack part of the compiler.

### Public source contract

The public `loomcxx/view.h` facade introduces:

- `loom::type::encoding<loom::encoding::role::layout, Rank>`
- `loom::type::view<T, Rows, Columns>`
- `loom::type::dynamic` and `loom::type::size_type`
- `encoding::layout::dense` and `encoding::layout::strided`
- `buffer::view`
- deduced and explicitly reshaped `view::subview`
- typed `view::load` and `view::store`

The source representations are real, trivially copyable C++ objects across LP64, LLP64, and ILP32. Static extents remain template arguments. Dynamic extents and layouts project to first-class Loom values. The C++ vocabulary deliberately follows the corresponding dialect operations, making `view.subview` and `loom::view::subview`, for example, directly searchable across source, IR, descriptors, and documentation.

### Importer design

The importer represents each admitted source value as a partition of destination components. It reserves the complete destination identity range before constructing dependent types, which lets parameters, results, calls, records, and structured-region results bind view types to their own extent and layout values.

Concrete intrinsic template specializations are resolved lazily from their admitted `loom::op` declarations. Intrinsic dispatch retains the source-level composite value until the selected binding projects it; scalar, vector, and pointer operations keep their existing compact path. View and encoding validation and construction live in a dedicated binding package rather than accumulating in the translation driver.

### Qualification

The focused import corpus covers dense and runtime-strided layouts, static and dynamic extents, deduced and width-changing subviews, helpers, returns, conditional expressions, conditional assignment, typed loads and stores, data-model object layout, source diagnostics, cleanup, and text roundtrip.

The native execution witness runs the complete source-to-device route: C++ import, source teardown, loombc linking, AMDGPU compilation, and execution. It copies from a runtime-strided source view at a nonzero pointer origin into a dense destination at a distinct origin. Independent cases cover zero, one, three, and seven rows; row pitches of 8, 11, and 17 elements; unchanged input storage; output guards; and device access reporting.

### Reviewer notes

- `value/representation`, `value/types`, and `value/signature` own component partitions and destination-bound dependent types.
- `binding/view` owns facade validation and operation construction.
- `translation.cc` and `symbol/functions.cc` integrate complete values at callable and structured-control boundaries.
- The focused importer golden exposes exact SSA identity flow; the native witness exercises the shipping compile and execution path.


Groverkss:
This is very cool!

